PR #20, opened 2024-09-17
feat: add table exporters

dolfim-ibm:


PeterStaar-IBM (commented):
Can you please add the to_document_tokens function (factor it out from the to_document_tokens from the Document class)?

dolfim-ibm:
> Can you please add the to_document_tokens function (factor it out from the to_document_tokens from the Document class)?

🤔 we need a bit more refactoring for it, since it would require knowledge about the parent page. let's keep it for the next iteration.

PeterStaar-IBM (approved):
never mind, let's do the to-doc-tags later

PR #21, opened 2024-09-17
feat: (experimental) introduce new document format

cau-git:


docling_core/types/newdoc/document.py
```diff
@@ -0,0 +1,95 @@
...
+
+class ProvenanceItem(BaseModel):
+    page_no: int
+    bbox: BoundingBox
+    charspan: Tuple[int, int]
+
+
+class DocItem(BaseModel):
+    dloc: str  # format spec ({document_hash}{json-path})
+    hash: int
```
PeterStaar-IBM, line 41: Can we specify explicitly,

from pydantic import BaseModel, conint

# Define a constrained integer for uint64
Uint64 = conint(ge=0, le=(2**64 - 1))

class DocItem(BaseModel):
    hash: Uint64

docling_core/types/newdoc/document.py
```diff
@@ -0,0 +1,95 @@
+from typing import Any, Dict, List, Optional, Tuple, Union
+
+from pydantic import AnyUrl, BaseModel, Field
+
+from docling_core.types.newdoc.base import BoundingBox, Size
+
+
+class FigureData(BaseModel):  # TBD
```
PeterStaar-IBM, line 8: call this with BaseFigureData

docling_core/types/experimental/document.py
```diff
@@ -0,0 +1,1118 @@
...
+    num_rows: int = 0
+    num_cols: int = 0
+
+    @computed_field  # type: ignore
+    @property
+    def grid(
+        self,
+    ) -> List[
+        List[TableCell]
+    ]:  # TODO compute grid representation on the fly from table_cells
```
PeterStaar-IBM, line 78: remove the TODO

docling_core/types/experimental/document.py
```diff
@@ -0,0 +1,1118 @@
...
+                    min(cell.end_col_offset_idx, self.num_cols),
+                ):
+                    table_data[i][j] = cell
+
+        return table_data
+
+
+class FileInfo(BaseModel):
+    """FileInfo."""
+
```
PeterStaar-IBM, line 112: dont we want the filename too?

PeterStaar-IBM: do we want a uint64 from the has too

docling_core/types/experimental/document.py
```diff
@@ -0,0 +1,1118 @@
...
+        else:
+            obj = doc.__getattribute__(path)
+
+        return obj
+
+
+class ImageRef(BaseModel):
+    """ImageRef."""
+
+    format: str  # png, etc.
```
PeterStaar-IBM, line 151: do we want this as an enum

docling_core/types/experimental/document.py
```diff
@@ -0,0 +1,1118 @@
...
+    parent: Optional[RefItem] = None
+    children: List[RefItem] = []
+
+    def get_ref(self):
+        """get_ref."""
+        return RefItem(cref=f"#{self.dloc.split('#')[1]}")
+
+    @computed_field  # type: ignore
+    @property
+    def hash(self) -> Uint64:  # TODO align with hasher on deepsearch-glm
```
PeterStaar-IBM, line 178: do we want the TODO?

docling_core/types/experimental/document.py
```diff
@@ -0,0 +1,1118 @@
...
+        :param new_line: str:  (Default value = "\n")
+        :param xsize: int:  (Default value = 100)
+        :param ysize: int:  (Default value = 100)
+        :param add_location: bool:  (Default value = True)
+        :param add_content: bool:  (Default value = True)
+        :param add_page_index: bool:  (Default value = True)
+
+        """
+        body = f"<{self.label.value}>"
+        # body = f"<{self.name}>"
```
PeterStaar-IBM, line 267: remove this

docling_core/types/experimental/document.py
```diff
@@ -0,0 +1,1118 @@
...
+                    body += f"<col_{j}>{cell_loc}{cell_label}{text}</col_{j}>"
+
+                body += f"</row_{i}>{new_line}"
+
+        body += f"{DocumentToken.END_TABLE.value}{new_line}"
+
+        return body
+
+
+class KeyValueItem(DocItem):
```
PeterStaar-IBM, line 571: lets add the key and the value

docling_core/types/experimental/document.py
```diff
@@ -0,0 +1,1118 @@
...
+            label, text, orig, prov, parent, item_cls=Section
+        )
+        item.level = level
+        return item
+
+    def num_pages(self):
+        """num_pages."""
+        return len(self.pages.values())
+
+    def build_page_trees(self):
```
PeterStaar-IBM, line 793: let's comment this out for the moment

docling_core/types/experimental/document.py
cau-git: Yes, we keep that for the moment.

docling_core/types/experimental/labels.py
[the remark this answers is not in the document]
cau-git: My experience with enum `auto()` is that it serializes as simple numbers, not recognizable without the definition. That's why I went away from it. We also need an explicit string value for the labels, for other codes to compare on which don't have `docling-core` dependency.

docling_core/types/experimental/document.py
```diff
@@ -0,0 +1,1106 @@
...
+                ):
+                    table_data[i][j] = cell
+
+        return table_data
+
+
+class FileInfo(BaseModel):
+    """FileInfo."""
+
+    filename: str
```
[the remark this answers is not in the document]
dolfim-ibm: I think here we want to have only the "stem" of the path, not the actual path.

docling_core/types/experimental/document.py
```diff
@@ -0,0 +1,1106 @@
...
+    image: Optional[ImageRef] = None
+    page_no: int
+
+
+class DoclingDocument(DocumentTrees):
+    """DoclingDocument."""
+
+    version: str = "0.0.1"  # = SemanticVersion(version="0.0.1")
+    description: Any
+    file_info: FileInfo
```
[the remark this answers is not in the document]
dolfim-ibm: we are trying to stay away from optional (=nullable) fields, to avoid users having to always check `if doc.file_info is not None`. if possible, good defaults (with the proper type) should be preferred.

> A document doesn't need to be backed by a file.
this is actually a very valid observation.

docling_core/types/experimental/labels.py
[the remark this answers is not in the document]
nikos-livathinos: In this case we have `StrEnum` and we set the value to the lowercased version of the member's name. That's what `auto()` does: 
> For Enum and IntEnum that appropriate value will be the last value plus one; for Flag and IntFlag it will be the first power-of-two greater than the highest value; for StrEnum it will be the lower-cased version of the member’s name.

docling_core/types/experimental/document.py
```diff
@@ -0,0 +1,1106 @@
...
+    hash: str  # page hash
+    size: Size
+    image: Optional[ImageRef] = None
+    page_no: int
+
+
+class DoclingDocument(DocumentTrees):
+    """DoclingDocument."""
+
+    version: str = "0.0.1"  # = SemanticVersion(version="0.0.1")
```
vagenas, line 599: Some thoughts:
- This would imply manual maintenance overhead & the error-proneness inherent to that (e.g. if dev forgets to update)
- One could consider the alternative of directly using the docling-core package version, which is currently managed automatically, so above issues would not exist — however multiple package versions would correspond to the same document data model, so there would be no clear version management of the document data model itself

I therefore think we should indeed stick to the manual version here.

👉 Note though, the first version in SemVer is typically [0.1.0](https://semver.org/#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase).

docling_core/types/experimental/document.py
```diff
@@ -0,0 +1,1106 @@
...
+            tbl_item.prov.append(prov)
+        if caption:
+            tbl_item.captions.append(caption.get_ref())
+
+        self.tables.append(tbl_item)
+        parent.children.append(RefItem(cref=cref))
+
+        return tbl_item
+
+    def add_figure(
```
ceberam, line 721: We could avoid some duplicate code among `add_figure`, `add_table`, and `add_paragraph`

docling_core/types/experimental/document.py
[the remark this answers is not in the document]
cau-git: Yes, I think we should have only a way to name a file, but it is _not_ the input file, and also it does not mean that it is written to a filesystem. Hence only `str`.

docling_core/types/experimental/document.py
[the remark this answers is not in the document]
cau-git: A document does not have to be backed by a file, but it will always need a unique identifier (`document_hash`) and it should have a human-recognizable name (`filename`). Even it if is constructed out of nothing on the fly. We have the `file_info` and `description` objects to ensure we don't scatter many fields on the root between what is supposed to be content.

docling_core/types/experimental/labels.py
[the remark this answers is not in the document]
cau-git: For the labels, I disagree, _we_ have to define also the label names as constants, since downstream code probably makes string-comparisons on labels if they don't want to depend on `docling-core`.

docling_core/types/experimental/document.py
cau-git: I was hoping that we could use a specific semver type that only accepts valid version strings, but it seemed broken in [`pydantic-extra-types`](https://pypi.org/project/pydantic-extra-types/)

ceberam (commented):
I have left some inline comments. In addition, some general comments:

In addition, some general comments:

- Labels like `DocItemLabel`: it is transparent to enumerate them in the code and we also need some explicit labels for certain functions (like `TABLE`), so I understand this choice over using pydantic generic models and the label set as a parameter (like we use in [ExportedCCSDcocument](https://github.com/DS4SD/docling-core/edit/main/docling_core/types/doc/document.py#L350) with identifiers, languages, etc.). However, this may remove flexibility if users want to extend the labels and use their own. For instance, different levels of subtitles or a`CLAIM` label in patents. Suggestion: classes that use the lables, like `DocItem` could be generic and the field `label` could be the union of `DocItemLabel` and other strings (if specified).
- I understand that we are trying to stay away from optional nullable fields, to avoid python users to check if a field is None or not. I see the advantages, but we need to be aware that this may be cumbersome for light documents and for some client applications (e.g., Elastic). I also see some contradictions (`GroupItem.label` is set to `UNSPECIFIED` by default but the `name`  field can be left as `None`)
- What is the criteria to include (pydantic) models in `base.py` or in `document.py` ? Some pydantic models like `ProvenanceItem`, `NodeItem` could be in `base.py`, for the same reason that `BoundingBox` is in `base.py`.
- Suggestion: we can add `DoclingDocument` in [ds_generate_docs](https://github.com/DS4SD/docling-core/blob/main/docling_core/utils/ds_generate_docs.py) to automatically generate markdown documentation of this class.
- In the previous type ([ExportedCCSDcocument](https://github.com/DS4SD/docling-core/edit/main/docling_core/types/doc/document.py#L350) ) we had extra labels that controlled how we can best index documents in search databases (Elasticsearch, OpenSearch) to support search applications. Do we want to do the same here? It does not impact the model interface, it is just adding tags in `json_schema_extra` field of pydantic models.
- Documentation with docstrings in Sphynx: we adopted the Google convention, as stated here https://github.com/DS4SD/docling-core/blob/main/.flake8#L5. The advantage of this convention is that it is lighter than Sphynx and it avoids redundancies (no need to specify again the static types if they are already in type hints). Check https://google.github.io/styleguide/pyguide.html and https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#google-vs-numpy. 
- Type hints: they are missing in some functions, specially the returned type. Since in many others we do have them, I would add the missing ones for consistency. 
- More negative tests would be helpful (like the ones we had in CCSDocument), to ensure we trigger errors in documents that we consider invalid.

test/test_docling_doc.py
```diff
@@ -0,0 +1,201 @@
...
+    # print(f"\n\n{yaml_dump}")
+    DoclingDocument.model_validate(yaml.safe_load(yaml_dump))
+
+
+def _test_export_methods(doc):
+    ### Iterate all elements
+    doc.print_element_tree()
+    ## Export stuff
+    print(doc.export_to_markdown())
+    print(doc.export_to_document_tokens())
```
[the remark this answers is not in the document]
cau-git: That's all for debugging purposes, we can call these methods without `print` just to ensure they don't raise any exception.

docling_core/types/experimental/document.py
```diff
@@ -0,0 +1,1106 @@
...
+
+    page_no: int
+    bbox: BoundingBox
+    charspan: Tuple[int, int]
+
+
+class NodeItem(BaseModel):
+    """NodeItem."""
+
+    dloc: str  # format spec ({document_hash}{json-path})
```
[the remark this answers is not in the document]
cau-git: @ceberam Agreed, consider it open to edit.

docling_core/types/experimental/document.py
```diff
@@ -0,0 +1,1106 @@
...
+    size: Size
+    image: Optional[ImageRef] = None
+    page_no: int
+
+
+class DoclingDocument(DocumentTrees):
+    """DoclingDocument."""
+
+    version: str = "0.0.1"  # = SemanticVersion(version="0.0.1")
+    description: Any
```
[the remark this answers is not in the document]
cau-git: This is simply unfinished business, I would like you to have a pass over that to check what should be the minimum set of fields we want to support.

docling_core/types/experimental/document.py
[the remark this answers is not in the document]
vagenas: > but it will always need a unique identifier (document_hash)

Until now we have had the document_hash from PDF conversion (which BTW was actually based on the original PDF bytes, not the resulting representation after conversion).

Now that we broadening Document creation beyond PDF conversion, we should consider (additionally?) having a generic identification / deduplication mechanism.

E.g. it could be provided by the Document class, as a checksum on its nodes/content (possibly incl. metadata too).
Implications:
- if it is a field, all add* methods would be responsible for recalculating/updating it
- if it is a method, there is no such need, but then doc serialization/deserialization gets bit more complicated.

docling_core/types/experimental/document.py
```diff
@@ -0,0 +1,1106 @@
...
+        to_element: Optional[int] = None,
+        labels: list[str] = [
+            "title",
+            "subtitle-level-1",
+            "paragraph",
+            "caption",
+            "table",
+            "Text",
+            "text",
+        ],
```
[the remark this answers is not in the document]
cau-git: yes.

docling_core/types/experimental/document.py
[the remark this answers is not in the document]
vagenas: Since each node is already connected with its `parent` and `children`, I am wondering why we need to materialize the path of `dloc` (even including the actual doc ID/hash).

In which scenario is `dloc` going to be used?

docling_core/types/experimental/labels.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class GroupLabel(str, Enum):
+    """GroupLabel."""
+
+    UNSPECIFIED = "unspecified"
+    LIST = "list"  # group label for list container (not the list-items)
+    CHAPTER = "chapter"
+    SECTION = "section"
+
```
maxmnemonic, line 42: Let's add "SLIDE" and "SHEET"

cau-git:
@ceberam we have some toxic stuff in the pre-commit pipeline, which removes unused imports also from `__init__.py` files. This must be removed to be able to create a maintainable import system on the package.

vagenas:
> @ceberam we have some toxic stuff in the pre-commit pipeline, which removes unused imports also from `__init__.py` files. This must be removed to be able to create a maintainable import system on the package. It seems to come from `autoflake`.

@cau-git indeed, it's probably going to get removed soon. Meanwhile you can just append `# noqa` to these lines, because even if not removed by autoflake, they will otherwise  break flake8.

cau-git:
> > @ceberam we have some toxic stuff in the pre-commit pipeline, which removes unused imports also from `__init__.py` files. This must be removed to be able to create a maintainable import system on the package. It seems to come from `autoflake`.
> 
> @cau-git indeed, it's probably going to get removed soon. Meanwhile you can just append `# noqa` to these lines, because even if not removed by autoflake, they will otherwise break flake8.

I found it to go _way_ further, also `flake8` and `mypy` seem to cause issues.

docling_core/types/experimental/document.py
```diff
@@ -0,0 +1,1106 @@
...
+                        has_title = True
+
+                    # secondary titles
+                    elif item_type in {"title", "subtitle-level-1"} or (
+                        has_title and item_type == "title"
+                    ):
+                        if strict_text:
+                            markdown_text = f"{text}"
+                        else:
+                            markdown_text = f"## {text}"
```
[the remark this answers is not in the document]
cau-git: Was forced to roll this exact stuff back in other places because the line length exceeded 88, and one _can not break_ these inline conditionals across lines.

docling_core/types/experimental/document.py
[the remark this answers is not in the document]
cau-git: `dloc` is a unique identifier string for any element, and makes it traceable back to a field in a document. For example it can act as provenance in a vector store or other places.

docling_core/types/experimental/document.py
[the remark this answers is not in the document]
cau-git: It could be a pydantic `@computed_field` instead.

docling_core/types/experimental/document.py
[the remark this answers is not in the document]
vagenas: I see. The way I mean it is that materializing this info on the node level may be making the whole design a bit too rigid (for example the design would otherwise still allow e.g. easy removals or additions of nodes at random positions)

> it can act as provenance in a vector store

This could be under the responsibility of a chunker having to chunkify a whole doc to nodes, producing the paths while iterating.

docling_core/types/experimental/document.py
```diff
@@ -0,0 +1,1106 @@
...
+
+
+class KeyValueItem(DocItem):
+    """KeyValueItem."""
+
+
+ContentItem = Union[TextItem, FigureItem, TableItem, KeyValueItem]
+
+
+class DocumentTrees(BaseModel):
```
vagenas, line 576: A class name in singular would be preferable.

docling_core/types/experimental/document.py
```diff
@@ -0,0 +1,1106 @@
...
+    """DocumentTrees."""
+
+    furniture: GroupItem = GroupItem(
```
vagenas, line 579: At first sight `furniture` appears a bit odd — is it some kind of terminology?
A short description comment would anyway be very helpful.

docling_core/types/experimental/document.py
```diff
@@ -0,0 +1,1106 @@
...
+        hash_int = int.from_bytes(hash_object.digest(), "big")
+
+        # Mask it to fit within 64 bits
+        return Uint64(hash_int & 0xFFFFFFFFFFFFFFFF)  # 64-bit unsigned integer mask
+
+
+class GroupItem(NodeItem):  # Container type, can't be a leaf node
+    """GroupItem."""
+
+    name: Optional[str] = None
```
vagenas, line 193: Some short documentation of what is meant by `name` would be helpful.

docling_core/types/experimental/document.py
```diff
@@ -0,0 +1,1106 @@
...
+    label: GroupLabel = GroupLabel.UNSPECIFIED
```
vagenas, line 194: In other (most) places we have explicit typing with subclasses etc, but here we have a single type with a label instead. Maybe we want to use something like a [discriminated union](https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions)?

docling_core/types/experimental/document.py
```diff
@@ -0,0 +1,1106 @@
...
+    tables: List[TableItem] = []
+    key_value_items: List[KeyValueItem] = []
+
+    pages: Dict[int, PageItem] = {}  # empty as default
+
+    def add_group(
+        self,
+        label: Optional[GroupLabel] = None,
+        name: Optional[str] = None,
+        parent: Optional[GroupItem] = None,
```
vagenas, line 615: Do we really want to allow the case of label, name, and parent all being None?

docling_core/types/experimental/document.py
```diff
@@ -0,0 +1,1106 @@
...
+        """num_pages."""
+        return len(self.pages.values())
+
+    def iterate_elements(
+        self,
+        root: Optional[NodeItem] = None,
+        with_groups: bool = False,
+        traverse_figures: bool = True,
+        page_no: Optional[int] = None,
+        _level=0,  # fixed parameter, carries through the node nesting level
```
vagenas, line 792: Typing information seems to be missing here.

docling_core/types/experimental/document.py
cau-git: Probably no. There must be a parent always. `label` has a default, so users don't need to provide it. `name` is optional.

docling_core/types/experimental/document.py
cau-git: https://en.wikipedia.org/wiki/Furniture_(typesetting)

docling_core/types/experimental/document.py
cau-git: Can you provide an example of how that would apply here? I don't see it yet.

cau-git:
Everyone, many thanks for your in-depth reviews and comments. I am already applying some uncritical changes according to your feedbacks.

One of the bigger topics is the underdeveloped part of `file_info` and `description` fields / models. To address this, I am taking the liberty to redesign the whole part as follows:
- We remove the entire anachronism of `file_info.filename`, and put as new top-level fields:
  - `name: str` # just the working name of the document (could be taken from originating doc, or just "Untitled 1")
  - `origin: FileOrigin` a new field / type to capture the origin of the document object (original filename, mimetype, binary hash, uri). This is optional, since a `DoclingDocument` could be created entirely without an origin (e.g. synthetic data)
  - `hash`: A _computed_ field that hashes the content of the `DoclingDocument` automatically, and is serialized among the rest  
  - entire `file_info` and it's anachronisms disappear.

Futher, on the topic of Elastic indexing: `DoclingDocument` is deliberately not considering any constraints that would be indexing-friendly for ES, neither does it consider specifics of any other downstream asset. There is only a _native_ representation of `DoclingDocument` in JSON/YAML, as pydantic naturally generates from `model_dump`. For any applications, be it RAG, ES indexing, VLM training or other, we should only consider to generate specific, optimized representations as inbuilt _export_ functions. These export formats can be lossy and don't need to carry around attributes to re-construct a `DoclingDocument` from it again. Currently we can export to Markdown and Doctags. An elastic-friendly format should be another sibling to these in my opinion.

docling_core/types/experimental/document.py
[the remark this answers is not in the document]
ceberam: I'm fine to leave it like this and consider it as resolved. But I'd like to point out that:
- the inline conditional assignment alternative fits within the 88 limit
- the assignments can be broken with a `()` after `=` to gain a couple of characters

ceberam:
> @ceberam we have some toxic stuff in the pre-commit pipeline, which removes unused imports also from __init__.py files.

@cau-git  It's an option from autoflake that can be controlled in `pyproject.toml`. By setting it to `false` (default) the unused imports are not removed.

```toml
[tool.autoflake]
remove-all-unused-imports = true
```

In most cases, it doesn't make sense to import libraries that are not used in the module and I find it useful to automatically style the modules accordingly with a pre-commit hook, like we already do with _Black_ and _isort_.

> This must be removed to be able to create a maintainable import system on the package. It seems to come from autoflake.

@cau-git  It can be removed but it does not have to, in order to create that import system on the package. According to PEP8:

>  To better support introspection, modules should explicitly declare the names in their public API using the __all__ attribute. Setting __all__ to an empty list indicates that the module has no public API.

So, for instance, if you want to expose the new `DoclingDocument` in the `types` package, you could add the following lines in the `docling_core/types/__init__.py` file:

```python
from docling_core.types.experimental.document import DoclingDocument

__all__ = ["DoclingDocument"]
```

This passes all the pre-commit hooks, including autoflake and flake8, it is PEP8 compliant, and it does not require any `# noqa` tag. Then in a client application you can import it from the package:

```python
from docling_core.types import DoclingDocument
```
@cau-git @vagenas is this solution satisfactory?

> I found it to go way further, also flake8 and mypy seem to cause issues. Flake8 has options to tame it but apparently ignores what I set in [tool.flake8] anyway.

@cau-git Flake8 options must be set in the separate file`.flake8`. For instance, we exclude the tests in the validation, since it is not critical to have docstring documentation in regression tests.
Let me know which other issues you encountered between flake8 and mypy.

test/test_docling_doc.py
```diff
@@ -0,0 +1,286 @@
...
+    KeyValueItem,
+    SectionItem,
+    PictureItem,
+    TableItem,
+    BasePictureData,
+    BaseTableData
+)
+from docling_core.types.experimental.labels import DocItemLabel, GroupLabel
+
+def test_docitems():
```
PeterStaar-IBM, line 24: @cau-git added a unit-test, so we can detect if new classes appear or we have added the serialisation of a DocItem subclass

docling_core/types/experimental/document.py
```diff
@@ -0,0 +1,1168 @@
...
+    model_config = ConfigDict(extra="forbid")
+
+    def get_ref(self):
+        """get_ref."""
+        return RefItem(cref=self.self_ref)
+
+
+class GroupItem(NodeItem):  # Container type, can't be a leaf node
+    """GroupItem."""
+
```
maxmnemonic, line 223: Add provenance field to GroupItem or NodeItem,
to support provenance of entire lists, or slides (that are collections of elements)
In case of lists - it might be that individual bounding boxes are not possible or hard to obtain, but bounding box of an entire list is readily available. And in any way we would like to keep bounding boxes of lists, for this Group should support provenance.

docling_core/types/experimental/document.py
cau-git: `parent` is using `self.body` as default if unset.

PeterStaar-IBM (dismissed):
Let's merge this!

test/data/experimental/dummy_doc.yaml
```diff
@@ -0,0 +1,177 @@
...
+# All elements that have a text-string representation, type TextItem.
+# This is a flat list of all elements without implied order.
+texts:
+  - orig: "arXiv:2206.01062v1 [cs.CV] 2 Jun 2022"
+    text: "arXiv:2206.01062v1 [cs.CV] 2 Jun 2022"
+    self_ref: "#/texts/0"
+    label: "page_header"
+    parent:
+      $ref: "#/furniture"
+    children: []
```
vagenas, line 42: @cau-git I see something that may be an inconsistency:
- the parent has the `#` prefix (fragment separator for [URI fragment](https://datatracker.ietf.org/doc/html/rfc6901#section-6)): `$ref: "#/furniture"`, while
- the children do not have this prefix, e.g. `- $ref: "/texts/1"`

Can you clarify?

cau-git: Well spotted. The children must have it too.

PeterStaar-IBM: @vagenas Can you do the update?

test/test_docling_doc.py
```diff
@@ -0,0 +1,379 @@
...
+            text="695944",
+        )
+    )
+    table_el = BaseTableData(num_rows=3, num_cols=3, table_cells=table_cells)
+    doc.add_table(data=table_el)
+
+    fig_caption = doc.add_text(
+        label=DocItemLabel.CAPTION, text="This is the caption of figure 1."
+    )
+    fig_item = doc.add_picture(data=BasePictureData(), caption=fig_caption)
```
vagenas, line 335: @cau-git fig_item here is not used.

test/data/experimental/dummy_doc.yaml
cau-git: The fix needed is only to correct manually this file, since it was hand-crafted.
We should also consider validating the `ref` property to ensure it always follows the `#{}/{}` pattern.

vagenas: @PeterStaar-IBM @cau-git
1. validation added (which individually as expected causes the tests to fail due to the invalid data) and
2. test data fixed

PeterStaar-IBM (approved):

maxmnemonic (approved):

dolfim-ibm (approved):

nikos-livathinos (approved):

PR #32, opened 2024-09-25
fix: change order of JSON Schema to search mapper transformations

ceberam:
- In the JSON Schema to search mapper, the suppression flags need to be addressed first, otherwise the `jsonref.replace_refs` function may remove them.
- Updating pydantic from `2.8.2` to `2.9.2` triggers a change in JSON Schema from models: the  lists with only 1 element get flattened. The documentation is automatically updated.

### DIFF
--- a/test/test_json_schema_to_search_mapper.py
+++ b/test/test_json_schema_to_search_mapper.py
@@ -52,8 +52,9 @@ def test_json_schema_to_search_mapper_0():
 
     diff = jsondiff.diff(index_ref, index_def)
     # print(json.dumps(index_def, indent=2))
-    print(diff)
-    assert index_def == index_ref
+    assert (
+        index_def == index_ref
+    ), f"Error in search mappings of ExportedCCSDocument. Difference:\n{json.dumps(diff, indent=2)}"
 
 
 def test_json_schema_to_search_mapper_1():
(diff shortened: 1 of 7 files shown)


dolfim-ibm (dismissed):
LGTM. Let's sync in which order we want to merge.

ceberam:
> LGTM. Let's sync in which order we want to merge.

This PR should be merged first, since the current version of pydantic (`2.9.2`) breaks `main`.

test/test_json_schema_to_search_mapper.py
```diff
@@ -51,7 +51,7 @@ def test_json_schema_to_search_mapper_0():
     index_ref = _load(filename)
 
     diff = jsondiff.diff(index_ref, index_def)
-    print(json.dumps(index_def, indent=2))
+    # print(json.dumps(index_def, indent=2))
     print(diff)
```
PeterStaar-IBM, line 55: can we comment out the diff

test/test_json_schema_to_search_mapper.py
```diff
@@ -51,7 +51,7 @@ def test_json_schema_to_search_mapper_0():
...
     assert index_def == index_ref
```
PeterStaar-IBM, line 56: Can you add a proper error message in case this assert fails

assert index_def == index_ref, "assert index_def != index_ref"

PeterStaar-IBM (changes requested):
see comments

test/test_json_schema_to_search_mapper.py
ceberam: In case of assertion failure, pytest should already provide enough context information but nevertheless I added a commit with more verbose, including the difference.

PeterStaar-IBM (approved):
nice!

cau-git (approved):

PR #37, opened 2024-09-30
fix: align chunk ref format with one used in Document

vagenas:


dolfim-ibm (approved):

docling_core/transforms/chunker/base.py
```diff
@@ -7,24 +7,40 @@
...
+
+
 class Chunk(BaseModel):
     """Data model for Chunk."""
 
     path: str
     text: str
+    heading: Optional[str] = None
+
+    @model_validator(mode="before")
```
ceberam, line 26: I would use a field validator, since you are validating a specific field of a model and you don't need the other data.
In addition, the mode should be `after`, since you can leverage the pydantic control on the type (`str`) first and then continue with your string assumption in `if path.startswith(`

docling_core/transforms/chunker/base.py
```diff
@@ -7,24 +7,40 @@
...
+    @classmethod
+    def _json_pointer_from_json_path(cls, data):
+        path = data.get("path")
+        if path.startswith("$."):
+            parts = path.split("[")
```
ceberam, line 31: Just to be safer and more transparent with the code, I would include the assumption of `[` in the if statement. Or use regex.

docling_core/transforms/chunker/base.py
vagenas: See https://github.com/DS4SD/docling-core/pull/39.

docling_core/transforms/chunker/base.py
vagenas: See https://github.com/DS4SD/docling-core/pull/39.

PR #38, opened 2024-10-07
fix: make doc metadata keys pure strings

vagenas:


dolfim-ibm (approved):

docling_core/transforms/metadata_extractor/simple_metadata_extractor.py
```diff
@@ -6,20 +6,18 @@
...
 
 
-from enum import Enum
 from typing import Any
 
 from docling_core.transforms.metadata_extractor import BaseMetadataExtractor
 from docling_core.types import Document as DLDocument
 
+_DL_DOC_HASH = "dl_doc_hash"
+_ORIGIN = "origin"
```
ceberam, line 15: Since we are using type annotations and for better documentation (global variables that will never change), I would mark these keys with `typing.Final`:
```python
from typing import Final

_DL_DOC_HASH: Final[str] = "dl_doc_hash"
_ORIGIN: Final[str] = "origin"
```

vagenas: See https://github.com/DS4SD/docling-core/pull/39.

PR #42, opened 2024-10-15
feat: v2 doc image

dolfim-ibm:


pyproject.toml
```diff
@@ -53,6 +53,7 @@ jsonref = "^1.1.0"
 json-schema-for-humans = "^1.0.0"
 tabulate = "^0.9.0"
 pandas = "^2.1.4"
+pillow = "^10.3.0"
```
dolfim-ibm, line 56: note: this is the oldest version without security issues

PR #44, opened 2024-10-16
feat: add basic document description fields

ceberam:


PeterStaar-IBM:
Will be followed up with something new.

PR #49, opened 2024-10-18
feat: extend chunk meta with schema, version, origin

vagenas:


docling_core/transforms/chunker/hierarchical_chunker.py
```diff
@@ -49,9 +68,39 @@ class DocMeta(BaseMeta):
...
+        doc_match = re.match(VERSION_PATTERN, v)
+        if (
+            doc_match is None
+            or current_match is None
+            or doc_match["major"] != current_match["major"]
+            or doc_match["minor"] > current_match["minor"]
+        ):
+            raise ValueError(f"incompatible version {v} with schema version {_VERSION}")
+        else:
+            return _VERSION
```
ceberam, line 103: we should reuse the validator in https://github.com/DS4SD/docling-core/blob/main/docling_core/types/doc/document.py#L1418 to avoid code duplication

docling_core/transforms/chunker/hierarchical_chunker.py
```diff
@@ -25,16 +29,31 @@
...
+_KEY_ORIGIN = "origin"
 
 _logger = logging.getLogger(__name__)
 
 
 class DocMeta(BaseMeta):
-    """Data model for Hierarchical Chunker metadata."""
+    """Data model for Hierarchical Chunker chunk metadata."""
 
+    schema_name: Literal["docling_core.transforms.chunker.DocMeta"] = Field(
```
ceberam, line 47: Agree with setting a full path to mark the schema. We may want to do the same with `DoclingDocument` schema name on https://github.com/DS4SD/docling-core/blob/main/docling_core/types/doc/document.py#L784

ceberam (approved):

dolfim-ibm (approved):

docling_core/transforms/chunker/hierarchical_chunker.py
vagenas: Good point, will just need to account for the fact that a validator parameter for target version should be included too as it is different for the two cases (https://github.com/pydantic/pydantic/discussions/2938)

PR #50, opened 2024-10-19
feat: improve export of doclingdocument

PeterStaar-IBM:


docling_core/types/doc/document.py
```diff
@@ -1106,7 +1106,7 @@ def export_to_dict(self) -> Dict:
         """export_to_dict."""
         return self.model_dump(mode="json", by_alias=True, exclude_none=True)
 
-    def export_to_markdown(  # noqa: C901
```
maxmnemonic, line 1109: I wouldn't rename `export_to_markdown` to `export_to_markdown_v1`, to not break compatibility

PeterStaar-IBM: it does not, we have export_to_markdown which points by default to v1

cau-git (approved):

cau-git (approved):

docling_core/types/doc/document.py
```diff
@@ -1146,133 +1150,142 @@ def export_to_markdown(  # noqa: C901
...
+            elif isinstance(item, TextItem) and item.label in [DocItemLabel.CAPTION]:
+                # captions are printed in picture and table ... skipping for now
+                continue
 
-                    # Combine parts
-                    markdown_text = "\n".join(parts)
+            elif isinstance(item, ListItem) and item.label in [DocItemLabel.LIST_ITEM]:
+                in_list = True
+                # Calculate indent based on list_nesting_level
+                indent = "    " * (
```
vagenas, line 1225: Instead of using literals like `"    "` (hard to read), I would rather include an indent argument like `json.dump` for the base indent and use that, e.g.:

```
def export_to_markdown(
  # ...
  base_indent: int = 4,
):
  # ...
  indent = (list_nesting_level - 1) * base_indent * " "
  # ...
```

PeterStaar-IBM: I tend to agree

cau-git: More so, any indentation of less than 4 is not legal and may render broken on some Markdown visualizers.

docling_core/types/doc/document.py
```diff
@@ -1399,6 +1420,121 @@ def export_to_document_tokens(
 
         return doctags
 
+    def _export_to_indented_text(
+        self, indent="  ", max_text_len: int = -1, explicit_tables: bool = False
+    ):
```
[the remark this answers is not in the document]
cau-git: It is used in debugging and private, we keep it.

PR #53, opened 2024-10-24
feat: add headers argument and a custom user-agents for http requests

PeterStaar-IBM:
Currently, when we use the docling cli, we often can not pull the html pages, because we do not have compliant requests with user-agents. This PR fixes that.

### DIFF
--- a/docling_core/utils/file.py
+++ b/docling_core/utils/file.py
@@ -5,7 +5,6 @@
 
 """File-related utilities."""
 
-import random
 import tempfile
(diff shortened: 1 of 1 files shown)


dolfim-ibm (changes requested):
my proposal is:
1. change the default user-agent to `docling/VERSION`
2. allow the user (both library and cli) to customize the headers. similar to what curl, requests, etc allow

dolfim-ibm (dismissed):

PR #66, opened 2024-11-11
fix: fixed selection logic for a slice of the document

dolfim-ibm:
See https://github.com/DS4SD/docling-core/issues/65


PeterStaar-IBM (approved):
Thanks for catching this!

PR #67, opened 2024-11-12
feat: add get_image for all DocItem

sh-gupta:
We want the `image` attribute to be available for `TextItem` to allow exposing equation images. I have moved `image` from `FloatingItem` to `DocItem` so that it becomes part of `TextItem` as well.

Related issue: [https://github.com/DS4SD/docling/issues/299](https://github.com/DS4SD/docling/issues/299)

### DIFF
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -552,7 +552,12 @@ class DocItem(
         return location
 
     def get_image(self, doc: "DoclingDocument") -> Optional[PILImage.Image]:
-        """Returns the image of this DocItem if the document stores page images."""
+        """Returns the image of this DocItem.
+
+        The function returns None if this DocItem has no valid provenance or
+        if a valid image of the page containing this DocItem is not available
+        in doc.
+        """
         if not len(self.prov):
             return None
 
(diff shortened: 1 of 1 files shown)


vagenas (changes requested):
@sh-gupta @cau-git ⚠️  note that any changes to the subtypes of DoclingDocument:
1. Need to be backwards compatible within docling-core 2.x.y, e.g. new field with default value, and
2. then imply a new minor version bump of the DoclingDocument as well as potentially DocMeta which also depends on the DocItem.

PeterStaar-IBM:
@sh-gupta We quickly discussed, we need to do the following points (@cau-git will explain),

- add get_image to DocItem (overloaded on PictureItem)
	- check if there are page-images
- enrichment models can leverage the backend
- deprecate this: https://github.com/DS4SD/docling/blob/main/docling/datamodel/pipeline_options.py#L84C5-L84C28

cau-git:
> Made the required changes:
> 
> 1. Moved `image` from `DocItem` back to `FloatingItem`
> 2. Added `get_image` function to `DocItem`. If returns `None` if provenance is not found or page image is not available, else it returns an appropriately cropped section of the page image
> 3. Overrode `get_image` in `FloatingItem`

This looks good now, thanks!

PeterStaar-IBM:
@sh-gupta Could we add some unit-tests for this functionality?

dolfim-ibm (approved):

PeterStaar-IBM (approved):
very nice!

sh-gupta:
@PeterStaar-IBM : I've added tests for `DocItem.get_image` and `FloatingItem.get_image`

@dolfim-ibm : I initially thought raising an error but then decided to stick with `None` to be consistent with the behavior of `FloatingItem.image` when `image` is not found. I can raise an exception if you prefer.

@vagenas : I've made the changes as per your discussion. Can you please take a look?

PeterStaar-IBM (approved):
Lovely! Time to get this in!

vagenas:
@sh-gupta in the docstring of the two `get_image()` methods, I would explicitly add that the method returns `None` if no image is available / applicable (exact wording up to you, but just to mention that `None` is returned "otherwise").

PeterStaar-IBM (approved):
super!

dolfim-ibm (approved):

PR #68, opened 2024-11-19
feat: add hybrid chunker

vagenas:
#### Planned for this PR
- [x] last chunk is disregarded due to bug in `_merge_chunks_with_matching_metadata()`
- [x] introduced transformers as a docling-core dependency
- [x] `_split_by_doc_items()` currently assumes `DocItem` has a `.text` i.e. assumes `TextItem`
- [x] unit tests needed

#### TBD
- [ ] we allow too many tokens when using another delimiter like "####"
- [ ] table splitting currently text-based, can be optimized, e.g. based on row or cell
- [ ] duplication of serialization logic in new chunker implem and BaseChunker `serialize()`; former should reuse latter
- [ ] general refactoring of new chunker implementation needed
- [ ] we currently indeed split in the middle of sentences; clarify if issue or not
- [ ] use of external library semchunk; to be clarified
- [ ] minor: streaming is superficial


pyproject.toml
```diff
@@ -55,6 +55,8 @@ pandas = "^2.1.4"
 pillow = "^10.3.0"
 pyyaml = ">=5.1,<7.0.0"
 typing-extensions = "^4.12.2"
+transformers = "^4.46.3"
+semchunk = "^2.2.0"
```
dolfim-ibm, line 59: we might need to move these to extras.
for example, if `datasets` would use `docling-core`, they would not accept this dependency.

vagenas: These have now been moved to an extra.

vagenas:
Converting to draft until all points planned for this PR are addressed (see description)..

vagenas:
Main changes with this PR:
1. **Adds a new chunker**, currently named **`TokenAwareChunker`**, which uses a hybrid approach, applying tokenization-aware refinements on top of document-layout-based (AKA hierarchical) chunking. More precisely:
    - it starts from the result of the hierarchical chunker and, based on the input tokenizer (typically to be aligned to the embedding model tokenizer), it:
    - does one pass where it splits chunks only when needed (i.e. oversized w.r.t. tokens), &
    - another pass where it merges chunks only when possible (i.e. undersized successive chunks with same headings & captions) — users can opt out of this step via param `merge_peers` (by default True)
2. **Extends `BaseChunker` interface** with a new method, currently named **`serialize()`**, which returns the enriched chunk serialization, by default being a concatenation of the relevant metadata (headings & captions) with the chunk's `.text`. This is the representation that should meet the token limitations applied by `TokenAwareChunker`.

👉 Any better naming ideas, particularly for the new chunker, are welcome. Some options in no particular order:
- `HybridChunker`
- `TokenizedLayoutChunker`
- `TokenizedHierarchicalChunker` (although it's technically not a subclass of `HierarchicalChunker`)
- `AdaptiveChunker`

dolfim-ibm (approved):

dolfim-ibm (approved):

vagenas:
Finally went for `HybridChunker` as that appears to be the standard industry term for this approach, see e.g.:
- https://www.ibm.com/architectures/papers/rag-cookbook/chunking
- https://zilliz.com/learn/guide-to-chunking-strategies-for-rag

PR #69, opened 2024-11-20
feat: adding HTML export to DoclingDocument

PeterStaar-IBM:


docling_core/types/doc/document.py
```diff
@@ -677,6 +678,89 @@ class PictureItem(FloatingItem):
 
     annotations: List[PictureDataType] = []
 
+    # Convert the image to Base64
+    def _image_to_base64(self, pil_image, format="PNG"):
```
PeterStaar-IBM, line 682: I am not sure if this is really needed (@dolfim-ibm @cau-git @vagenas )

docling_core/types/doc/document.py
```diff
@@ -1097,28 +1192,41 @@ def add_text(
...
-        if not parent:
-            parent = self.body
+        # Catch a few cases that are in principle allowed
+        # but that will create confusion down the road
+        if label in [DocItemLabel.TITLE]:
+            return self.add_title(text=text, orig=orig, prov=prov, parent=parent)
 
-        if not orig:
-            orig = text
+        elif label in [DocItemLabel.LIST_ITEM]:
```
PeterStaar-IBM, line 1200: mitigating potential mistakes a user could do ...

docling_core/types/doc/document.py
```diff
@@ -1097,28 +1192,41 @@ def add_text(
...
-        text_item = TextItem(
-            label=label,
-            text=text,
-            orig=orig,
-            self_ref=cref,
-            parent=parent.get_ref(),
-        )
-        if prov:
-            text_item.prov.append(prov)
+        elif label in [DocItemLabel.SECTION_HEADER]:
```
PeterStaar-IBM, line 1203: mitigating potential mistakes a user could do ...

docling_core/types/doc/document.py
```diff
@@ -677,6 +678,89 @@ class PictureItem(FloatingItem):
...
+        )  # Encode to Base64 and decode to string
+        return img_base64
+
+    def export_to_markdown(
+        self,
+        doc: "DoclingDocument",
+        add_caption: bool = True,
+        image_placeholder: str = "<!-- image -->",
+        image_mode: ImageRefMode = ImageRefMode.EMBEDDED,
+        filename: Optional[Path] = None,
```
cau-git, line 697: Putting a filename argument here as optional argument is not the way export methods should work. Current export methods only return an in-memory representation and do _not_ write to file-system. See other comment on `export_to_dict`.

docling_core/types/doc/base.py
```diff
@@ -12,6 +12,7 @@ class ImageRefMode(str, Enum):
 
     PLACEHOLDER = "placeholder"
     EMBEDDED = "embedded"
+    EXPORTED_TO_PNG = "exported_to_png"
```
cau-git, line 15: I don't think `EXPORTED_TO_PNG` is a good name, this must be a generic term for externally referenced files. It should be equally valid to put a web URI or other protocol for this case, and the image format must not be part of it. Depending on what future backends give us it could also be used for a native image format in the source doc, e.g. SVG or other.

Proposals:
- `URI`
- `FILE_URI`

docling_core/types/doc/document.py
```diff
@@ -1321,6 +1480,7 @@ def export_to_markdown(  # noqa: C901
         indent: int = 4,
         text_width: int = -1,
         page_no: Optional[int] = None,
+        image_dir: Optional[Path] = None,
```
cau-git, line 1483: It is not a proper solution to provide a path argument for the `image_dir`. This would mean:
- The `image_dir` is being written to by the export method straight
- The exported document representation (i.e. Dict, MD string, HTML string) is returned by the export method but not written to the same place, and you make it the users responsibility to then save the exported object where the images are.

docling_core/types/doc/document.py
```diff
@@ -677,6 +678,89 @@ class PictureItem(FloatingItem):
...
+                "Please use `PdfPipelineOptions(generate_picture_images=True)`"
+                " --> "
+            )
+            return text
+
+    def export_to_html(
+        self,
+        doc: "DoclingDocument",
+        add_caption: bool = True,
+        filename: Optional[Path] = None,
```
cau-git, line 735: As with `export_to_markdown`, putting a filename argument here is not a proper solution.

docling_core/types/doc/document.py
```diff
@@ -1518,6 +1699,179 @@ def export_to_text(  # noqa: C901
             image_placeholder="",
         )
 
+    def export_to_html(  # noqa: C901
+        self,
+        from_element: int = 0,
+        to_element: int = sys.maxsize,
+        labels: set[DocItemLabel] = DEFAULT_EXPORT_LABELS,
+        image_dir: Optional[Path] = None,
```
cau-git, line 1707: Same comment as for `export_to_dict`, we cannot simply have an `image_dir` argument in here.

cau-git (changes requested):
@PeterStaar-IBM Great to get HTML export, but I see some fundamental design issues w.r.t. to handling image and document export to files by adding arguments.

cau-git (changes requested):
Proposal:

```python
from pathlib import Path

from docling_core.types import DoclingDocument
from docling_core.types.doc import ImageRefMode
from test.test_docling_doc import _construct_doc

doc: DoclingDocument = _construct_doc()

# Hard to understand, many ways to make mistakes:
# 1. To export images and MD referencing the images, the two methods below must be exactly called in this order.
#    What if a user decides to call in reverse order?
# 2. What does a call to `export_to_markdown(image_mode=ImageRefMode.REFERENCED)` do when
#    `doc.save_images_to_disk(...)` wasn't called earlier?
# 3. Why `save_images_to_disk` creates a new DoclingDocument instance? Users could miss it and keep using `doc`
# 4. How would relative file paths be used for the images like this?

new_doc = doc.save_images_to_disk(image_dir=Path("./test/data/constructed_images/"))
md_pred = new_doc.export_to_markdown(image_mode=ImageRefMode.REFERENCED)

with Path("./out.md").open('w') as fp:
    fp.write(md_pred)

# -------------- Alternative solution ---------------------------------------
# For export formats, one can not save images without saving the MD/HTML/JSON alongside, only one path argument.

# To just save markdown/HTML/whatever with referenced images:
doc.save_as_markdown("out.md", image_mode=ImageRefMode.REFERENCED) # image_mode has default, internally calls doc.export_to_markdown
doc.save_as_html("out.html", image_mode=ImageRefMode.REFERENCED) # image_mode has default, internally calls doc.export_to_html

# To just turn the document into one with all images as references
new_doc = doc.with_image_references(image_dir="...")

# Additional: To just turn the document into one with all images as references
new_doc = doc.with_embedded_images()

doc.save_as_json("out.json") # internally calls doc.export_to_dict
doc.save_as_json("out.json",
                 image_mode=ImageRefMode.REFERENCED) #optional image_mode, internally calls doc.export_to_dict



```

Open points:
- How do we control _which_ document items are exported as images? Need additional parameters as with labels.

docling_core/types/doc/document.py
```diff
@@ -1297,17 +1581,241 @@ def iterate_items(
...
+                    item.image is not None
+                    and isinstance(item.image.uri, AnyUrl)
+                    and str(item.image.uri).startswith("file:")
+                ):
+                    local_path = Path(str(item.image.uri).replace("file:", ""))
+                    result.append(local_path)
+
+        return result
+
+    def _with_embedded_pictures(self) -> "DoclingDocument":
```
dolfim-ibm, line 1610: [optional] this could be extended with an argument `fetch_remote=True`, which will download the remote urls

docling_core/types/doc/document.py
```diff
@@ -981,6 +1200,23 @@ class PageItem(BaseModel):
 class DoclingDocument(BaseModel):
     """DoclingDocument."""
 
+    HTML_DEFAULT_HEAD: str = r"""<head>
```
dolfim-ibm, line 1203: make this `_HTML_DEFAULT_HEAD`, then we don't need any weird delete logic below

docling_core/types/doc/document.py
```diff
@@ -1297,17 +1581,241 @@ def iterate_items(
...
+        mode: str = "json",
+        by_alias: bool = True,
+        exclude_none: bool = True,
+    ) -> Dict:
+        """Export to dict."""
+        out = self.model_dump(mode=mode, by_alias=by_alias, exclude_none=exclude_none)
+
+        for key in ["HTML_DEFAULT_HEAD"]:
+            if key in out:
+                del out[key]
```
dolfim-ibm, line 1767: not needed when using proper private properties (see comment above)

docling_core/types/doc/document.py
```diff
@@ -1297,17 +1581,241 @@ def iterate_items(
...
+
+        new_doc: "DoclingDocument" = DoclingDocument(name=self.name)
+        if image_mode == ImageRefMode.PLACEHOLDER:
+            new_doc = self
+        elif image_mode == ImageRefMode.REFERENCED:
+            new_doc = self._with_pictures_refs(image_dir=artifacts_dir)
+        elif image_mode == ImageRefMode.EMBEDDED:
+            new_doc = self._with_embedded_pictures()
+        else:
+            raise ValueError("Unsupported ImageRefMode")
```
dolfim-ibm, line 1715: it would be better to factor out this block which we are now repeating >3 times

dolfim-ibm (changes requested):

docling_core/types/doc/document.py
```diff
@@ -1297,17 +1581,241 @@ def iterate_items(
                         page_no=page_no,
                     )
 
+    def clear_cache(self):
+        """Clear cache storage of all images."""
+        for ix, (item, level) in enumerate(self.iterate_items(with_groups=True)):
```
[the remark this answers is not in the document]
cau-git: This should be either a private API method or work generically, the name `clear_cache` does not imply it is about pictures.

cau-git (approved):

docling_core/types/doc/document.py
```diff
@@ -677,6 +734,152 @@ class PictureItem(FloatingItem):
...
+                img_text = f'<img src="data:image/png;base64,{imgb64}">'
+
+                return f"<figure>{caption_text}{img_text}</figure>"
+            else:
+                return default_response
+
+        elif image_mode == ImageRefMode.REFERENCED:
+
+            if not isinstance(self.image, ImageRef) or (
+                isinstance(self.image.uri, AnyUrl) and self.image.uri.scheme == "data"
```
dolfim-ibm, line 867: this is matching the case of calling `export_to_html()` with `ImageRefMode.REFERENCED` instead of `save_as_html()`.

should we signal it to the user with a warning instead of "silently" put placeholder in the output?

docling_core/types/doc/document.py
```diff
@@ -1297,17 +1580,216 @@ def iterate_items(
...
+            if isinstance(item, PictureItem):
+
+                if item.image is not None and (
+                    (
+                        isinstance(item.image.uri, AnyUrl)
+                        and item.image.uri.scheme == "file"
+                    )
+                    or isinstance(item.image.uri, Path)
+                ):
+                    tmp_image = PILImage.open(str(item.image.uri))
```
dolfim-ibm, line 1627: this won't work when the uri is `file://path...`

docling_core/types/doc/document.py
```diff
@@ -1297,17 +1580,216 @@ def iterate_items(
...
+                    item.image = ImageRef.from_pil(tmp_image, dpi=item.image.dpi)
+
+        return result
+
+    def _with_pictures_refs(
+        self, image_dir: Path, reference_path: Optional[Path] = None
+    ) -> "DoclingDocument":
```
dolfim-ibm, line 1634: this is currently ignoring remote figures, which remains as remote uri. is it what we want?
the docstring seems to imply that all images will be local files after calling this function.

dolfim-ibm (changes requested):

docling_core/types/doc/document.py
```diff
@@ -677,6 +730,157 @@ class PictureItem(FloatingItem):
...
+            else:
+                return error_response
+
+        elif image_mode == ImageRefMode.REFERENCED:
+
+            if (
+                isinstance(self.image, ImageRef)
+                and isinstance(self.image.uri, AnyUrl)
+                and str(self.image.uri).startswith("file:")
+                and True  # self.image.uri.exists()
```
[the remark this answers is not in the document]
cau-git: Resolved.

docling_core/types/doc/document.py
```diff
@@ -677,6 +730,157 @@ class PictureItem(FloatingItem):
...
+            else:
+                return default_response
+
+        elif image_mode == ImageRefMode.REFERENCED:
+
+            if (
+                isinstance(self.image, ImageRef)
+                and isinstance(self.image.uri, AnyUrl)
+                and str(self.image.uri).startswith("file:")
+                and True  # self.image.uri.exists()
```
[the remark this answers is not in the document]
cau-git: Resolved.

docling_core/types/doc/document.py
```diff
@@ -1297,17 +1581,241 @@ def iterate_items(
...
+                if item.image is not None and item.image._pil is not None:
+                    item.image._pil.close()
+
+    def _list_images_on_disk(self) -> List[Path]:
+        """List all images on disk."""
+        result: List[Path] = []
+
+        for ix, (item, level) in enumerate(self.iterate_items(with_groups=True)):
+            if isinstance(item, PictureItem):
+                if item.image is not None and isinstance(item.image.uri, Path):
```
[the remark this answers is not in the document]
cau-git: Resolved.

docling_core/types/doc/document.py
cau-git: Resolved.

docling_core/types/doc/document.py
cau-git: Factored it out now.

docling_core/types/doc/document.py
```diff
@@ -1297,17 +1581,241 @@ def iterate_items(
...
+        self,
+        filename: Path,
+        artifacts_dir: Optional[Path] = None,
+        image_mode: ImageRefMode = ImageRefMode.EMBEDDED,
+        by_alias: bool = True,
+        exclude_none: bool = True,
+        default_flow_style: bool = False,
+    ):
+        """Save as yaml."""
+        if artifacts_dir is None:
```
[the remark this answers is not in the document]
cau-git: This is covered.

docling_core/types/doc/document.py
```diff
@@ -1297,17 +1581,241 @@ def iterate_items(
...
+        if image_mode == ImageRefMode.PLACEHOLDER:
+            new_doc = self
+        elif image_mode == ImageRefMode.REFERENCED:
+            new_doc = self._with_pictures_refs(image_dir=artifacts_dir)
+        elif image_mode == ImageRefMode.EMBEDDED:
+            new_doc = self._with_embedded_pictures()
+        else:
+            raise ValueError("Unsupported ImageRefMode")
+
+        out = new_doc.export_to_dict(
```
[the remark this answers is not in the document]
cau-git: This is covered.

docling_core/types/doc/document.py
```diff
@@ -1297,17 +1581,241 @@ def iterate_items(
...
```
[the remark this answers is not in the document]
cau-git: This is fixed.

docling_core/types/doc/document.py
cau-git: We need to think this through further, to avoid that by default we create a dependency on network resources when calling this function. How to signal it to the user is an open question.

docling_core/types/doc/document.py
```diff
@@ -1297,17 +1581,241 @@ def iterate_items(
...
+        for ix, (item, level) in enumerate(self.iterate_items(with_groups=True)):
+            if isinstance(item, PictureItem):
+                if item.image is not None and item.image._pil is not None:
+                    item.image._pil.close()
+
+    def _list_images_on_disk(self) -> List[Path]:
+        """List all images on disk."""
+        result: List[Path] = []
+
+        for ix, (item, level) in enumerate(self.iterate_items(with_groups=True)):
```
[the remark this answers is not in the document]
cau-git: Resolved.

docling_core/types/doc/document.py
cau-git: Fixed.

docling_core/types/doc/document.py
cau-git: We should but I don't know how that would best be done.

docling_core/types/doc/document.py
```diff
@@ -449,10 +456,12 @@ def pil_image(self) -> PILImage.Image:
...
             self._pil = PILImage.open(BytesIO(decoded_img))
-        else:
-            self._pil = PILImage.open(str(self.uri))
+        elif str(self.uri).startswith("file:"):
+            parsed_url = urlparse(str(self.uri))
+            self._pil = PILImage.open(unquote(parsed_url.path))
+        # else: Handle http request or other protocols...
 
-        return self._pil
+        return self._pil  # type: ignore # mypy can't resolve this.
```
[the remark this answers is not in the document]
cau-git: I think for the time being we must accept that the function returns `None`, and we establish how to fetch resources from a network in a future iteration.

cau-git (dismissed):

cau-git (approved):

PR #70, opened 2024-11-21
fix: temporarily force pydantic < 2.10

vagenas:
For https://github.com/DS4SD/docling/issues/406.


PeterStaar-IBM (approved):
good temp fix! We will need to solve it properly though

vagenas:
> good temp fix! We will need to solve it properly though

@PeterStaar-IBM I would keep this constraint until the [various known Pydantic 2.10 issues](https://github.com/pydantic/pydantic/issues/10910) are fixed upstream 😉

PeterStaar-IBM:
> > good temp fix! We will need to solve it properly though
> 
> @PeterStaar-IBM I would keep this constraint until the [various known Pydantic 2.10 issues](https://github.com/pydantic/pydantic/issues/10910) are fixed upstream 😉

💯

PR #71, opened 2024-11-21
feat: simple method to load DoclingDocument from .json files

Eric-Musa:
I wrote utility functions in a script to save and load `DoclingDocument` objects to file:
```python
def save_document(document: DoclingDocument, document_json_path: Union[str, Path]):
        with open(document_json_path, 'w') as f:
            f.write(document.model_dump_json(indent=4))

def load_document(document_json_path: Union[str, Path]) -> DoclingDocument:
    assert os.path.exists(document_json_path), f"Document JSON file not found: {document_json_path}"
    with open(document_json_path, 'r') as f:
        json_data = json.load(f)
    return DoclingDocument(**json_data)
```

Which saved a lot of time when testing workflows with very long, complex documents. Not needing to reprocess the documents each time was very useful, but I did not see a builtin tool to save to a file directly (like `pandas.DataFrame.to_csv`), so I added the following:
```python
class DoclingDocument(BaseModel):
...
    def save_to_json_file(self, document_json_path, indent: int=4) -> int:
        """export_to_json."""
        with open(document_json_path, 'w') as f:
            return f.write(self.model_dump_json(indent=indent))
    
    @classmethod
    def load_from_json_file(cls, document_json_path):
        """load_from_json."""
        with open(document_json_path, 'r') as f:
            return cls(**json.load(f))
...
```

I think adding some simple built-in functions may help save people the time in figuring out the best way to save and load processed documents!

### DIFF
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -1,8 +1,6 @@
 """Models for the Docling Document data type."""
 
 import base64
-import copy
-import hashlib
 import json
 import mimetypes
 import os
@@ -1738,112 +1736,20 @@ class DoclingDocument(BaseModel):
             elif isinstance(item, DocItem):
                 print(" " * level, f"{ix}: {item.label.value}")
 
-    def export_to_element_tree(self) -> str:
-        """Export_to_element_tree."""
-        texts = []
-        for ix, (item, level) in enumerate(self.iterate_items(with_groups=True)):
-            if isinstance(item, GroupItem):
-                texts.append(
-                    " " * level + f"{ix}: {item.label.value} with name={item.name}"
-                )
-            elif isinstance(item, DocItem):
-                texts.append(" " * level + f"{ix}: {item.label.value}")
+    def export_to_dict(self) -> Dict:
+        """export_to_dict."""
+        return self.model_dump(mode="json", by_alias=True, exclude_none=True)
 
-        return "\n".join(texts)
-
-    def save_as_json(
-        self,
-        filename: Path,
-        artifacts_dir: Optional[Path] = None,
-        image_mode: ImageRefMode = ImageRefMode.EMBEDDED,
-        indent: int = 2,
-    ):
-        """Save as json."""
-        artifacts_dir, reference_path = self._get_output_paths(filename, artifacts_dir)
-
-        if image_mode == ImageRefMode.REFERENCED:
-            os.makedirs(artifacts_dir, exist_ok=True)
-
-        new_doc = self._make_copy_with_refmode(
-            artifacts_dir, image_mode, reference_path=reference_path
-        )
-
-        out = new_doc.export_to_dict()
-        with open(filename, "w") as fw:
-            json.dump(out, fw, indent=indent)
-
-    def save_as_yaml(
-        self,
-        filename: Path,
-        artifacts_dir: Optional[Path] = None,
-        image_mode: ImageRefMode = ImageRefMode.EMBEDDED,
-        default_flow_style: bool = False,
-    ):
-        """Save as yaml."""
-        artifacts_dir, reference_path = self._get_output_paths(filename, artifacts_dir)
-
-        if image_mode == ImageRefMode.REFERENCED:
-            os.makedirs(artifacts_dir, exist_ok=True)
-
-        new_doc = self._make_copy_with_refmode(
-            artifacts_dir, image_mode, reference_path=reference_path
-        )
-
-        out = new_doc.export_to_dict()
-        with open(filename, "w") as fw:
-            yaml.dump(out, fw, default_flow_style=default_flow_style)
-
-    def export_to_dict(
-        self,
-        mode: str = "json",
-        by_alias: bool = True,
-        exclude_none: bool = True,
-    ) -> Dict:
-        """Export to dict."""
-        out = self.model_dump(mode=mode, by_alias=by_alias, exclude_none=exclude_none)
-
-        return out
-
-    def save_as_markdown(
-        self,
-        filename: Path,
-        artifacts_dir: Optional[Path] = None,
-        delim: str = "\n",
-        from_element: int = 0,
-        to_element: int = sys.maxsize,
-        labels: set[DocItemLabel] = DEFAULT_EXPORT_LABELS,
-        strict_text: bool = False,
-        image_placeholder: str = "<!-- image -->",
-        image_mode: ImageRefMode = ImageRefMode.PLACEHOLDER,
-        indent: int = 4,
-        text_width: int = -1,
-        page_no: Optional[int] = None,
-    ):
-        """Save to markdown."""
-        artifacts_dir, reference_path = self._get_output_paths(filename, artifacts_dir)
-
-        if image_mode == ImageRefMode.REFERENCED:
-            os.makedirs(artifacts_dir, exist_ok=True)
-
-        new_doc = self._make_copy_with_refmode(
-            artifacts_dir, image_mode, reference_path=reference_path
-        )
-
-        md_out = new_doc.export_to_markdown(
-            delim=delim,
-            from_element=from_element,
-            to_element=to_element,
-            labels=labels,
-            strict_text=strict_text,
-            image_placeholder=image_placeholder,
-            image_mode=image_mode,
-            indent=indent,
-            text_width=text_width,
-            page_no=page_no,
-        )
-
-        with open(filename, "w") as fw:
-            fw.write(md_out)
+    def save_to_json_file(self, document_json_path, indent: int=4) -> int:
+        """export_to_json."""
+        with open(document_json_path, 'w') as f:
+            return f.write(self.model_dump_json(indent=indent))
+    
+    @classmethod
+    def load_from_json_file(cls, document_json_path):
+        """load_from_json."""
+        with open(document_json_path, 'r') as f:
+            return cls(**json.load(f))
 
     def export_to_markdown(  # noqa: C901
         self,
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -1741,11 +1741,13 @@ class DoclingDocument(BaseModel):
         """export_to_dict."""
         return self.model_dump(mode="json", by_alias=True, exclude_none=True)
 
-    def save_to_json_file(self, path: Union[str, Path], indent: int=4) -> int:
+    def save_to_json_file(self, path: Union[str, Path], indent: int=4):
         """
         export_to_json.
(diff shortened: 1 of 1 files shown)


docling_core/types/doc/document.py
```diff
@@ -1309,6 +1310,17 @@ def export_to_dict(self) -> Dict:
         """export_to_dict."""
         return self.model_dump(mode="json", by_alias=True, exclude_none=True)
 
+    def save_to_json_file(self, document_json_path, indent: int=4) -> int:
```
PeterStaar-IBM, line 1313: could you rename `document_json_path` to path and add the type `Path`?

docling_core/types/doc/document.py
```diff
@@ -1309,6 +1310,17 @@ def export_to_dict(self) -> Dict:
...
+        """export_to_json."""
+        with open(document_json_path, 'w') as f:
+            return f.write(self.model_dump_json(indent=indent))
+    
+    @classmethod
+    def load_from_json_file(cls, document_json_path):
```
PeterStaar-IBM, line 1319: could you rename `document_json_path` to path and add the type `Path` and add the return type `-> DoclingDocument`.

I would also remove the assert.

PeterStaar-IBM (changes requested):
Can you do some quick updates?

docling_core/types/doc/document.py
```diff
@@ -1309,6 +1310,17 @@ def export_to_dict(self) -> Dict:
         """export_to_dict."""
         return self.model_dump(mode="json", by_alias=True, exclude_none=True)
 
+    def save_to_json_file(self, document_json_path, indent: int=4) -> int:
+        """export_to_json."""
+        with open(document_json_path, 'w') as f:
+            return f.write(self.model_dump_json(indent=indent))
```
cau-git, line 1316: Please don't use `model_dump_json` here, but something like `json.dumps(self.export_to_dict())`. The `export_to_dict` method already ensures that important pydantic serialization options are set.

docling_core/types/doc/document.py
```diff
@@ -1309,6 +1310,17 @@ def export_to_dict(self) -> Dict:
...
+    
+    @classmethod
+    def load_from_json_file(cls, document_json_path):
+        """load_from_json."""
+        with open(document_json_path, 'r') as f:
+            return cls(**json.load(f))
```
cau-git, line 1322: Please use the official pydantic way to load a model from JSON, which would be:
```
cls.model_validate_json(f.read())
```

see https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_validate_json

cau-git (changes requested):
Thanks for adding these convenience methods, please check the few remarks below.

docling_core/types/doc/document.py
Eric-Musa: Done, thank you!

docling_core/types/doc/document.py
Eric-Musa: Done, thank you!

docling_core/types/doc/document.py
Eric-Musa: Done, thank you!

PeterStaar-IBM:
@Eric-Musa Can you run (multiple times) the following command,

```sh
poetry run pre-commit run --all-files
```

you should see,

```sh
poetry run pre-commit run --all-files
Black....................................................................Passed
isort....................................................................Passed
autoflake................................................................Passed
MyPy.....................................................................Passed
Flake8...................................................................Passed
Pytest...................................................................Passed
Docs.....................................................................Passed
Poetry...................................................................Passed
```

docling_core/types/doc/document.py
```diff
@@ -1738,112 +1736,35 @@ def print_element_tree(self):
...
-            if isinstance(item, GroupItem):
-                texts.append(
-                    " " * level + f"{ix}: {item.label.value} with name={item.name}"
-                )
-            elif isinstance(item, DocItem):
-                texts.append(" " * level + f"{ix}: {item.label.value}")
-
-        return "\n".join(texts)
-
-    def save_as_json(
```
PeterStaar-IBM, line 1754: we should not remove any save_as_... methods

docling_core/types/doc/document.py
```diff
@@ -1738,112 +1736,35 @@ def print_element_tree(self):
...
-        )
+    def export_to_dict(self) -> Dict:
+        """export_to_dict."""
+        return self.model_dump(mode="json", by_alias=True, exclude_none=True)
 
-        out = new_doc.export_to_dict()
-        with open(filename, "w") as fw:
-            json.dump(out, fw, indent=indent)
-
-    def save_as_yaml(
```
PeterStaar-IBM, line 1775: we should not remove any save_as_... methods

docling_core/types/doc/document.py
```diff
@@ -1738,112 +1736,35 @@ def print_element_tree(self):
...
-        mode: str = "json",
-        by_alias: bool = True,
-        exclude_none: bool = True,
-    ) -> Dict:
-        """Export to dict."""
-        out = self.model_dump(mode=mode, by_alias=by_alias, exclude_none=exclude_none)
-
-        return out
-
-    def save_as_markdown(
```
PeterStaar-IBM, line 1807: we should not remove any save_as_... methods

docling_core/types/doc/document.py
```diff
@@ -1738,112 +1736,35 @@ def print_element_tree(self):
...
-
-        new_doc = self._make_copy_with_refmode(
-            artifacts_dir, image_mode, reference_path=reference_path
-        )
-
-        out = new_doc.export_to_dict()
-        with open(filename, "w") as fw:
-            yaml.dump(out, fw, default_flow_style=default_flow_style)
-
-    def export_to_dict(
```
PeterStaar-IBM, line 1796: we should not remove any export_to_... methods

PeterStaar-IBM (changes requested):
Please do not remove any save_as_... or export_to methods. Those need to be kept.

It would be great to have for many `save_as_` have an equivalent `load_from_...`

Eric-Musa:
@PeterStaar-IBM will do, actually running into some issues deserializing image URIs. Am in the middle of writing a pretty long comment 😅. 

Tldr; `DoclingDocument.model_validate_json(f.read())` is breaking embedded base64-encoded images. `DoclingDocument(**json.load(f))` is a functional fix, but it goes against what @cau-git suggested earlier...

---

Testing DoclingDocument.load_from_json_file to load a saved Document that includes images, trying to export the loaded document to markdown results in the following error and traceback:
```
  File "<stdin>", line 1, in <module>
  File "<string>", line 51, in <module>
  File ".venv/lib/python3.12/site-packages/docling_core/types/doc/document.py", line 1923, in export_to_markdown
    line = item.export_to_markdown(
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/docling_core/types/doc/document.py", line 795, in export_to_markdown
    img = self.get_image(doc)
          ^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/docling_core/types/doc/document.py", line 726, in get_image
    return self.image.pil_image
           ^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/docling_core/types/doc/document.py", line 464, in pil_image
    self._pil = PILImage.open(self.uri)
                ^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/PIL/Image.py", line 3431, in open
    fp = builtins.open(filename, "rb")
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: [Errno 63] File name too long: 'Documents/Projects/Python/docling-subsupscript-examples/data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAF4AAABoCAIAAAD3k9CJAABH/UlEQVR4nO29Z3iTV7Y2/Kj3Xi25yJab3LsBGxfAmGIw1fSSQkJCyGTSmCRDeiZh0khmQkINhBZMNRhwsDHu3cZdLpItN8kqVu/1+7HzOhxCMmTmnO96r+876wcXkvez9TxLe++11r3utQT5/pvEarV6vV6XywVeer3eB/+q1WqNRqPP5/N4PLNjurq6wJuzolQq7969O/vSbrfbbLaJiYmHhgFxuVxarXb2/3a73e122+12q9Xq8/kcDoder58d7Ha7/+gTIaHfFYvF4vP5iETibw1wuVxwOLy4uNhsNvv7+zudTpfLlZ6eLhaLCQRCRkYGBEFdXV13795dtWoViUSyWq2HDx9etGgRl8utqqq6fft2Wlpabm5ueXk5m83W6XSTk5M2mw2Hw/X390skEqvV2tTUlJiYuGLFChqNdvHixaSkpNDQUKPReOPGjc7Ozj179shkMo
...
```

This is happening because the DoclingDocument's `ImageRef.uri` attribute is a pydantic `AnyUrl` type when created in memory by a DocumentConverter, but it is a pathlib `PosixPath` when loaded from .json file. Both types are valid according to the ImageRef model, but they have different ways of returning the associated PIL image depending on the URI type ([document.py line 442](https://github.com/DS4SD/docling-core/blob/main/docling_core/types/doc/document.py#L442)).

The `load_from_json_file` function in this PR is essentially just `DoclingDocument.model_validate_json(f.read())`. But there is no issue with the URI if I load the data using `DoclingDocument(**json.load(f))` instead which under the hood is `DoclingDocument.model_validate_python(json.load(f))`.

I believe this is an issue with double forward slashes "//" present in image URIs from the DocumentConverter output, but the double slashes are removed I think upon using `model_validate_json`. Somewhere in the deserialization process, the URI loaded from JSON is being validated as the `Path` type instead of the `AnyUrl` type it is originally serialized from, I _think_ because of the "//"s having been removed.

Please advise!

Here is my test PDF and output JSON:
- [research paper text.pdf](https://github.com/user-attachments/files/17967951/research.paper.text.pdf)
- [research paper text.json](https://github.com/user-attachments/files/17967952/research.paper.text.json)

docling_core/types/doc/document.py
Eric-Musa: I apologize for this, definitely not intentional, git has a way of making me do foolish things. Working to restore this code to this branch.

cau-git:
Hi @Eric-Musa  thanks for your detailed findings, this deserialization is infact important to mail down. I will look into this to see what can be done about model_validate_json not correctly populating the uri.

cau-git:
@Eric-Musa Confirming the issue you found and rolling the code back to what works.

cau-git (approved):

PR #79, opened 2024-11-29
feat: extend source resolution with streams and workdir

dolfim-ibm:
Extension to the resolution of file source
1. Add `workdir` to `resolve_file_source()` by keeping it backwards compatible. 
2. Add a new version `read_file_source()` which returns a document stream to completely avoid writing files

This will be used in Docling to avoid having leftover files in the temporary dir.

### DIFF
--- a/docling_core/utils/file.py
+++ b/docling_core/utils/file.py
@@ -25,6 +25,36 @@ class DocumentStream(BaseModel):
     stream: BytesIO
 
 
+def resolve_remote_filename(
+    http_url: AnyHttpUrl,
+    response_headers: Dict[str, str],
+    fallback_filename="file",
+) -> str:
+    """Resolves the filename from a remote url and its response headers.
+
+    Args:
+        source AnyHttpUrl: The source http url.
+        response_headers Dict: Headers received while fetching the remote file.
+        fallback_filename str: Filename to use in case none can be determined.
+
+    Returns:
+        str: The actual filename of the remote url.
+    """
+    fname = None
+    # try to get filename from response header
+    if cont_disp := response_headers.get("Content-Disposition"):
+        for par in cont_disp.strip().split(";"):
+            # currently only handling directive "filename" (not "*filename")
+            if (split := par.split("=")) and split[0].strip() == "filename":
+                fname = "=".join(split[1:]).strip().strip("'\"") or None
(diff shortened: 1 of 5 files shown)


PeterStaar-IBM (approved):
lgtm

docling_core/utils/file.py
```diff
@@ -55,15 +65,55 @@ def resolve_file_source(
...
         try:
             local_path = TypeAdapter(Path).validate_python(source)
+            stream = BytesIO(local_path.read_bytes())
+            doc_stream = DocumentStream(name=local_path.name, stream=stream)
         except ValidationError:
             raise ValueError(f"Unexpected source type encountered: {type(source)}")
+    return doc_stream
+
+
+def resolve_file_source(
```
PeterStaar-IBM, line 81: add as depricated

cau-git, line 81: Reminder: Mark this one as @deprecated

cau-git:
We need to keep in mind that user code is directly importing `DocumentStream` from docling right now. We need to put a forward in docling where it is now.

docling_core/utils/file.py
```diff
@@ -7,28 +7,38 @@
...
 import requests
-from pydantic import AnyHttpUrl, TypeAdapter, ValidationError
+from pydantic import AnyHttpUrl, BaseModel, ConfigDict, TypeAdapter, ValidationError
 
 
-def resolve_file_source(
-    source: Union[Path, AnyHttpUrl, str], headers: Optional[Dict[str, str]] = None
-) -> Path:
-    """Resolves the source (URL, path) of a file to a local file path.
+class DocumentStream(BaseModel):
```
PeterStaar-IBM, line 18: move this class within docling-core and remove from docling

vagenas (commented):
Submitted my suggested improvement as #84:
- Exposes two cleanly named functions:
  - `resolve_source_to_stream()` and
  - `resolve_source_to_path()`
- Deprecates `resolve_file_source()`
- Adds typing-extensions pkg, needed for using the `@deprecated` decorator.

vagenas (dismissed):

cau-git (approved):

vagenas (approved):

PR #80, opened 2024-12-01
Added the InvisibleTextItem

PeterStaar-IBM:
Adding the ability to add Invisible text.

### DIFF
--- a/docling_core/types/doc/labels.py
+++ b/docling_core/types/doc/labels.py
@@ -28,12 +28,23 @@ class DocItemLabel(str, Enum):
     # Additional labels for markup-based formats (e.g. HTML, Word)
     PARAGRAPH = "paragraph"  # explicitly a paragraph and not arbitrary text
     REFERENCE = "reference"
-
+    INVISIBLE_TEXT = "invisible_text"
+    
     def __str__(self):
         """Get string value."""
         return str(self.value)
 
-
+class InvisibleTextLabel(str, Enum):
+    """InvisibleTextLabel."""
+    
+    UNSPECIFIED = "unspecified"
+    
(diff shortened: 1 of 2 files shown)


PeterStaar-IBM:
@vagenas can you add the issue wrt word comments?

vagenas:
It was not a GitHub issue, but rather an internal discussion — shared it privately now 😉

docling_core/types/doc/labels.py
```diff
@@ -28,12 +28,22 @@ class DocItemLabel(str, Enum):
...
     PARAGRAPH = "paragraph"  # explicitly a paragraph and not arbitrary text
     REFERENCE = "reference"
+    INVISIBLE_TEXT = "invisible_text"
 
     def __str__(self):
         """Get string value."""
         return str(self.value)
 
 
+class InvisibleTextLabel(str, Enum):
```
cau-git, line 38: I would not introduce an additional dimension just for`InvisibleTextLabel`, let's avoid to have a second-level dependency to `DocItemLabel.INVISIBLE_TEXT`, and just put straight:

```
DocItemLabel.COMMENT
DocItemLabel.NOTE
```

as regular labels. Then, we can add these to the furniture and enable that the furniture can be iterated over together with the rest.

cau-git:
Agreements from discussion:
- There should be no third category of labels.
- Things like slide comments, Word review comments, etc. should go into the `furniture` root and be put inside a group with new `GroupLabel.COMMENT_SECTION`
- Optional: We can extend the type hierachy with a `CommentItem`, which additionally stores an author name and timestamp with the text.
- We must add filter argument for _group labels_ on the export functions (additional to `DocItemLabel`s)

PR #82, opened 2024-12-02
fix: allow all url types in referenced exports

dolfim-ibm:
This changes impact only the case when a backend (or some user manipulation) defined an image with a remote url. It should be perfectly fine to return it in the exported html and markdown.

Refs: https://github.com/DS4SD/docling/discussions/477


PeterStaar-IBM (approved):
awesome!

cau-git (approved):

PR #85, opened 2024-12-02
expose resolve_remote_filename as a utility function

dolfim-ibm:


docling_core/utils/file.py
```diff
@@ -25,6 +25,33 @@ class DocumentStream(BaseModel):
...
+    # try to get filename from response header
+    if cont_disp := response_headers.get("Content-Disposition"):
+        for par in cont_disp.strip().split(";"):
+            # currently only handling directive "filename" (not "*filename")
+            if (split := par.split("=")) and split[0].strip() == "filename":
+                fname = "=".join(split[1:]).strip().strip("'\"") or None
+                break
+    # otherwise, use name from URL:
+    if fname is None:
+        fname = Path(http_url.path or "").name or "file"
```
vagenas, line 50: Could make sense instead of the hard-coded `"file"` to expose it as the default value of a kwarg like `fallback`.

docling_core/utils/file.py
```diff
@@ -25,6 +25,35 @@ class DocumentStream(BaseModel):
...
+def resolve_remote_filename(
+    http_url: AnyHttpUrl,
+    response_headers: Dict[str, str],
+    fallback_filename="file",
+) -> str:
+    """Resolves the filename from a remote url and its response headers.
+
+    Args:
+        source AnyHttpUrl: The source http url.
+        response_headers Dict: Headers received while fetching the remote file.
```
vagenas, line 37:
```suggestion
        response_headers Dict: Headers received while fetching the remote file.
        fallback_filename str: Filename to use in case none can be determined.
```

vagenas (approved):

PR #88, opened 2024-12-02
Bug: Support XLSX Mimetype

ctandrewtran:
### Context
Using latest docling-core (v2.6.1) and docling (v2.8.1) xlsx files are unable to be processed due to a ValueError mentioning that the mimetype `"application/vnd.openxmlformats-officedocument.spreadsheetml.sheet"` is not valid as per DocumentOrigin

`"application/vnd.openxmlformats-officedocument.spreadsheetml.sheet"` is not a mimetype offered by default by the `mimetype` pkg being used by docling and hence has to be added in the extras section

### Fix
Add the mimetype as an extra mimetype

### DIFF
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -382,6 +382,7 @@ class DocumentOrigin(BaseModel):
         "application/vnd.openxmlformats-officedocument.presentationml.presentation",
         "text/asciidoc",
         "text/markdown",
+        "application/vnd.openxmlformats-officedocument.spreadsheet.ml.sheet"
     ]
 
     @field_validator("binary_hash", mode="before")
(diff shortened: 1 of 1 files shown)


ctandrewtran:
> @ctandrewtran can you please provide a few more details on the failure that you observed?
> 
> * which operating system?
> * which python version?
> * was this on a specific file?
> * running which code/command?

**OS:** Amazon Linux 2

**Kernel Version:** 5.10.228-219.884.amzn2.x86_64

**Python Version:** 3.12.6

**Specific file:** any .xlsx file that will have mimetype `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet` can reproduce the issue

Code:

```python



```

ctandrewtran:
> @ctandrewtran can you please provide a few more details on the failure that you observed?
> 
> * which operating system?
> * which python version?
> * was this on a specific file?
> * running which code/command?

Closing the loop, someone else in the issue I raised in the Docling Repo mentioned having the same bug

cau-git (approved):

PR #91, opened 2024-12-03
feat: utilities converting document formats

dolfim-ibm:
- [x] Move from DoclingDocument to legacy
- [x] Add new payload argument in legacy format
- [ ] ~Add convert for legacy to DoclingDocument~ Will be added as new feature

### DIFF
--- a/test/data/doc/dummy_doc.yaml.dt
+++ b/test/data/doc/dummy_doc.yaml.dt
@@ -4,7 +4,6 @@
 <location><page_1><loc_59><loc_0><loc_91><loc_75></location>
 <caption>Figure 1: Four examples of complex page layouts across different document categories</caption>
 </figure>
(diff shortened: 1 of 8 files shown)


test/data/doc/dummy_doc.yaml.dt
```diff
@@ -4,7 +4,6 @@
 <location><page_1><loc_59><loc_0><loc_91><loc_75></location>
 <caption>Figure 1: Four examples of complex page layouts across different document categories</caption>
 </figure>
-<section_header><location><page_1><loc_42><loc_57><loc_49><loc_61></location>OPERATION (cont.)</section_header>
```
PeterStaar-IBM, line 7: Do we understand why a section header is missing?

dolfim-ibm: This is correct, because it was not supposed to be printed, since it is enclosed in a picture.

cau-git (approved):

PeterStaar-IBM (approved):
LGTM!

PR #93, opened 2024-12-04
build: set pydantic version to 2.10.3

ceberam:
- Revert constraint on pydantic version from commit 289b629 after resolution of conflicts in versions 2.10.0, 2.10.1, and 2.10.2.
- Update dependencies


dolfim-ibm:
Running all Docling tests in https://github.com/DS4SD/docling/pull/512.

dolfim-ibm (approved):

Issue #81, opened 2024-12-01
Deserialization of converted documents w/ images using `DoclingDocument.model_validate_json(json_file.read())` is broken.


Eric-Musa:
I discovered this in working on this PR (#71) to add `save_as_json` and `load_from_json` methods to `DoclingDocument`. 

Depending on which Pydantic method is used to load data back from a JSON file, the resulting DoclingDocument objects are not identical; I have confirmed this for the `ImageRef.uri` attribute, specifically when an image is embedded into the data (you can see the base64 encoding inline in the serialized JSON output).

I have summarized the script I used to reproduce the error below (I can't attach a Python file to this issue) and I've attached all the other files involved as well:
- The test PDF [test image.pdf](https://github.com/user-attachments/files/17970062/test.image.pdf)
- The DoclingDocument JSON output [document.json](https://github.com/user-attachments/files/17970063/document.json)
- The error message with traceback [error.txt](https://github.com/user-attachments/files/17970083/error.txt)

I think this problem is being caused by something in the `model_validate_json` pydantic method that is removing consecutive forward slash characters "//+" from `ImageRef.uri` attributes. When these necessary slashes are removed from the ImageRef.uri when loading the DoclingDocument, the uri is corrupted and the attribute type is set to `pathlib.Path`, instead of the `pydantic.AnyUrl` type, which calls the wrong image-handling code when `export_to_markdown` is called from the parent document.

```python
pipeline_options = PdfPipelineOptions()
pipeline_options.generate_picture_images = True

doc_converter = DocumentConverter(format_options=...)
document = doc_converter.convert(document_path).document

## using new method to save as json
document.save_as_json(document_json_path, image_mode=ImageRefMode.EMBEDDED)

## Load the document back from JSON using the model_validate_json method
with open(document_json_path, 'r') as f:
    document_json = DoclingDocument.model_validate_json(f.read())

## Load the document back from JSON using the model_validate_python method within the BaseModel with open(document_json_path, 'r') as f:
    document_py = DoclingDocument(**json.load(f))    

# ! ! ! ! ! ! ! ! ! ! ! ! ! ! ! ! ! !
print(document == document_json)   # >> FALSE, loading the data directly from JSON string is NOT identical
print(document == document_py)     # >> TRUE, loading the data into Python objects first is identical
# ! ! ! ! ! ! ! ! ! ! ! ! ! ! ! ! ! !

print('Exporting documents to markdown...')
[...]

PR #95, opened 2024-12-06
Fix: solved embedded image serialization/deserialization with Pydantic.

Eric-Musa:
Solved the problem in this issue (#81). Identified the problem is due to unexpected behavior with how Pydantic loads `Union` types:
```python
class ImageRef(BaseModel):
    mimetype: str
    dpi: int
    size: Size
    uri: Union[AnyUrl, Path]
    ...
```
[This docs page](https://docs.pydantic.dev/latest/concepts/unions/) lists `left_to_right` behavior as the simplest but **_not the default behavior_** of how `Union` types are processed. [Smart mode](https://docs.pydantic.dev/latest/concepts/unions/#smart-mode) is the default mode, which is a dynamic algorithm that picks the "best" option from the Union of types.

That was not the case in this particular instance, so I've manually defined the `uri` type to have `left_to_right` behavior when loading into memory. 
```python
    uri: Union[AnyUrl, Path] = Field(union_mode="left_to_right")
```

This **_does not break_** documents with `ImageRefMode.REFERENCED`! I added some tests to `test_docling_doc.py` to confirm that a `DoclingDocument` that has been saved to JSON file (with either ImageRefMode) is equivalent _in memory_ to the document loaded back from that JSON file.

I also revised the `DoclingDocument.load_from_json` function to use `model_validate_json` as @cau-git mentioned is best practice. In my investigation, I learned that Pydantic runs a project called [Jiter](https://pypi.org/project/jiter/) that is their home-grown parsing library written in Rust - its JSON parser is significantly faster than Python's `json.load`, so now that this deserialization issue is resolved, `DoclingDocument` can utilize the faster of the deserialization methods (since both will work now).


Eric-Musa:
And by testing the unit tests I wrote, I made sure the pre-commit checks all pass 👍🏼

cau-git (approved):
LGTM!

PR #100, opened 2024-12-10
feat: Add include_picture_contents argument to export/save methods

cau-git:


dolfim-ibm (commented):
Do we know how image annotations should be handled? Let's prepare for it, before committing with such an interface.

cau-git:
Keeping this on the shelf for the moment.

PeterStaar-IBM:
closing for now

PR #105, opened 2024-12-12
fix: improve doc item typing

vagenas:


docling_core/types/doc/document.py
```diff
@@ -1272,9 +1294,25 @@ def export_to_document_tokens(
...
+
+
+ContentItem = Annotated[
+    Union[
+        TextItem,
+        SectionHeaderItem,
+        ListItem,
+        PictureItem,
+        TableItem,
+        KeyValueItem,
```
PeterStaar-IBM, line 1313: add FormItem

PeterStaar-IBM (changes requested):
I think you forgot the FormItem in the ContentItem

docling_core/types/doc/document.py
vagenas: Well spotted, added now.

docling_core/types/doc/document.py
```diff
@@ -644,8 +659,10 @@ def export_to_document_tokens(
 class SectionHeaderItem(TextItem):
     """SectionItem."""
 
-    label: typing.Literal[DocItemLabel.SECTION_HEADER] = DocItemLabel.SECTION_HEADER
-    level: LevelNumber
+    label: typing.Literal[DocItemLabel.SECTION_HEADER] = (
+        DocItemLabel.SECTION_HEADER  # type: ignore[assignment]
+    )
+    level: LevelNumber = 1
```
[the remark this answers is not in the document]
vagenas: Yes, `label` is now non-overlapping — and is actually used as the discriminator field in `ContentItem` further below.

dolfim-ibm (commented):
Can we easily enforce the type also in the `add_text()` method?

vagenas:
> Can we easily enforce the type also in the `add_text()` method?

Technically it will now be "enforced" when the `TextItem` is created within that method.

If you mean in terms of reflecting it to the typing of the `label` param: if we invest more efforts in that area, I would rather do it in a way that reuses the type validation of the various `DocItems`, e.g. have a single `add_node()`, but I see that as requiring some redesign, and in general outside the scope of this PR.

PeterStaar-IBM (approved):
Awesome, let's get it merged!

dolfim-ibm (approved):

PR #106, opened 2024-12-12
fix: add REFERENCE to exported labels and remove CAPTION

dolfim-ibm:
- Add references in the exported output
- Remove the code-removal of captions and make it instead explicit via exported set


PeterStaar-IBM (approved):
Awesome!

PR #108, opened 2024-12-12
feat: add legacy to DoclingDocument utility

dolfim-ibm:


docling_core/utils/legacy.py
```diff
@@ -342,5 +349,276 @@ def _make_spans(cell: TableCell, table_item: TableItem):
...
+                label=DocItemLabel.FOOTNOTE, text=text_item.text, parent=doc.furniture
+            )
+
+    # main-text content
+    if legacy_doc.main_text is not None:
+        item: Optional[Union[BaseCell, BaseText]]
+
+        # collect all captions embedded in table and figure objects
+        # to avoid repeating them
+        embedded_captions: Dict[str, int] = {}
```
ceberam, line 425: Style comment: from python 3.9 (minimum required in `docling-core`) on, we can use the native types for type hints for `dict`, `list`, `bool`, `tuple` (i.e., no need to import `Dict` from `typings`).

docling_core/utils/legacy.py
```diff
@@ -342,5 +349,276 @@ def _make_spans(cell: TableCell, table_item: TableItem):
...
+
+                new_item = doc.add_table(data=table_data, prov=prov)
+                if (caption_item := floating_to_caption.get(ix)) is not None:
+                    if caption_item.text is not None:
+                        caption_prov = _transform_prov(caption_item)
+                        caption = doc.add_text(
+                            label=DocItemLabel.CAPTION,
+                            text=caption_item.text,
+                            prov=caption_prov,
+                            parent=new_item,
```
ceberam, line 599: Just to keep it in mind: this is currently not allowed according to the data schema (`parent` can only be `None` or a `GroupItem` but not a `TableItem`). With PR #103 , the constraint would be loosened to just a `NodeItem`.

ceberam (approved):

test/test_doc_legacy_convert.py
```diff
@@ -13,3 +21,20 @@ def test_new_to_old():
...
 
     docling_document_to_legacy(doc)
+
+
+def test_old_to_new():
+    filepath = Path("test/data/legacy_doc/doc-export.json")
+    leg_doc = DsDocument.model_validate_json(filepath.read_text())
+
+    doc = legacy_to_docling_document(leg_doc)
+
```
PeterStaar-IBM, line 31: Can you add here the 

```python
 leg_doc2 = docling_document_to_legacy(doc)
 assert leg_doc==leg_doc2
 ```

PeterStaar-IBM (commented):
see here: https://github.com/DS4SD/docling-core/pull/108/files#r1883701364

cau-git (approved):

test/test_doc_legacy_convert.py
dolfim-ibm: Changes found:
- name of labels (upper vs lower case)
- caption of figures are not in main-text anymore
- s3_data removed
- model metadata removed
- logs removed
- document hash cannot be preserved

PeterStaar-IBM (approved):
Let's add comments to the functions what is expected to change.

cau-git (approved):

PeterStaar-IBM (approved):

PR #129, opened 2025-01-15
feat: Add CodeItem as pydantic type, update export methods and APIs

Matteo-Omenetti:


cau-git (approved):

dolfim-ibm (dismissed):
LGTM

dolfim-ibm:
I think we should anyway consider having a stable enum for the programming languages. From the test about I see upcoming problems with Python vs python vs py, CSharp vs C# vs c#, etc

docling_core/types/doc/document.py
```diff
@@ -656,6 +655,15 @@ def export_to_document_tokens(
         return body
 
 
+class CodeItem(TextItem):
```
vagenas, line 658: I assume this should then also be included into the `ContentItem` tagged union.

cau-git: yes it should be.

docling_core/types/doc/labels.py
```diff
@@ -138,3 +138,69 @@ class TableCellLabel(str, Enum):
...
+    SCALA = "Scala"
+    SCHEME = "Scheme"
+    SWIFT = "Swift"
+    TYPESCRIPT = "TypeScript"
+    VISUALBASIC = "VisualBasic"
+    XML = "XML"
+    YAML = "YAML"
+    BC = "bc"
+    DC = "dc"
+    UNKNOWN = "unknown"
```
vagenas, line 202: These are partially sorted alphabetically, but not completely — I would best make the sorting consistent.

cau-git (approved):

dolfim-ibm (approved):

PR #131, opened 2025-01-17
fix: fix hybrid chunker token constraint

vagenas:
Resolves DS4SD/docling#723

- Root cause was that delimiters (both (1) between headings/captions & text, and (2) within headings/captions) were not properly accounted for when calculating the available number of tokens, which would manifest itself with different tokenizers (e.g. granite) or even with the default tokenizer (all-MiniLM-L6-v2) when used with more complex delimiters like `####`.
- Refactored its inner workings to reuse common API elements (`.serialize()`, `.delim`) instead of unnecessarily replicating serialization logic.
- Added test to cover addressed case.


dolfim-ibm:
python3.13 should be fixed with https://github.com/DS4SD/docling-core/pull/129/commits/f79791f97401e337111d23ad5ec777fa450c7ef9

dolfim-ibm (approved):

cau-git (approved):

PR #135, opened 2025-01-27
feat: add escaping_underscores option to markdown export

### LINKED ISSUE #134: feat: add escaping_underscores option to markdown export
If you use docling to prepare documents for input to gpt models using [RAG](https://en.wikipedia.org/wiki/Retrieval-augmented_generation), which uses [full-text-search](https://en.wikipedia.org/wiki/Full-text_search) or hybrid search, then you need to convert to unescaped markdown, since the keywords from the user's query may contain underscores and the required document chunks will not be found.

I suggest adding an option to `export_to_markdown()` which disable escaping when exporting to Markdown, and also using it with the False value when converting to text. The default value remains True.

Vdaleke:
BREAKING CHANGE: export to text no longer escapes underscores.

Add `escaping_underscores` option to `export_to_markdown(). 
Set default value to `escaping_underscores` to True.

resolves #134

### DIFF
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -1994,6 +1994,7 @@ class DoclingDocument(BaseModel):
         to_element: int = sys.maxsize,
         labels: set[DocItemLabel] = DEFAULT_EXPORT_LABELS,
         strict_text: bool = False,
+        escaping_underscores: bool = True,
         image_placeholder: str = "<!-- image -->",
         image_mode: ImageRefMode = ImageRefMode.PLACEHOLDER,
         indent: int = 4,
@@ -2016,6 +2017,7 @@ class DoclingDocument(BaseModel):
             to_element=to_element,
             labels=labels,
             strict_text=strict_text,
+            escaping_underscores=escaping_underscores,
             image_placeholder=image_placeholder,
             image_mode=image_mode,
             indent=indent,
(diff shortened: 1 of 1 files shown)


PeterStaar-IBM:
@Vdaleke I am in principle OK with this, but I would like to understand which problem it solves. In what examples do we not want to escape underscores?

Vdaleke:
> In what examples do we not want to escape underscores?

I gave an explanation in #134. This solves the problem when using docling to prepare documents for [RAG](https://en.wikipedia.org/wiki/Retrieval-augmented_generation), where text search of chunks is used among other things. When the document contains underscores, but the user query does not, the corresponding chunks are not found.

vladnosiv:
> @Vdaleke are you sure you actually need Markdown for your downstream application? Alternatively, you can use the option `strict_text=True` which is producing plain text.

Hello @dolfim-ibm !

In our case we want to use docling as data preparation for RAG. The data should be used both during search and during LLM context processing. LLM interacts better with Markdown, since it was learned largely from it (depends on the LLM I think, but in our case we are sure of it). 

So we want to convert arbitrary files to Markdown for search and processing by the model. For model processing, underscores don't seem to be a problem. But for search, particularly for text search (as part of hybrid search) text is just text and if the word A\_B\_C occurs in the text, it is not the same text as in a user query containing A_B_C. 

So we want to use markdown for model processing, but also markdown without escaping for search. Using strict text for lookup may be worse, as we may lose the importance of headers and table structure, which can be important when processing for vector search (again as part of hybrid search). 

We realize that such a flag will generate incorrect syntax, but since docling is positioned as a tool for preparing data for LLM too, we hope that adding such a flag will not be a problem.

dolfim-ibm (approved):

PeterStaar-IBM (approved):

PR #136, opened 2025-01-28
feat: added the geometric operations to BoundingBox

PeterStaar-IBM:


cau-git (changes requested):
I like that we get more functions but we should never add page_height as an argument, neither to existing nor to new methods. If the "other" bbox does not agree in coord_origin with the self bbox, it must raise an exception instead. It is user business to ensure this beforehand.

docling_core/types/doc/base.py
```diff
@@ -81,6 +83,20 @@ def normalized(self, page_size: Size) -> "BoundingBox":
 
         return out_bbox
 
+    def scale_to_size(self, page_size: Size) -> "BoundingBox":
```
dolfim-ibm, line 86: The `Size` argument is used to scale with different factors in each direction. I don't think the name `page_size` is correct here, because it has (almost) nothing to do with the page.

What about `size`? or `scales`? or `multipliers`?

dolfim-ibm:
This is also a useful geometric operation to add: https://github.com/DS4SD/docling/blob/main/docling/models/base_model.py#L71

docling_core/types/doc/base.py
```diff
@@ -59,27 +59,72 @@ def scaled(self, scale: float) -> "BoundingBox":
...
 
         """
-        out_bbox = copy.deepcopy(self)
-        out_bbox.l *= scale
-        out_bbox.r *= scale
-        out_bbox.t *= scale
-        out_bbox.b *= scale
-
-        return out_bbox
+        warnings.warn(
```
cau-git, line 62: I see absolutely no reason to deprecate methods, they can be used exactly as they were. If you want to support separate x and y scales, then the argument scale can be a Union[float, Size] and the implementation has to check.

docling_core/types/doc/base.py
```diff
@@ -59,27 +59,72 @@ def scaled(self, scale: float) -> "BoundingBox":
...
 
         :param page_size: Size:
 
         """
-        out_bbox = copy.deepcopy(self)
-        out_bbox.l /= page_size.width
-        out_bbox.r /= page_size.width
-        out_bbox.t /= page_size.height
-        out_bbox.b /= page_size.height
+        warnings.warn(
```
cau-git, line 75: See above comment

docling_core/types/doc/base.py
```diff
@@ -59,27 +59,72 @@ def scaled(self, scale: float) -> "BoundingBox":
...
+        return BoundingBox(
+            l=self.l / size.width,
+            r=self.r / size.width,
+            t=self.t / size.height,
+            b=self.b / size.height,
+            coord_origin=self.coord_origin,
+        )
 
-        return out_bbox
+    def scale_to_size(self, size: Size) -> "BoundingBox":
```
cau-git, line 96: scale_to_size is unnecessary since scaled can accommodate this, and also a misnomer. If anything it would be scale_by_size.

docling_core/types/doc/base.py
```diff
@@ -176,3 +263,116 @@ def to_top_left_origin(self, page_height: float) -> "BoundingBox":
...
+
+    def is_left_of(self, other: "BoundingBox") -> bool:
+        """is_left_of."""
+        return self.r < other.r
+
+    def is_strictly_left_of(self, other: "BoundingBox", eps: float = 0.001) -> bool:
+        """is_strictly_left_of."""
+        return (self.r + eps) < other.l
+
+    def is_above_of(self, other: "BoundingBox") -> bool:
```
cau-git, line 331: is_above_of?? -> is_above

docling_core/types/doc/base.py
```diff
@@ -176,3 +263,116 @@ def to_top_left_origin(self, page_height: float) -> "BoundingBox":
...
+            self.coord_origin == CoordOrigin.BOTTOMLEFT
+            and other.coord_origin == CoordOrigin.BOTTOMLEFT
+        ):
+            return self.t > other.t
+        else:
+            raise ValueError("BoundingBox is not BOTTOMLEFT")
+
+        return False
+
+    def is_strictly_above_of(self, other: "BoundingBox", eps: float = 0.001) -> bool:
```
cau-git, line 343: Remove _of from the name

docling_core/types/doc/base.py
```diff
@@ -176,3 +263,116 @@ def to_top_left_origin(self, page_height: float) -> "BoundingBox":
...
+
+            if self.b < max_ij and min_ij < self.t:  # overlap_y
+                return False
+
+            if self.l < elem_i.r and elem_j.l < self.r:
+                return True
+
+            return False
+        else:
+            raise ValueError("BoundingBox is not BOTTOMLEFT")
```
cau-git, line 376: This error looks like it should not exist. The only error i can legitimately see is "The coordinate origin of this BoundingBox and other BoundingBox is not the same". For all other cases the geometric functions must simply work.

dolfim-ibm:
Just mentioning here one possibility that was discussed.

One wish is a simple way of dealing with rescaling from a given page size to a new one. In the current solution this can be done as a combination of `.normalize(old_page_size).scale(new_page_size)`.

Would it worth leaving the `scale()` and `normalize()` untouched, and bring a new function

```py
def resize_by_scale(self, x_scale: float, y_scale: float):
    return BoundingBox(
        l=self.l * x_scale,
        r=self.r * x_scale,
        t=self.t * y_scale,
        b=self.b * y_scale,
        coord_origin=self.coord_origin,
    )

def scale_to_size(self, orig_page_size: Size, new_page_size: Size):
    return self.resize_by_scale(x_scale=new_page_size.width/orig_page_size.width, y_scale=new_page_size.height/orig_page_size.height)

# same as before, but using the implementation above
def scale(self, scale: float):
    return self.resize_by_scale(scale, scale)

# same as before, but using the implementation above
def normalized(self, page_size: Size):
    return self.scale_to_size(orig_page_size=page_size, new_page_size=Size(1.0,1.0))
```

PeterStaar-IBM:
> ```python
> def resize_by_scale(self, x_scale: float, y_scale: float):
>     return BoundingBox(
>         l=self.l * x_scale,
>         r=self.r * x_scale,
>         t=self.t * y_scale,
>         b=self.b * y_scale,
>         coord_origin=self.coord_origin,
>     )
> 
> def scale_to_size(self, orig_page_size: Size, new_page_size: Size):
>     return self.resize_by_scale(x_scale=new_page_size.width/orig_page_size.width, y_scale=new_page_size.height/orig_page_size.height)
> 
> # same as before, but using the implementation above
> def scale(self, scale: float):
>     return self.resize_by_scale(scale, scale)
> 
> # same as before, but using the implementation above
> def normalized(self, page_size: Size):
>     return self.scale_to_size(orig_page_size=page_size, new_page_size=Size(1.0,1.0))
> ```

well, that was basically what was originally proposed, just different names. Anyhow, adapted this now ...

docling_core/types/doc/base.py
PeterStaar-IBM: done

docling_core/types/doc/base.py
PeterStaar-IBM: done

cau-git (approved):

dolfim-ibm (approved):

docling_core/types/doc/base.py
PeterStaar-IBM: done

docling_core/types/doc/base.py
PeterStaar-IBM: done

PR #143, opened 2025-01-31
fix: add html escape in md export and fix formula escapes

dolfim-ibm:
Resolves https://github.com/DS4SD/docling/issues/764

Escape html in md and html exports and skip the underscore escape in formulas and code.


cau-git (approved):

PeterStaar-IBM (approved):
nice!

PR #144, opened 2025-01-31
feat(HTML): Export formulas with mathml

dolfim-ibm:
feat(HTML): Allow the export of formulas as mathml
feat(HTML): Retain the original latex using html standards
fix(HTML): Keep newlines in pre blocks

Chore:
- Rework the sanitize function and move newline cleanup in there
- Remove functions which would never be executed (because the labels logic changed)

Open questions:
1. Should this be the default?
2. Should the new dependency always be there or as extras?


PeterStaar-IBM (approved):
Let's go!! 🚀

vagenas (dismissed):

vagenas (approved):

PeterStaar-IBM (approved):

cau-git (approved):

PR #150, opened 2025-02-04
fix test data restoring alphabetical order

vagenas:


vagenas:
Handled directly within #148

PR #156, opened 2025-02-11
feat: add inline groups, revamp Markdown export incl. list groups

vagenas:
Contains:
- new group label
- updated `iterate_items()`
- updated `export_to_markdown()` as an example of how export functions can be extended for inline groups

Case of inline group within list item (see `inline_within_list`) still to be addressed.


PeterStaar-IBM (approved):
Great!

dolfim-ibm (approved):

PR #176, opened 2025-02-27
feat: add document tokens for SMILES

lucas-morin:
Add document tokens `<smiles>` and `</smiles>` for representing SMILES sequences. 

Example sequence:
`<smiles>CC1=CN2C=C(C3=NC(C)=NN3)N=C(CC3=CC=CC=C3F)C2=N1</smiles>`

### DIFF
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -1003,6 +1003,20 @@ class PictureItem(FloatingItem):
             predicted_class = classifications[0].predicted_classes[0].class_name
             body += DocumentToken.get_picture_classification_token(predicted_class)
 
+        smiles_annotations = [
+            ann for ann in self.annotations if isinstance(ann, PictureMoleculeData)
+        ]
+        if len(smiles_annotations) > 0:
+            body += (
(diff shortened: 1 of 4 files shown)


cau-git (approved):

docling_core/types/doc/document.py
```diff
@@ -1003,6 +1003,20 @@ def export_to_document_tokens(
...
 
+        smiles_annotations = [
+            ann for ann in self.annotations if isinstance(ann, PictureMoleculeData)
+        ]
+        if len(smiles_annotations) > 0:
+            body += (
+                "<"
+                + DocumentToken.SMILES.value
+                + ">"
+                + smiles_annotations[0].smi
```
vagenas, line 1014: @lucas-morin just to confirm: is it on purpose that we only consider the first element of `smiles_annotations`?

lucas-morin: Yes, because there should only be one SMILES annotation per picture.
Should I change this to avoid using a list?

vagenas (approved):

PR #180, opened 2025-03-03
fix(markdown): fix escaping in case of nesting

vagenas:
Avoid escaping HTML/underscores on results from nested calls, as these have already been escaped as needed.


test/data/doc/constructed_doc.dt
```diff
@@ -30,7 +30,7 @@ Affiliation 2</text>
 <list_item>item 2 of neighboring list</list_item>
 <unordered_list><list_item>item 1 of sub list</list_item>
 <paragraph>Here a code snippet:</paragraph>
-<code<_unknown_>print("Hello world")</code
+<code<_unknown_><p>Hello world</p></code
```
dolfim-ibm, line 33: this looks weird. but I'm not sure it is related to the current PR

vagenas: This is a change I did in the input test data to trigger the bug & get it included in the test data.
I.e. if one removes the fix from document.py, the result is this one:
Screenshot 2025-03-03 at 15 35 02

PeterStaar-IBM (approved):

cau-git (approved):

test/data/doc/constructed_doc.dt
vagenas: DocTags code export issue addressed with https://github.com/DS4SD/docling-core/pull/181.

PR #182, opened 2025-03-03
feat: add serializers, text formatting, update Markdown export

vagenas:


PeterStaar-IBM (approved):

docling_core/types/doc/document.py
```diff
@@ -2319,10 +2543,10 @@ def save_as_markdown(
 
     def export_to_markdown(  # noqa: C901
         self,
-        delim: str = "\n\n",  # TODO deprecate
+        delim: str = "\n\n",
         from_element: int = 0,
         to_element: int = sys.maxsize,
-        labels: set[DocItemLabel] = DEFAULT_EXPORT_LABELS,
+        labels: set[DocItemLabel] = DOCUMENT_TOKENS_EXPORT_LABELS,
```
vagenas, line 2549: Changed this, as otherwise the new logic would literally filter out any non-listed labels from showing up, i.e. we would see no captions at all in the output.

test/test_docling_doc.py
```diff
@@ -767,6 +780,54 @@ def _construct_doc() -> DoclingDocument:
 
     doc.add_form(graph=graph)
 
+    inline_fmt = doc.add_group(label=GroupLabel.INLINE)
+    doc.add_text(
+        label=DocItemLabel.PARAGRAPH, text="Some formatting chops:", parent=inline_fmt
```
PeterStaar-IBM, line 785: Dont use `PARAGRAPH`, we should only use `TEXT`. `PARAGRAPH` comes from a legacy, which we should deprecate (or at least mark as deprecated)

PeterStaar-IBM (changes requested):

PeterStaar-IBM (approved):

dolfim-ibm (approved):
Well done!

PR #187, opened 2025-03-11
feat: Adding load_from_doctags method to Docling document

maxmnemonic:


cau-git (dismissed):

docling_core/types/doc/document.py
```diff
@@ -2950,6 +2951,329 @@ def _image_fallback(item: TextItem):
 
         return html_text
 
+    def load_from_document_tokens(  # noqa: C901
+        self,
+        pages_doctags: List[str],
+        pages_images: List[PILImage.Image],
+    ) -> "DoclingDocument":
```
vagenas, line 2958: The rest of the `load_from_*` methods here are actually loading from a file (technically a `Path`).
Is this something possible here too, so these are consistent?

If not, perhaps we should consider a different naming to avoid confusion.

In any case, I would additionally consider encapsulating everything needed for representing a document in DocTags (presumably `pages_doctags` and `pages_images`?) in a single class.

cau-git: @vagenas The specialty with DocTags is that they don't constitute a self-sufficient format. The image must be always present as a paired resource for the DocTags to make sense. Then, we also have two cases for the DocTags:
1. Every single page as one DocTags file, and many of them together form the document
2. A single DocTags file, which has `<page_break>` tags inside to denote multiple pages.

Hence, I think it will be difficult to keep the signature of this method aligned with the other `load_from_*`. It would be impractical for users to construct an object first that needs to be passed into the load method, when the typical input they have is a folder with images and doctags.txt.

vagenas: @cau-git I still believe it makes sense to encapsulate DocTags, as that will greatly facilitate working with it going forward.

Here a quick outline of a possible approach (naming is probably improvable):

```python

class DocTagsPage(BaseModel):
  text: str

class FullDocTagsPage(DocTagsPage):
  image: PILImage.Image

class DocTagsDoc(BaseModel):
  pages: List[DocTagsPage]
  # should provide initialization that also considers the
  # multi-page text scenario with `<page_break>` and map it
  # to this unified representation
```

nassarofficial (approved):

docling_core/types/doc/document.py
vagenas: @cau-git I see this was now incorporated so resolving this conversation.

cau-git (approved):

cau-git (dismissed):

nassarofficial (approved):
Tried with prediction, looks good to me.

PR #197, opened 2025-03-18
fix: Add caption to the table in load_from_doctags

Saidgurbuz:


maxmnemonic (approved):
Wonderful! Tested with SmolDocling, issue with table captions fixed.
Tests on `docling-core` and `docling` are unaffected

cau-git (approved):

PR #198, opened 2025-03-18
chore: bring HybridChunker to init file for easy access

ceberam:
To be aligned with the `HierarchicalChunker` and for an easy import from applications, the `HybridChunker` has been added to the `__init__.py` file of the `docling_core/transforms/chunker` directory


vagenas (changes requested):
The hybrid chunker depends on extra requirements, so if we add it to the __init__.py, it will prevent proper package access to installations without the extra.

PR #203, opened 2025-03-19
Adding handling for str as filename argument to load/save methods

muhark:
## Summary

This PR makes a minor change to the handling of the `filename` argument in the `save_as_*` and `load_from_*` methods of the `document` and `page` classes.

See: https://github.com/docling-project/docling/issues/1189

## Details

Currently, if a `str` is passed to the `save_as_*` or `load_from_*` methods of the `document` or `pages` types, it will fail with an `AttributeError`.

The changes in this PR handle `str` inputs to these methods, and automatically casts them to `pathlib.Path`.

All tests have succesffuly passed :)

### DIFF
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -2475,12 +2475,14 @@ class DoclingDocument(BaseModel):
 
     def save_as_json(
         self,
-        filename: Path,
+        filename: str | Path,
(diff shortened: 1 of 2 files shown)


muhark:
ah, I didn't see the conventional commit.
I'll amend.

PR #207, opened 2025-03-20
fix: markdown picture item export

### LINKED ISSUE #208: Bug with PictureItem markdown export
Incorrect initialization of the MarkdownDocSerializer class, self (Picture Item) is passed to the doc field instead of doc (DoclingDocument)
https://github.com/docling-project/docling-core/blob/2371c11b8f74628169a9bb377036511235070af0/docling_core/types/doc/document.py#L964

which produces:
```
>       serializer = MarkdownDocSerializer(
            doc=self,
            image_mode=image_mode,
        )
E       pydantic_core._pydantic_core.ValidationError: 1 validation error for MarkdownDocSerializer
E       doc
E         Input should be a valid dictionary or instance of DoclingDocument [type=model_type, input_value=PictureItem(self_ref='#/p...ge=None, annotations=[]), input_type=PictureItem]
```

Vdaleke:
Use `doc` instead of `self` for MarkdownDocSerializer init in PictureItem.

Resolves #208


Vdaleke:
Hello @vagenas, could you take a look at the bug fix from your commit a7cdc87411102b15da5459e0b5d00fbe966aea9d?

vagenas (approved):
Good catch!

PR #217, opened 2025-03-26
feat: allow images in doctags deserializer to be optional and support multipage

kmyusk:
* Allow the list of images argument to `from_doctags_and_image_pairs` to be optional
* Verify that `load_from_doctags` works for `DocTagsDocument` without image
* Provide deserializer for `DocTagsDocument` from multipage DocTags with `<page_break>`

### DIFF
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -551,8 +551,6 @@ class DocTagsDocument(BaseModel):
             raise ValueError("Number of page doctags must be equal to page images!")
         doctags_doc = cls()
 
-        # if images is None:
-        #     images = [None] * len(doctags)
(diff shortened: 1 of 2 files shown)


kmyusk:
The utility of allowing image to be `None` seems limited. As the provenance is not populated in case there is no image, the `DocSerializer` can't infer the `page_ranges` and treats the whole doc as a single page https://github.com/docling-project/docling-core/blob/main/docling_core/experimental/serializer/common.py#L175. So the page split information is completely lost. There could be some value in making the deserializer more lenient but I'm not aware of a real use-case. Maybe better to just make https://github.com/docling-project/docling-core/blob/main/docling_core/types/doc/document.py#L535 required? It will greatly simplify the `load_from_doctags` code as well not having to check for empty image.

PR #219, opened 2025-03-27
fix: Automatic transformation of output cells bbox coord origin defined by input in get_cells_in_bbox

maxmnemonic:
Added automatic transformation of page cell_bbox coord origin to the same as input bbox.coord_origin when requesting get_cells_in_bbox.

Current behavior - is to throw error and stop if the input coord_origin differs from backend page coord_origin, which is almost always when get_cells_in_bbox is used in a client code of Docling. This PR fixes it, and presents resulting cells pre-transformed into the same origin as input crop bbox.

This small fix removes the overhead in the client code we put in Docling.


cau-git (approved):
🛠️ nice fix.

Issue #211, opened 2025-03-23
Add manipulation operations on items of a DoclingDocument.


PeterStaar-IBM:
# Support merge, update, insert, and delete operations for DocItem using self-references

For agentic frameworks that need to generate documents, we can not just rely on the simple `add_***` functionality, but we need to be able to  merge, update, and delete Docitems/GroupItems.

## Description:

We need to extend the DocItem framework to support merge, update, and delete operations using the concept of self-references — that is, paths pointing to the current item’s location within the larger JSON document structure.

This functionality is important for enabling fine-grained manipulation of document elements, especially in interactive workflows (e.g., LLM-based editing or user-driven document transformations).

cau-git:
@vagenas this is a good start. One use case for which this is requested frequently is to merge multiple DoclingDocument instances into one, and in extension, splitting one DoclingDocument up into several (e.g. by page ranges).

vagenas:
The complexity arises from having the insert / delete operations applied on our current, **static**-reference representation.

💡 If we decouple these operations from reference materialization by introducing a **dynamic**-reference representation, i.e. based on memory pointers instead of static strings, things look more streamlined.

The figure below shows how it could look like on a high level:

<img width="1688" alt="Image" src="https://github.com/user-attachments/assets/372eac9d-e77b-4210-bd6c-171eb6e3ad16" />

PeterStaar-IBM:
Merged!

PR #220, opened 2025-03-27
feat: added initial delete and insert methods in DoclingDocument

PeterStaar-IBM:
## Description of new features

1. added methods in NodeItem: `get_parent_ref`, `delete_child`, `update_child`, `add_child`, `add_sibling`
2. added manipulation methods: `insert_item`, `delete_items`, `insert_item_at_refitem`
3. added helper methods: `get_stack_of_refitem`, `iterate_items_with_stack`

- [x] added tests for manipulation

issues: 

1. https://github.com/docling-project/docling-core/issues/211


docling_core/types/doc/document.py
```diff
@@ -2348,6 +2701,71 @@ def iterate_items(
                     included_content_layers=my_layers,
                 )
 
+    def iterate_items_with_stack(
+        self,
+        root: Optional[NodeItem] = None,
+        with_groups: bool = False,
+        traverse_pictures: bool = False,
+        page_no: Optional[int] = None,
+        included_content_layers: set[ContentLayer] = DEFAULT_CONTENT_LAYERS,
```
vagenas, line 2710: Mutable default arguments is [bad practice].

docling_core/types/doc/document.py
```diff
@@ -2348,6 +2701,71 @@ def iterate_items(
                     included_content_layers=my_layers,
                 )
 
+    def iterate_items_with_stack(
```
vagenas, line 2704: - first let's double check if this is really needed (see PR comment on exposing "`stack`" on the API)
- if needed, let's check:
    - if it should be part of the public API
    - what we can reuse with `iterate_items()`

vagenas (commented):
Regarding the _implementation_ approach, you can see my proposal [here](https://github.com/docling-project/docling-core/issues/211#issuecomment-2758774199).

Regarding the _API design_, here a couple first comments in the spirit of simplicity, consistency, and maintainability:

1. The API already exposes one way of referring to a node, using ref strings, e.g. `"#/texts/10"`. The PR proposes introducing an additional one with `stack`. Why is that needed? Cannot the operations be defined on the existing ref string notation (e.g. internalizing the "conversion" from "stack" to self-ref)?
1. If I understand correctly, if one calls `add_child()` on a node that already has children, it does nothing, just returns `False`. Calling a method named `add_child()` and having no child getting added can seem counter-intuitive. One could consider renaming (e.g. `add_child_if_no_children()`, or preferably, expose a more explicit operation set as proposed [here](https://github.com/docling-project/docling-core/issues/211#issuecomment-2758278176):
    - `append_child()` (semantics of `list.append()`, i.e. append to the end of it regardless if it is empty or not)
    - `add_sibling_before()`
1. I have left a couple in-line comments
1. Please also rebase on latest changes

docling_core/types/doc/document.py
```diff
@@ -458,8 +458,12 @@ def get_ref(self):
         populate_by_name=True,
     )
 
+    def path(self):
+        """Get the path of the reference."""
+        return self.cref.split("/")
```
vagenas, line 463: Instead of extending the public API, I would just add this as an internal helper function on the module level (e.g. `_split_ref_to_path(ref_str)`)

docling_core/types/doc/document.py
```diff
@@ -598,10 +602,79 @@ class NodeItem(BaseModel):
 
     model_config = ConfigDict(extra="forbid")
 
-    def get_ref(self):
+    def get_ref(self) -> RefItem:
         """get_ref."""
         return RefItem(cref=self.self_ref)
 
+    def get_parent_ref(
```
vagenas, line 609: If that's only to be used by our implementation, instead of extending the public API, I would just add this as an internal helper function on the module level.

docling_core/types/doc/document.py
vagenas, line 2704: - first let's double check if this is really needed (see PR comment on exposing "`stack`" on the API)
- if needed, let's check:
    - if it should be part of the public API
    - what we can reuse with `iterate_items()`

docling_core/types/doc/document.py
vagenas, line 2710: Mutable default arguments is [bad practice](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments).

vagenas (commented):
Regarding the _implementation_ approach, you can see my proposal [here](https://github.com/docling-project/docling-core/issues/211#issuecomment-2758774199).

Regarding the _API design_, here a couple first comments in the spirit of simplicity, consistency, and maintainability:

1. The API already exposes one way of referring to a node, using ref strings, e.g. `"#/texts/10"`. The PR proposes introducing an additional one with `stack`. Why is that needed? Cannot the operations be defined on the existing ref string notation (e.g. internalizing the "conversion" from "stack" to self-ref)?
1. If I understand correctly, if one calls `add_child()` on a node that already has children, it does nothing, just returns `False`. Calling a method named `add_child()` and having no child getting added can seem counter-intuitive. One could consider renaming (e.g. `add_child_if_no_children()`, or preferably, expose a more explicit operation set as proposed [here](https://github.com/docling-project/docling-core/issues/211#issuecomment-2758278176):
    - `append_child()` (semantics of `list.append()`, i.e. append to the end of it regardless if it is empty or not)
    - `add_sibling_before()`
1. I have left a couple in-line comments
1. Please also rebase on latest changes

docling_core/types/doc/document.py
PeterStaar-IBM: I think that the original iterate_items can just reuse the `iterate_items_with_stack` where the level = len(stack)

PeterStaar-IBM: I fixed the duplicate code.

docling_core/types/doc/document.py
PeterStaar-IBM: fixed that

docling_core/types/doc/document.py
PeterStaar-IBM: done

docling_core/types/doc/document.py
PeterStaar-IBM: done

PeterStaar-IBM:
> Regarding the _implementation_ approach, you can see my proposal [here](https://github.com/docling-project/docling-core/issues/211#issuecomment-2758774199).
> 
> Regarding the _API design_, here a couple first comments in the spirit of simplicity, consistency, and maintainability:
> 
> 1. The API already exposes one way of referring to a node, using ref strings, e.g. `"#/texts/10"`. The PR proposes introducing an additional one with `stack`. Why is that needed? Cannot the operations be defined on the existing ref string notation (e.g. internalizing the "conversion" from "stack" to self-ref)?
> 2. If I understand correctly, if one calls `add_child()` on a node that already has children, it does nothing, just returns `False`. Calling a method named `add_child()` and having no child getting added can seem counter-intuitive. One could consider renaming (e.g. `add_child_if_no_children()`, or preferably, expose a more explicit operation set as proposed [here](https://github.com/docling-project/docling-core/issues/211#issuecomment-2758278176):
>    
>    * `append_child()` (semantics of `list.append()`, i.e. append to the end of it regardless if it is empty or not)
>    * `add_sibling_before()`
> 3. I have left a couple in-line comments
> 4. Please also rebase on latest changes

So, good points, but ...

1. the stack provides a different type of information as a reference: namely the natural order in the tree of the node-items. The References can not be compared (eg which ref comes in front of another ref). This ordering is important to delete items of the tree in place
2. `add_child` is defined as adding a new "layer" to a leaf in the tree. This can only happen if there are no children so add_child_if_no_children() is a tautology. add_sibling is only allowed on an existing node.This question I have is if we should swallow the error (return False) or actually raising an error.

vagenas:
> the stack provides a different type of information as a reference: namely the natural order in the tree of the node-items. The References can not be compared (eg which ref comes in front of another ref). This ordering is important to delete items of the tree in place

This "stack" information may be useful information for the (current) _implementation_, but does the _user_ need to know?

If possible, I would expose the ref string — to have a consistent node reference notation — and do the mapping internally.

vagenas:
> add_child is defined as adding a new "layer" to a leaf in the tree. This can only happen if there are no children so add_child_if_no_children() is a tautology. 

I understand this is how it is _implemented_, what I am saying is that the name does not unambiguously reflect that.

> This question I have is if we should swallow the error (return False) or actually raising an error

Kind of depends on the client, i.e. agent code — ideally raise, but perhaps returning a code (e.g. `False`) is simpler and bound to work in more scenarios.

cau-git (commented):
@PeterStaar-IBM I checked this proposal and I realize it exposes low-level primitives (refs, stack), which comes with several complexities that we do not want to expose to a user.

I discussed with @vagenas again and we would suggest a minimal and safe user-facing API to cover the desired cases:

```python
  def append_child_item(self,
               node_item: NodeItem,
               parent: Optional[NodeItem] = None,
               ):
      '''Adds an item, given its node_item instance, as the last child to the parent. Equivalent to add_text, add_list, ...'''
      pass

  def insert_item_after(self,
               new_item: NodeItem,
               other: NodeItem,
               ):
      '''Inserts an item, given its node_item instance, after other as a sibling.'''
      pass

  def insert_item_before(self,
               new_item: NodeItem,
               other: NodeItem,
               ):
      '''Inserts an item, given its node_item instance, before other as a sibling.'''
      pass

  def delete_items(self, node_item: NodeItem):
      '''Deletes an item, given its instance or ref, and any children it has.'''
      pass

  def replace_item(self, node_item: NodeItem, new_item: NodeItem):
      '''Replaces an item, given a new_item instance, at the position in the tree of node_item.
      Convenience method that combines deleting the node_item and inserting the new_item.'''
      pass
```

This makes manipulation of documents a lot safer.
1. Users should not deal with `ref`s, since after every manipulation, the refs may change and as such be invalidated. If users need to go from a given `ref` they can resolve it first to the item instance with `ref.resolve()`.
2. We see no need to expose the stack primitive, this should be hidden implementation detail.

PeterStaar-IBM:
> @PeterStaar-IBM I checked this proposal and I realize it exposes low-level primitives (refs, stack), which comes with several complexities that we do not want to expose to a user.
> 
> I discussed with @vagenas again and we would suggest a minimal and safe user-facing API to cover the desired cases:
> 
> ```python
>   def append_child_item(self,
>                node_item: NodeItem,
>                parent: Optional[NodeItem] = None, # default parent is the root node.
>                ):
>       '''Adds an item, given its node_item instance, as the last child to the parent. Equivalent to add_text, add_list, ...'''
>       pass
> 
>   def insert_item_after_sibling(self,
>                new_item: NodeItem,
>                sibling: NodeItem,
>                ):
>       '''Inserts an item, given its node_item instance, after other as a sibling.'''
>       pass
> 
>   def insert_item_before_sibling(self,
>                new_item: NodeItem,
>                sibling: NodeItem,
>                ):
>       '''Inserts an item, given its node_item instance, before other as a sibling.'''
>       pass
> 
>   def delete_items(self, node_items: List[NodeItem]):
>       '''Deletes an item, given its instance or ref, and any children it has.'''
>       pass
> 
>   def replace_item(self, old_item: NodeItem, new_item: NodeItem):
>       '''Replaces an item, given a new_item instance, at the position in the tree of old_item.
>       Convenience method that combines deleting the old_item and inserting the new_item.'''
>       pass
> ```
> 
> This makes manipulation of documents a lot safer.
> 
> 1. Users should not deal with `ref`s, since after every manipulation, the refs may change and as such be invalidated. If users need to go from a given `ref` they can resolve it first to the item instance with `ref.resolve()`.
> 2. We see no need to expose the stack primitive, this should be hidden implementation detail.
> 
> ## Constraints
> * Inserting or adding a NodeItem which itself has children is not allowed (raises)
> * If the "sibling" after or before which to insert is not in the tree, it must raise

Implemented as discussed, looking forward to next review @cau-git @vagenas

vagenas (commented):
Left some inline comments.

test/test_docling_doc.py
```diff
@@ -1248,3 +1257,99 @@ def test_save_to_disk():
     _verify_saved_output(filename=filename, paths=paths)
 
     assert True
+
+
+def test_document_manipulation():
```
vagenas, line 1262: Let's also cover adding a sibling to the root (body), which should raise an exception.

docling_core/types/doc/document.py
```diff
@@ -624,10 +628,79 @@ class NodeItem(BaseModel):
...
+            return True
+        elif len(stack) > 1 and stack[0] < len(self.children):
+            item = self.children[stack[0]].resolve(doc)
+            return item._update_child(doc=doc, stack=stack[1:], new_ref=new_ref)
+
+        return False
+
+    def _add_child(
+        self, doc: "DoclingDocument", stack: list[int], new_ref: RefItem
+    ) -> bool:
```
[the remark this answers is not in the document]
PeterStaar-IBM: not here, that is the responsibility of the method that calls this one (since it is recursive)

docling_core/types/doc/document.py
```diff
@@ -624,10 +628,98 @@ class NodeItem(BaseModel):
...
+        """Add child to a child-less node."""
```
vagenas, line 678: Description can be updated since now node does not need to be child-less.

docling_core/types/doc/document.py
```diff
@@ -1683,6 +1775,371 @@ def transform_to_content_layer(cls, data: dict) -> dict:
                     item["content_layer"] = "furniture"
         return data
 
+    # ---------------------------
+    # Public Manipulation methods
+    # ---------------------------
+
+    def append_child_item(
+        self, *, child: NodeItem, parent: Optional[NodeItem] = None
+    ) -> RefItem:
```
vagenas, line 1784: Let's just return `None` and not return the reference str, as the latter is static.

The user anyway holds (a pointer to) the `child` object, so at any point they can access its current reference with `child.get_ref()`.

Same applies for all public methods.

test/test_docling_doc.py
```diff
@@ -1248,3 +1257,125 @@ def test_save_to_disk():
...
+        2,
+        2,
+        0,
+        2,
+        0,
+        0,
+    ], f"stack==[2, 2, 2, 0, 2, 0, 0] for stack: {stack}"
+
+
+def test_document_manipulation():
```
vagenas, line 1283: Still uncovered cases:
- `replace_item()`
- error case of adding a sibling to the root (body), which should raise an exception
- error case of adding a child which itself has children

vagenas (commented):
Some final comments inline.

vagenas (approved):

cau-git (approved):

PR #221, opened 2025-03-28
feat: integrate serialization API into chunkers

vagenas:
todo:
- [x] ensure HybridChunker is aligned with passed serializer
- [x] ensure grounding/span info is propagated for DocTags (completely missing) and complete for Markdown
- [x] add picture-based customization example code


simonschoe:
@vagenas any chance this will be covered by this or a future PR as well? 😇

vagenas:
> @vagenas any chance this will be covered by this or a future PR as well? 😇

@simonschoe were you intending to include some link or something? 🤔

simonschoe:
> > @vagenas any chance this will be covered by this or a future PR as well? 😇
> 
> @simonschoe were you intending to include some link or something? 🤔

Indeed, sorry for that. https://github.com/docling-project/docling/issues/962

docling_core/experimental/serializer/base.py
```diff
@@ -25,10 +26,19 @@
 )
 
 
+class Span(BaseModel):
```
PeterStaar-IBM, line 29: What is the purpose of a new class Span if it only contains DocItem? 

1. Why not just use the DocItem
2. Are we sure we do not want to use NodeItem (I guess an INLINE group could be useful)?

vagenas: > What is the purpose of a new class Span if it only contains DocItem?

By already _encapsulating_ what has provenance / bounding box info, the idea is to contain / minimize the impact of expected future extensions like the ones discussed below (similarly to why we had already encapsulated the `SerializationResult` instead of just a string).

> 1. Why not just use the DocItem

The current implementation indeed just includes the `DocItem`, but to address more advanced use cases (e.g. cross-page paragraphs), going forward we can include a `prov_idx` (commented out in the code below), where 0 means the _first_ text "span" / provenance within that doc item, 1 the _second_ etc.

> 2. Are we sure we do not want to use NodeItem (I guess an INLINE group could be useful)?

In principle, groups do not have prov info, so that is why it's `DocItem`-based.
However, if a `SerializationResult` corresponds to a whole group it could indeed be useful to have a pointer to the group too, that's why an optional group could be added as well in the future (commented out in the code below).

PeterStaar-IBM: Ok, the point is that lists are groups, so we can have ListItem, Table (with captions), Picture, etc as a Span, but not an entire list. That looks a bit unsymmetric, since I believe we do want to serialize entire lists in the chunks.

vagenas: This is why the `SerializationResult` contains a `list[Span]` comprising all contributing `DocItem`s (e.g. list items) wrapped as `Span`s.

PeterStaar-IBM (dismissed):

ceberam (approved):

dolfim-ibm (approved):

PR #225, opened 2025-03-31
feat: allow images in doctags deserializer to be optional and support multipage

kmyusk:
* Allow the list of images argument to `from_doctags_and_image_pairs` to be optional
* Verify that `load_from_doctags` works for `DocTagsDocument` without image
* Provide deserializer for `DocTagsDocument` from multipage DocTags with <page_break>


cau-git (approved):
LGTM

PR #228, opened 2025-04-02
fix: properly handle document_index in load_from_doctags()

kmyusk:
`DOCUMENT_INDEX` is currently handled by the `TextItem` part in `load_from_doctags`, which fails because it is not a supported label. Properly handle `DOCUMENT_INDEX` label by adding it as a trivial `TableItem`. I have never seen a document index with structure so far, thus the assumption that the content is one long string.


kmyusk:
Agreed. I will close this PR then. We can open a new one once we decide what to do with document_index.

PR #229, opened 2025-04-02
feat: Support of DocTags charts (serialization and deserialization)

maxmnemonic:
- Doctags charts deserialization into Docling Document
- Added PictureTabularChartData
- Added chart serializer for DocTags (<chart> tag)
- Fixed DocTags table deserialization
- Updated tests


cau-git (dismissed):

docling_core/experimental/serializer/doctags.py
```diff
@@ -234,9 +262,15 @@ def serialize(
                 parts.append(cap_text)
 
         text_res = "".join(parts)
+        pic_label = DocItemLabel.PICTURE
+        if is_chart:
+            pic_label = DocItemLabel.CHART
+        print(pic_label)
```
vagenas, line 268: Let's remove this print.

docling_core/types/doc/document.py
```diff
@@ -3927,8 +3991,13 @@ def parse_key_value_item(
                     else:
                         self.add_table(data=table_data, caption=caption)
 
-                elif tag_name == DocItemLabel.PICTURE:
+                elif tag_name in [DocItemLabel.PICTURE, DocumentToken.CHART.value]:
```
vagenas, line 3994: It seems a bit odd that one value is used from the `DocItemLabel` enum and one from the `DocumentToken` enum — let's double check and align if possible.

docling_core/types/doc/labels.py
```diff
@@ -8,6 +8,7 @@ class DocItemLabel(str, Enum):
     """DocItemLabel."""
 
     CAPTION = "caption"
+    CHART = "chart"
```
vagenas, line 11: In principle we maintain a mapping between the various `DocItem` subtypes and the `DocItemLabel` values — and also we maintain a union of the exact possible types (although the latter is still not used).

So, as this introduces a new label, let's also:
- capture which `DocItem` type this label belongs to (in my understanding that would be something like [1] — but feel free to update accordingly)
- extend `ContentItem` in document.py to include the new type.

[1]
```python
class ChartItem(PictureItem):
    label: typing.Literal[DocItemLabel.CHART] = DocItemLabel.CHART
```

vagenas (commented):
Left some comments inline.

docling_core/types/doc/labels.py
maxmnemonic: @vagenas it's a PictureItem, there is no need to introduce new item in Docling itself, it's only an artifact of Doctags, that was not documented or reflected in the labels, hence I added it.

docling_core/experimental/serializer/doctags.py
maxmnemonic: Done

docling_core/types/doc/document.py
maxmnemonic: Done

docling_core/types/doc/labels.py
vagenas: Right now `PictureItem` does not allow this label: https://github.com/docling-project/docling-core/blob/32c17fe2a559d7169e26877f480cb69cdf82176c/docling_core/types/doc/document.py#L1062

If it's not a new `PictureItem` subtype, then we should at least capture it in the allowed labels for `PictureItem`.

docling_core/types/doc/document.py
```diff
@@ -3761,6 +3794,40 @@ def parse_table_content(otsl_content: str) -> TableData:
...
+                "vbar_categorical",
+                "hbar_categorical",
+            ]
+
+            # Current SmolDocling can predict different labels:
+            chart_labels_mapping = {
+                "line": PictureClassificationLabel.LINE_CHART,
+                "dot_line": PictureClassificationLabel.LINE_CHART,
+                "vbar_categorical": PictureClassificationLabel.BAR_CHART,
+                "hbar_categorical": PictureClassificationLabel.BAR_CHART,
```
vagenas, line 3818: If these four strings aren't yet governed anywhere, I would add introduce what makes sense (e.g. enum) to govern them (and reuse them).

cau-git: These are inconsistent labels learned from different training data, we want to map these and not have them as citizens.

vagenas, line 3818: @maxmnemonic @cau-git can then look just like this:

```suggestion

            # temporary labels currently predicted by SmolDocling
            tmp_label_line = "line"
            tmp_label_dot_line = "dot_line"
            tmp_label_vbar_categorical = "vbar_categorical"
            tmp_label_hbar_categorical = "hbar_categorical"

            label = None
            chart_labels = [
                PictureClassificationLabel.PIE_CHART,
                PictureClassificationLabel.BAR_CHART,
                PictureClassificationLabel.STACKED_BAR_CHART,
                PictureClassificationLabel.LINE_CHART,
                PictureClassificationLabel.FLOW_CHART,
                PictureClassificationLabel.SCATTER_CHART,
                PictureClassificationLabel.HEATMAP,
                tmp_label_line,
                tmp_label_dot_line,
                tmp_label_vbar_categorical,
                tmp_label_hbar_categorical,
            ]

            # Current SmolDocling can predict different labels:
            chart_labels_mapping = {
                tmp_label_line: PictureClassificationLabel.LINE_CHART,
                tmp_label_dot_line: PictureClassificationLabel.LINE_CHART,
                tmp_label_vbar_categorical: PictureClassificationLabel.BAR_CHART,
                tmp_label_hbar_categorical: PictureClassificationLabel.BAR_CHART,
```

maxmnemonic: 🤷

PR #232, opened 2025-04-03
feat: add HTML serializer

PeterStaar-IBM:


docling_core/experimental/serializer/html.py
```diff
@@ -0,0 +1,922 @@
...
+                '<meta name="generator" content="Docling HTML Serializer">'
+            )
+
+        # Add default styles or custom CSS
+        if params.css_styles:
+            head_parts.append(f"<style>\n{params.css_styles}\n</style>")
+        elif self.params.split_page_view:
+            head_parts.append(_get_css_for_split_page())
+        else:
+            head_parts.append(_get_css_for_single_column())
```
cau-git, line 911: To make the result predictable, this code should check the that `output_style == OutputStyle.CUSTOM` (see suggestion above), and raise an exception when `css_styles` is provided but the output_style is not CUSTOM.

docling_core/experimental/serializer/html.py
```diff
@@ -0,0 +1,922 @@
...
+    html_lang: str = "en"
+    css_styles: Optional[str] = None
+    add_document_metadata: bool = True
+    prettify: bool = True  # Add indentation and line breaks
+
+    # Formula rendering options
+    formula_to_mathml: bool = True
+
+    # Allow for split page view (only possible if page-images are present)
+    split_page_view: bool = False
```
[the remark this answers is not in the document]
dolfim-ibm: Just thinking out loud (not really advocating for it), but the split html could actually be a new serializer which derives from the normal one and overwrites the single `serialize()` function.

docling_core/experimental/serializer/html.py
```diff
@@ -0,0 +1,922 @@
...
+    ) -> SerializationResult:
+        """Serialize a document out of its pages."""
+        # Create HTML structure
+        html_parts = [
+            "<!DOCTYPE html>",
+            self._generate_head(),
+            "<body>",
+        ]
+
+        if self.params.split_page_view:
```
dolfim-ibm, line 787: for later: we should not abuse html columns for making layout. there are html standards for making this kind of views. but this can be improved later, it is only an internal detail

docling_core/experimental/serializer/html.py
```diff
@@ -0,0 +1,922 @@
...
+            return SerializationResult(text="")
+
+        graph_serializer = HTMLGraphDataSerializer()
+
+        # Add key-value if available
+        key_value = graph_serializer.serialize(
+            item=item.graph,
+            doc_serializer=doc_serializer,
+            doc=doc,
+            tag="form-container",
```
dolfim-ibm, line 603: I don't think we should create our own "tags". in html it is better to use `<div class="form-container"></div>`

PeterStaar-IBM: yes, I think this is just wrong parameter naming, it should be called `class_name`, because we indeed use `<div class="form-container"></div>`

docling_core/experimental/serializer/html.py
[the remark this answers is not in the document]
PeterStaar-IBM: @dolfim-ibm I am not a big fan of making more classes. If we define a parameter class and that covers all the cases, then we just make it more complicated.

@cau-git Just to be clear, the output_style can only have two values: `single-column` or `split-page`. In each case, you could define your own css styling, if you wish. The split-page can only happen if you have page-images. Hence, since the output_style and the css_style are independent of each other, I would not unroll all posibilities.

docling_core/experimental/serializer/html.py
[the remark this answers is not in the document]
PeterStaar-IBM: see suggestion above!

vagenas (commented):
@PeterStaar-IBM various improvements proposed with https://github.com/docling-project/docling-core/pull/242 e.g.:
- including span information in the returned `SerializationResult`s
- caption handling in case of floating text items (e.g. code item)
- caption handling in case holding element is excluded

vagenas (approved):

vagenas (approved):

cau-git (approved):
🚢

PR #235, opened 2025-04-04
feat: Markdown chart serializer (picture+table)

maxmnemonic:
Implementation of chart serializer in Markdown, checks if Image has chart tabular data attached to it, and renders table with extracted values underneath.
Can be controlled by `enable_chart_tables` parameter (True by default)


docling_core/experimental/serializer/markdown.py
```diff
@@ -57,6 +58,7 @@ class MarkdownParams(CommonParams):
     layers: set[ContentLayer] = {ContentLayer.BODY}
     image_mode: ImageRefMode = ImageRefMode.PLACEHOLDER
     image_placeholder: str = "<!-- image -->"
+    disable_chart_tables: bool = False
```
cau-git, line 61: Let's use the convention of "positive names", otherwise we end up with double negation. Hence, "enable_chart_tables: bool = True".

maxmnemonic: Done!

Saidgurbuz (approved):
LGTM!

cau-git (approved):

PR #243, opened 2025-04-09
Update page.py

cahoover:
feat: added optional font_size

My project needs a font size (if available) within the pdf spans. 

It appears the C++ within docling parse already contains this information, so we updated pdf_parser.py with a corresponding field. (See the pull request over at docling parse). 

*This* change updates page.py to include a field to store the font size.


cahoover:
Looks like the issue is deeper in the C++

PR #244, opened 2025-04-10
feat: document question and answer (DocQnA)

ceberam:


PeterStaar-IBM:
@ceberam Good start, but I would like to have support directly for multi-turn conversations with different roles,

```
messages = [
    {"role": "system", "content": "You are a helpful assistant."},
    {"role": "user", "content": "What's the weather like today?"},
    {"role": "assistant", "content": "Today's weather is sunny with a high of 25°C."}
]
```

Can we replace the `QA` class with a MessageItem (single line) and the `ConversationItem` with be a list of `MessageItem`. We also need an enum for role, and MessageLabel.

PR #249, opened 2025-04-11
perf: fix serialization performance

vagenas:
Addresses docling-project/docling#1335.


docling_core/experimental/serializer/common.py
```diff
@@ -50,6 +50,49 @@
...
+    page_break_i = 0
+    for item, _ in doc.iterate_items(
+        root=node,
+        with_groups=True,
+        included_content_layers=layers,
+        traverse_pictures=traverse_pictures,
+    ):
+        if isinstance(item, DocItem):
+            if item.prov:
+                page_no = item.prov[0].page_no
```
PeterStaar-IBM, line 83: I think I would like to understand a bit deeper why we only look at the first element. I think, we could do slightly better here since you yield DocItems.

PeterStaar-IBM (changes requested):

PeterStaar-IBM (approved):
🎖️

dolfim-ibm (approved):
lgtm

PR #263, opened 2025-04-16
feat: add visualizers

vagenas:


docling_core/transforms/visualizer/layout_visualizer.py
```diff
@@ -0,0 +1,195 @@
...
+                        fill=cell_color,
+                    )
+                # Draw cluster rectangle
+                x0, y0, x1, y1 = c.bbox.as_tuple()
+                x0 *= scale_x
+                x1 *= scale_x
+                y0 *= scale_x
+                y1 *= scale_y
+
+                if y1 <= y0:
```
PeterStaar-IBM, line 76: Let's use the coordinate transformations from BoundingBox.

docling_core/transforms/visualizer/layout_visualizer.py
```diff
@@ -0,0 +1,195 @@
...
+from docling_core.types.doc.document import ContentLayer, DocItem, DoclingDocument
+from docling_core.types.doc.page import TextCell
+
+
+class Cluster(BaseModel):
+    """Cluster."""
+
+    id: int
+    label: DocItemLabel
+    bbox: BoundingBox
```
PeterStaar-IBM, line 24: Let's make this a BoundingRectangle. Every BoundingRectangle can be mapped into a BoundingBox. If we ever want rotated BoundingBox, we at least have the right type from the start.

PeterStaar-IBM (changes requested):

docling_core/transforms/visualizer/layout_visualizer.py
vagenas: Done.

PeterStaar-IBM (dismissed):
I wonder if we want to keep only the first 3 pages, to reduce the number of images we "keep" in the git-tree.

PeterStaar-IBM (approved):

dolfim-ibm (approved):

PR #267, opened 2025-04-21
chore: Types fixes

Fogapod:
This mostly addresses missing `Any` on `kwargs` and incomplete `Dict` annotations. These break external code.

`disallow_untyped_defs` catches `kwargs` but there are still ~80 errors when it's enabled, mostly missing return types. Dont know if you want these (pyright infers them) so I didn't fix that.

As for untyped `Dict`s I couldn't find a mypy rule to catch these


docling_core/types/doc/document.py
```diff
@@ -300,7 +300,7 @@ class TableCell(BaseModel):
     @classmethod
     def from_dict_format(cls, data: Any) -> Any:
         """from_dict_format."""
-        if isinstance(data, Dict):
+        if isinstance(data, dict):
```
Fogapod, line 303: This looked like a bug but apparently it's not:
```python
>>> from typing import Dict
>>> isinstance({}, Dict)
True
>>> isinstance({}, dict)
True
```
Replaced it to avoid confusion. Typehints shouldn't be used for instance checks

dolfim-ibm (approved):
lgtm

ceberam (approved):

PR #269, opened 2025-04-24
fix: UnboundLocal variable

Saidgurbuz:
To fix `UnboundLocalError: cannot access local variable 'body_str' where it is not associated with a value` error in the edge case of an empty table.


dolfim-ibm (approved):
lgtm

PR #273, opened 2025-04-28
fix: make load_from_doctags method static

Saidgurbuz:
- made `load_from_doctags` method static
- added `document_name` method to be able to configure the name of the DoclingDocument


dolfim-ibm (approved):
lgtm

Issue #279, opened 2025-05-01
html serializer for table is too slow


ta0a2000t:
### Bug
HTMLTableSerializer is using an inefficient method to read the table.

``` python
                    cell: TableCell = item.data.grid[i][j]

```

however, in the markdown serializer, cell is accessed differently

``` python
            rows = [
                [
                    # make sure that md tables are not broken
                    # due to newline chars in the text
                    col.text.replace("\n", " ")
                    for col in row
                ]
                for row in item.data.grid
            ]
```

#### Conclusion:
grid is made of a 2d linked list, so bracket access is O(n**2)

...

### Steps to reproduce

document.export_to_html() # slow, cant finish with large tables in xlsx document

vs 

document.export_to_markdown() # time is okay

...

### Docling version

Docling version: 2.31.0
Docling Core version: 2.28.1
Docling IBM Models version: 3.4.2
Docling Parse version: 4.0.1
Python: cpython-312 (3.12.0)

...

### Python version
Python 3.12.0

...

ta0a2000t:
i have the code to fix HTMLTableSerializer , let me know if i can

ta0a2000t:
https://github.com/docling-project/docling-core/pull/281 fixes it

PR #281, opened 2025-05-03
fix: HTMLTableSerializer efficiency and dir attribute

ta0a2000t:
HTMLTableSerializer Improvements
- Changed indexing to iteration for the 2D linked list grid due to performance issues with large XLSX exports to Markdown. The original implementation was extremely slow and wouldn't finish. It is good now.
- Also a tiny issue with text direction, and this PR fixes it.

fixes https://github.com/docling-project/docling-core/issues/279#issue-3033288664

### DIFF
--- a/docling_core/transforms/serializer/html.py
+++ b/docling_core/transforms/serializer/html.py
@@ -395,7 +395,7 @@ class HTMLTableSerializer(BaseTableSerializer):
 
                     text_dir = get_text_direction(content)
                     if text_dir == "rtl":
-                        opening_tag += f' dir="{dir}"'
+                        opening_tag += f' dir="{text_dir}"'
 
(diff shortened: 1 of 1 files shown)


vagenas (approved):

ta0a2000t:
Hi @vagenas thanks for your message and for merging a commit I made.
Regarding the grid loop efficiency part, the issue is still there.
The fix in [373 ](https://github.com/docling-project/docling-core/pull/373)does not resolve the issue.
I have just tried this on my version and on the fix from https://github.com/docling-project/docling-core/pull/373 , which took two minutes then I killed the process, while mine took 1 second to iterate through the 2d loop. (The test file is a long xlsx file with multiple sheets).

Can you please show me how to return my other commit that resolves the HTMLTableSerializer issue?


Thanks!
CC: @cau-git

PR #285, opened 2025-05-06
feat: Make TableCell a NodeItem to support rich table cell content

cau-git:


test/test_rich_tables.py
```diff
@@ -0,0 +1,87 @@
...
+    doc.add_list_item(text="bar", parent=list_node)
+
+    return doc
+
+
+def test_rich_table():
+
+    doc = _construct_doc()
+
+    html_pred = doc.export_to_html()
```
dolfim-ibm, line 86: similar to other tests (see [here](https://github.com/docling-project/docling-core/blob/main/test/test_docling_doc.py#L960) for example) we could simplify the test results removing the styling

```suggestion
    html_pred = doc.export_to_html(html_head="")
```

docling_core/types/doc/document.py
```diff
@@ -1178,6 +1116,98 @@ def export_to_doctags(
         return text
 
 
+class TableCell(NodeItem):
+    """TableCell."""
+
+    bbox: Optional[BoundingBox] = None
+    row_span: int = 1
```
PeterStaar-IBM, line 1123: Do we need a variable here? I think this can be a method computing `end_row_offset_idx-start_row_offset_idx`?

docling_core/types/doc/document.py
```diff
@@ -1178,6 +1116,98 @@ def export_to_doctags(
...
+    col_span: int = 1
```
PeterStaar-IBM, line 1124: Do we need a variable here? I think this can be a method computing `end_col_offset_idx-start_col_offset_idx`?

docling_core/types/doc/document.py
```diff
@@ -1178,6 +1116,98 @@ def export_to_doctags(
...
+    start_row_offset_idx: int
+    end_row_offset_idx: int
```
PeterStaar-IBM, line 1126: can we check somewhere that `end_*>start_*`?

PeterStaar-IBM (changes requested):

docling_core/types/doc/document.py
cau-git: We can do something about this but it is unrelated to the changes of this PR, and we cannot break backwards-compatibility.

PR #288, opened 2025-05-08
Feat(Enrichment):  Added Figure Description with indexed figure numbers back in markdown

shivanikabu:
Added Figure Description with indexed figure numbers back in markdown. We have created an image mode named description to pass back the description into the markdown result file


Nikhil200030:
image
image

PR #290, opened 2025-05-12
feat: add parameter to choose of which pages export the doctags

rickymaggio02:
Add parameter to export only certain pages to doctors from a Docling document


docling_core/types/doc/document.py
```diff
@@ -3885,6 +3885,7 @@ def export_to_doctags(  # noqa: C901
         add_table_cell_location: bool = False,
         add_table_cell_text: bool = True,
         minified: bool = False,
+        pages: Optional[list[int]] = None,
```
vagenas, line 3888: Let's make it a set since the underlying parameter is also a set.
```suggestion
        pages: Optional[set[int]] = None,
```

docling_core/types/doc/document.py
```diff
@@ -3903,6 +3904,7 @@ def export_to_doctags(  # noqa: C901
         :param # table specific flagsadd_table_cell_location: bool
         :param add_table_cell_text: bool:  (Default value = True)
         :param minified: bool:  (Default value = False)
+        :param pages: list[int]: (Default value = None)
```
vagenas, line 3907:
```suggestion
        :param pages: set[int]: (Default value = None)
```

vagenas (changes requested):
@rickymaggio02
- I have left some  inline comments/suggestions.
- Besides, let's test-cover the new param (you can draw inspiration from how we test `minified` in test_docling_doc.py).

vagenas (approved):

PR #295, opened 2025-05-16
feat: add annotations in MD & HTML serialization

vagenas:
- included annotations in MD & HTML serialization as default (users can opt out)
  - MD: appended to text
  - HTML: appended to `figcaption` content
  - (DocTags currently not touched as apparently specialized annotation handling already included there)
  - exposed via param `include_annotations` in respective DoclingDocument methods
- annotation handling single-sourced in common.py
- annotation types currently handled: classification, description, & SMILES

### DIFF
--- a/docling_core/transforms/serializer/html.py
+++ b/docling_core/transforms/serializer/html.py
@@ -35,7 +35,7 @@ from docling_core.transforms.serializer.base import (
 from docling_core.transforms.serializer.common import (
     CommonParams,
     DocSerializer,
-    _serialize_picture_annotation,
+    _get_picture_annotation_text,
     create_ser_result,
 )
 from docling_core.transforms.serializer.html_styles import (
@@ -458,12 +458,6 @@ class HTMLPictureSerializer(BasePictureSerializer):
                     res_parts.append(
                         create_ser_result(text=html_table_content, span_source=item)
                     )
-
-        if params.include_annotations:
-            for annotation in item.annotations:
-                if ann_text := _serialize_picture_annotation(annotation=annotation):
-                    ann_ser_res = create_ser_result(text=ann_text, span_source=item)
-                    res_parts.append(ann_ser_res)
 
         text_res = "".join([r.text for r in res_parts])
         if text_res:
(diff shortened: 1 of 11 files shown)


docling_core/transforms/serializer/html.py
```diff
@@ -943,18 +948,37 @@ def serialize_captions(
...
+                    if isinstance(it := cap.resolve(self.doc), TextItem)
+                    and it.self_ref not in excluded_refs
+                ]
+            )
+
+        if params.include_annotations and item.self_ref not in excluded_refs:
+            if isinstance(item, PictureItem):
+                for ann in item.annotations:
+                    if ann_text := _get_picture_annotation_text(annotation=ann):
+                        ann_ser_res = create_ser_result(
```
dolfim-ibm, line 967: In `html` I think it would be more natural to group the annotations in a `<div class="annotations">` with a given class.

It would be nice to have a different `<div class="annotation-TYPE">` for each different one, but I think also a single one is good for now.

docling_core/transforms/serializer/html.py
```diff
@@ -110,6 +111,10 @@ class HTMLParams(CommonParams):
     # Enable charts to be printed into HTML as tables
     enable_chart_tables: bool = True
 
+    include_annotations: bool = True
+    annotation_opening_marker: str = "<!-- generated beginning -->"
+    annotation_closing_marker: str = "<!-- generated end -->"
```
dolfim-ibm, line 116: I don't consider annotations as _always_ "model generated", so a more generic naming might be better.
For example, the content of `blah` could also fit there.

```suggestion
    annotation_opening_marker: str = "<!-- annotations beginning -->"
    annotation_closing_marker: str = "<!-- annotations end -->"
```

vagenas: Called them "annotations" instead with the latest commit.

dolfim-ibm (approved):
lgtm

docling_core/transforms/serializer/html.py
vagenas: Resolved in #305

PeterStaar-IBM (dismissed):
lgtm!

dolfim-ibm (approved):

PeterStaar-IBM (approved):
lgtm!

PR #302, opened 2025-05-23
feat(doctags): add enclosing bbox to inline

vagenas:


docling_core/types/doc/document.py
```diff
@@ -3626,7 +3661,7 @@ def parse_key_value_item(
                     # no closing tag; only the existence of the item is recovered
                     full_chunk = f"<{tag_name}></{tag_name}>"
 
-                doc_label = tag_to_doclabel.get(tag_name, DocItemLabel.PARAGRAPH)
+                doc_label = tag_to_doclabel.get(tag_name, DocItemLabel.TEXT)
```
vagenas, line 3664: I assume the default text label should actually be "text" (and not "paragraph").

cau-git (approved):

nassarofficial (approved):

PR #303, opened 2025-05-23
chore: fix deprecation warnings

ceberam:
It partially addresses https://github.com/docling-project/docling/issues/1557

- Update code to avoid deprecation warnings
- Catch and ignore deprecation warnings when we intentionally use deprecated fields


PeterStaar-IBM (approved):
lgtm!

dolfim-ibm (approved):

PR #306, opened 2025-05-27
fix(HybridChunker): refine `max_tokens` auto-detection

vagenas:
HF transformers expose a token window as **`model_max_length`**, which is the one **currently used** for auto-detecting the max tokens when the respective parameter is not passed.

SentenceTransformers embeddings wrapping around HF transformers can in principle define their own token window as **`max_seq_length`** — which is not accessible via the HF transformers library (it's accessible via sentence-transformers lib, but that requires torch).

E.g. for all-MiniLM-L6-v2 (our default model):
- `model_max_length` exposed by HuggingFace is 512 ([src](https://huggingface.co/sentence-transformers/all-MiniLM-L6-v2/blob/main/tokenizer_config.json#L1))
- `max_seq_length` actually used in SentenceTransformers is 256 ([src](https://huggingface.co/sentence-transformers/all-MiniLM-L6-v2/blob/main/sentence_bert_config.json#L2))

👉  The PR suggests only auto-detecting `max_tokens` for SentenceTransformers, and that in an opportunistic fashion by reading the respective config file via HF Hub, whenever available.
Whenever not available (e.g. because model is not SentenceTransformer (e.g. google-bert/bert-base-uncased) or because config file not in HF cache and running offline etc.), it will emit an error telling users to pass `max_tokens` explicitly.

Note: using `model_max_length` as a fallback was deemed inappropriate in this PR as it could lead to different auto-detected max token limit for the same SentenceTransformer model depending on offline mode/local availability of config file.

Context:
https://stackoverflow.com/a/75908603


ceberam (approved):
Just a small comment: we may want to document that this automated `max_tokens` detection will imply storing a file on the default folder `.cache/huggingface/`

PR #307, opened 2025-05-27
refactor: use uv as dependencies management and packaging

dolfim-ibm:
Additionally to the usual checks, we add a new CI job which validates if the wheel contains all the required files (including the package-data).

_Note: the switch from black to ruff will follow in a new PR (which will likely change lots of files)._


PeterStaar-IBM (dismissed):
lgtm! this is a game changer!

ceberam (dismissed):

vagenas (approved):

PR #309, opened 2025-05-28
chore: exclude test data from GH Linguist

vagenas:
For fixing the language detection, which currently shows up like this: 
Screenshot 2025-05-28 at 13 17 44

See https://github.com/orgs/community/discussions/54854


dolfim-ibm (approved):
nice catch!

PR #313, opened 2025-06-03
fix: Support section_header levels in doctags deserialization

cau-git:
Addresses issue where any tag from `<section_header_level_2>` up are lost in de-serialization.


dolfim-ibm (commented):
Should this increase the version number of the DoclingDocument?

vagenas:
> Should this increase the version number of the DoclingDocument?

I do not see a reason for that.

PR #319, opened 2025-06-05
feat: add subscript & superscript formatting

vagenas:
Added subscript & superscript support, for now only leveraged in HTML (the Markdown syntax seems to be ["uncommon"](https://www.markdownguide.org/extended-syntax/#subscript) & won't render e.g. on VS Code).

### DIFF
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -836,6 +836,8 @@ class Formatting(BaseModel):
     italic: bool = False
     underline: bool = False
     strikethrough: bool = False
+    subscript: bool = False
+    superscript: bool = False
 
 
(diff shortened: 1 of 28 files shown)


docling_core/types/doc/document.py
```diff
@@ -836,6 +836,8 @@ class Formatting(BaseModel):
     italic: bool = False
     underline: bool = False
     strikethrough: bool = False
+    subscript: bool = False
```
PeterStaar-IBM: this will need to be an enum: you can not have subscript and superscript equal to false.

PeterStaar-IBM (approved):
awesome!

dolfim-ibm (approved):

PR #325, opened 2025-06-10
feat: visualizing the tables

PeterStaar-IBM:


docling_core/transforms/visualizer/table_visualizer.py
```diff
@@ -0,0 +1,135 @@
...
+                    else:
+                        _log.error(
+                            f"Nothing to visualize, update params: {self.params}"
+                        )
+
+                else:
+                    raise RuntimeError(f"Cannot visualize page-image for {page_nr}")
+
+            else:
+                _log.error("Can not yet visualise tables with multiple provenances")
```
vagenas, line 115: As this does not stop the execution, I would rather use a warning instead of error.

docling_core/transforms/visualizer/table_visualizer.py
```diff
@@ -0,0 +1,135 @@
...
+                            table=elem,
+                            page_height=doc.pages[page_nr].size.height,
+                            page_image=image,
+                            scale_x=image.width / doc.pages[page_nr].size.width,
+                            scale_y=image.height / doc.pages[page_nr].size.height,
+                        )
+                    else:
+                        _log.error(
+                            f"Nothing to visualize, update params: {self.params}"
+                        )
```
vagenas: If `show_cells` is indeed always mandatory, I would simplify and remove it altogether from the params.

docling_core/transforms/visualizer/table_visualizer.py
```diff
@@ -0,0 +1,135 @@
...
+            page_image = doc.pages[page_nr].image
+            if page_image is None or (pil_img := page_image.pil_image) is None:
+                raise RuntimeError("Cannot visualize document without images")
+            elif page_nr not in my_images:
+                image = deepcopy(pil_img)
+                my_images[page_nr] = image
+
+        for idx, (elem, _) in enumerate(
+            doc.iterate_items(
+                included_content_layers={ContentLayer.BODY, ContentLayer.FURNITURE}
```
vagenas: Would best future-proof this like `included_content_layers={c for c in ContentLayer}`
(actually needs to be updated also in the other visualizers).

PR #333, opened 2025-06-17
fix(HybridChunker): improve long heading handling

### LINKED ISSUE #327: HybridChunker ignores too big chunks instead of only passing chunk_text
When using the hybridChunker we came into the issue that chunks which are together with their headers and captions are longer than the allowed chunk size these are ignored and therefore are lost when processing a document.

Is this the desired behavior? Our idea would be to in this case maybe only return the text and disregard the headers and captions or only take the headers and captions partially to fit. 
Additionally, it would be helpful to not only depict the text in the warning but also the meta information to log all lost information. (change in warning message {doc_chunk.text=} to {doc_chunk=})

The Warning comes from:
[/opt/conda/envs/Python-RT24.1-Premium/lib/python3.11/site-packages/docling_core/transforms/chunker/hybrid_chunker.py:235: ](https://github.com/docling-project/docling-core/blob/7094828eaa00d57d3b820ba423df7c1160c2f711/docling_core/transforms/chunker/hybrid_chunker.py#L235C17-L239C18)
UserWarning: Headers and captions for this chunk are longer than the total amount of size for the chunk, chunk will be ignored: doc_chunk.text='<CHUNK_TEXT>'

To reproduce this is our setup:
docling-core
Version: 2.34.2

```
tokenizer = AutoTokenizer.from_pretrained(
    "./multilingual-e5-small", local_files_only=True
)

chunker = HybridChunker(
    tokenizer=tokenizer,
    max_tokens=300, 
)

doc = DoclingDocument(**docling_json)
chunker.chunk(doc)
```

We attached the document in which this occurs.

[IBM_Power_Redbook.pdf](https://github.com/user-attachments/files/20692853/IBM_Power_Redbook.pdf)

vagenas:
Resolves #327.


dolfim-ibm (approved):
lgtm

cau-git (approved):
Too long didn't read.

PR #337, opened 2025-06-20
feat: add page chunking

vagenas:


docling_core/transforms/chunker/page_chunker.py
```diff
@@ -0,0 +1,59 @@
...
+                    meta=DocMeta(
+                        doc_items=ser_res.get_unique_doc_items(),
+                        headings=None,
+                        captions=None,
+                        origin=dl_doc.origin,
+                    ),
+                )
+        else:
+            # if no pages, treat whole document as single chunk
+            ser_res = my_doc_ser.serialize()
```
PeterStaar-IBM, line 49: I think that here we need to have a parameter that sets the max size of the chunk (measured in chars or string length), otherwise we might get into trouble due to a few poisonous documents.

PeterStaar-IBM (changes requested):

nikhildigde:
can this be merged please?

docling_core/transforms/chunker/page_chunker.py
vagenas: @PeterStaar-IBM, as the premise is not just to chunk the text, but also to provide the doc items contributing to each chunk, adding such a limit would be somewhat more involved.

👉 I therefore propose introducing this capability in its simple form for now to address strictly page-based use cases; a max_chars mechanism could be then be included in the future if deemed necessary.

dolfim-ibm: I agree with the current proposal, and with the intent of enhancing it in the future.

In general, I would be in favour of a maximum size, but we would propose it as a solution which doesn't split doc items, simply add until the limit is reached.

nikhildigde: @vagenas , I tried this locally. However I dont see the images part of the metadata. Wouldnt it be good to have?

vagenas: @nikhildigde consistently with our other chunkers, this implementation provides the chunk objects, which:
1. contain the text from the respective items
2. provide the contextualized version thereof, i.e. including the respective section headers (see [docs](https://docling-project.github.io/docling/concepts/chunking/)), and
3. contain references to the respective items, which can be used for getting all relevant metadata

You can resolve things like images from point 3, e.g. see our [visual grounding example](https://docling-project.github.io/docling/examples/visual_grounding).

vagenas:
@PeterStaar-IBM please consider providing an updated review based on my comment [further above](https://github.com/docling-project/docling-core/pull/337#discussion_r2206541697), otherwise this PR appears blocked by the previous review.

docling_core/transforms/chunker/page_chunker.py
nikhildigde: Yes thats what I plan to do, but would be convenient if the references contain also the "picture" item refs of the metadata. And not only picture, everything maybe.

nikhildigde:
Any clue on when the next release is planned? Thank you!

PR #344, opened 2025-07-01
feat: Additional DoclingDocument methods for use in MCP document manipulation

Ryfernandes:
## Feature

A collection of methods for the DoclingDocument data type that enable the creation of more robust document editing tools in docling-mcp. Methods include:

- **Insert_\* methods**: similar to the add_* methods, except that they specify a sibling node rather than a parent node. They create the desired NodeItem and insert it into the document at the specified location (useful for non-sequential document generation/editing)
- **Range manipulation methods**: these methods allow NodeItems to be added/inserted/removed in groups, based either on a list of items, two start/end items, or a DoclingDocument. These methods can be useful in automated manipulation, but may also be useful in eventually allowing for parts of DoclingDocuments to be "re-converted" with different settings and then inserted back into the document
- **Table data manipulation methods**: a few methods added to the class definition of table data to allow rows to be added/inserted based on a list definition of the content in the row. A simple implementation targeted at making the resulting MCP tool similarly simple to use. Other methods allow for removing/popping rows from the table

**! Bug Changes:** While implementing these features, I came across two (what I believe to be) bugs in the existing code that I changed:
- ```_add_sibling```: when ```after``` is not asserted for the insertion, the last element of the stack should be able to be equal to the length of the children list
- ```_pop_item```: the ```+1``` was on the wrong side of the inequality, giving an incorrect check for if the provided NodeItem was the last in the array

### DIFF
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -366,6 +366,152 @@ class TableData(BaseModel):  # TBD
 
         return table_data
 
+    def remove_rows(self, indices: List[int]) -> List[List[TableCell]]:
+        """Remove rows from the table by their indices.
+
+        :param indices: List[int]: A list of indices of the rows to remove. (Starting from 0)
+        """
+        if not indices:
+            return []
+
+        indices = sorted(indices, reverse=True)
+
+        removed_cells = []
+        for row_index in indices:
+            if row_index < 0 or row_index >= self.num_rows:
+                raise IndexError(
+                    f"Row index {row_index} is out of bounds for the current number of rows {self.num_rows}."
+                )
+            removed_cells.append(self.remove_row(row_index))
+
+        return removed_cells
+
+    def pop_row(self) -> List[TableCell]:
+        """Remove and return the last row from the table."""
+        if self.num_rows == 0:
+            raise IndexError("Cannot pop from an empty table.")
+
+        row_index = self.num_rows - 1
+        return self.remove_row(row_index)
+
+    def remove_row(self, row_index: int) -> List[TableCell]:
+        """Remove a row from the table by its index.
+
+        :param row_index: int: The index of the row to remove. (Starting from 0)
+        """
+        if row_index < 0 or row_index >= self.num_rows:
+            raise IndexError(
+                f"Row index {row_index} is out of bounds for the current number of rows {self.num_rows}."
+            )
+
+        start_idx = row_index * self.num_cols
+        end_idx = start_idx + self.num_cols
+        removed_cells = self.table_cells[start_idx:end_idx]
+
+        # Remove the cells from the table
+        self.table_cells = self.table_cells[:start_idx] + self.table_cells[end_idx:]
+
+        # Update the number of rows
+        self.num_rows -= 1
+
+        # Reassign row offset indices for existing cells
+        for index, cell in enumerate(self.table_cells):
+            new_index = index // self.num_cols
+            cell.start_row_offset_idx = new_index
+            cell.end_row_offset_idx = new_index + 1
+
+        return removed_cells
+
+    def insert_rows(
+        self, row_index: int, rows: List[List[str]], after: bool = False
+    ) -> None:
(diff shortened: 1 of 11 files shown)


PeterStaar-IBM:
@Ryfernandes Looks very good, but it would be great to add some tests for coverage.

docling_core/types/doc/document.py
```diff
@@ -2788,6 +2933,1242 @@ def add_form(
...
+
+        # Insert the new group into the document structure
+        new_ref = group.get_ref()
+
+        success = self.body._add_sibling(
+            doc=self, stack=stack, new_ref=new_ref, after=after
+        )
+
+        if not success:
+            self._pop_item(item=group)
```
ceberam: We may want to align the behavior of public manipulation methods when the operation fails, specially if we return the item. In `append_child_item` we raise a `ValueError`. @vagenas @cau-git feel free to comment on this.

docling_core/types/doc/document.py
```diff
@@ -2788,6 +2933,1242 @@ def add_form(
...
+        self.insert_node_items(
+            sibling=sibling, node_items=node_items, doc=doc, after=after
+        )
+
+    def add_document(
+        self,
+        doc: "DoclingDocument",
+        parent: Optional[NodeItem] = None,
+    ) -> None:
+        """Adds the content from a DoclingDocument to this document under a specific parent.
```
ceberam: You could specify that the content is from the **body** of the DoclingDocument, as in `insert_document`

ceberam (changes requested):
Thanks for including this manipulation operations, which can be useful for the agents indeed. Also, fixing the bugs and removing redundant code is highly appreciated.
Some small comments for improvement:
- Reusable code:
  - We can save some double checks in certain methods. For instance: `remove_rows` could have all the logic of removing the rows and then `remove_row` could simply call `self.removes_row([row_index])`
  - There is quite a few common code at the beginning of many `insert_*` functions. You could have a private method and reuse it in all of them. 
  - Check whether it is worth removing duplicate code between each `insert_` and `add_` method pair or not. For instance, `add_title` calling `insert_title`. I have my doubts on this.
- On the docstrings:
  - Add the description of the returned field (`:return:`)
  - Add the function description of the methods you grouped as _Node Item Insertion Methods_
- Add regression tests

docling_core/types/doc/document.py
Ryfernandes: Yes, that definitely makes sense

docling_core/types/doc/document.py
Ryfernandes: Yes, I agree here as well. I originally did not raise a ValueError, since it was not done in _insert_item_at_stack. However, I agree that if we do not raise a ValueError, there is no easy way for the user to tell if their item was inserted or not. The refactoring of the code that I did means that if we want to remove the ValueError, we only need to delete one line each in _insert_item_at_stack and _insert_in_structure

Ryfernandes:
@PeterStaar-IBM @ceberam 

Thank you very much for the reviews—your detailed comments were helpful for me to effectively implement your proposed changes. 

I implemented almost all of the proposed changes with the exception of bridging the add_* and insert_* methods, as the add_* methods are currently prepared to handle cases where a parent does not have any children, while the insert_* methods are not (need a sibling to exist under a parent). I believe that this could be a worthwhile refactor in the future, but it may be better to do in a separate pr, as I anticipate the best way to do this would be to create private _insert_* methods that allow the specification of a parent instead of a sibling. Then, both public groups could abstract their logic down to these private methods (making alignment of error handling and other standards easier). If you like this implementation better than the current version, please let me know, and I can either implement it as a separate, follow-up pr to this one, or all-in-one on top of this current pr.

Aside from that, all other changes have been implemented as you specified, and the test coverage should be up to par (let me know if it isn't though, as I am fairly new to unit tests)

ceberam (approved):
Excellent work @Ryfernandes 🏆  Thanks for your contribution

PR #346, opened 2025-07-02
Extending the usage of BoundingRectangle

rickymaggio02:
This PR introduces support for representing bounding boxes using 4-vertex convex polygons instead of axis-aligned rectangles. This allows for more accurate localization of elements within a document, especially in cases where the document is photographed under imperfect conditions, such as misalignment, bending, or stretching of the paper.

While the class name BoundingRectangle is currently used, it may be misleading and should potentially be renamed to better reflect the generalized shape (e.g., BoundingPolygon). 

To support this new structure in SmolDocling, a dedicated localization token <rec_ has been added.

Significant effort has gone into ensuring backward compatibility within the Prov class and related methods, so existing workflows remain unaffected.


cau-git:
superseded by #348

PR #347, opened 2025-07-02
fix: BoundingRectangle angle computation when in `CoordOrigin.TOPLEFT`

ClemDoum:
# Bug description

When in `CoordOrigin.TOPLEFT`, the computation of the `BoundingRectangle.angle` have to be adapted since the angle between the (x, y), axis is `3*pi / 2` and not `pi / 2`. 

If `BoundingRectangle.angle` describes the angle between (`r_0`, `r_1`) and the `x` axis, angle computation have to be adapted in the `CoordOrigin.TOPLEFT` case:

IMG_9545

### Fixed
- fixed the `BoundingRectangle.angle`  computation in the case of the `CoordOrigin.TOPLEFT` case


cau-git (approved):

PeterStaar-IBM (approved):
🎖️ Brilliant work!

PR #360, opened 2025-07-24
feat: Key-value visualizer

cau-git:


docling_core/types/doc/document.py
```diff
@@ -5543,27 +5557,61 @@ def get_visualization(
         self,
         show_label: bool = True,
         show_branch_numbering: bool = False,
+        visualizer: str = "reading_order",
```
vagenas: - let's call this something like `viz_mode` (as "visualizer" typically refers to the object) and
- perhaps make it an enum or `Literal` that reflects the options

docling_core/types/doc/document.py
```diff
@@ -57,6 +68,9 @@
 from docling_core.types.doc.tokens import _LOC_PREFIX, DocumentToken, TableToken
 from docling_core.types.doc.utils import relative_path
 
+if TYPE_CHECKING:
+    from docling_core.transforms.visualizer.base import BaseVisualizer
```
vagenas: Perhaps this can be done like for the other visualizer types, i.e. within `get_visualization()`.

dolfim-ibm (approved):
lgtm

PR #361, opened 2025-07-29
fix: referenced artifacts relative to the document location

dolfim-ibm:
This PR is fixing two issues related to the creation of referenced images.

### Relative paths

The management of input directory was only interpreting "non-absolute directories" relative to the current working directory.

It failed to work in the actual setup it was designed, i.e. the artifacts location relative to the target document output.

### Loading a docling document with embedded pages images

The logic saving the picture images relied on the old version, before the page images were introduced.

It failed when doing
- save a DoclingDocument with embedded images
- load the DoclingDocument
- save as markdown/html with referenced images
    - no image was found and exported

Refs https://github.com/docling-project/docling-serve/issues/245

### DIFF
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -1373,11 +1373,11 @@ class PictureItem(FloatingItem):
         )  # Encode to Base64 and decode to string
         return img_base64
 
-    def _image_to_hexhash(self) -> Optional[str]:
+    def _image_to_hexhash(self, img: Optional[PILImage.Image]) -> Optional[str]:
         """Hexash from the image."""
-        if self.image is not None and self.image._pil is not None:
+        if img is not None:
             # Convert the image to raw bytes
-            image_bytes = self.image._pil.tobytes()
+            image_bytes = img.tobytes()
 
             # Create a hash object (e.g., SHA-256)
             hasher = hashlib.sha256(usedforsecurity=False)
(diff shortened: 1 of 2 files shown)


docling_core/types/doc/document.py
```diff
@@ -1373,11 +1373,11 @@ def _image_to_base64(self, pil_image, format="PNG"):
         )  # Encode to Base64 and decode to string
         return img_base64
 
-    def _image_to_hexhash(self) -> Optional[str]:
+    def _image_to_hexhash(self, img: Optional[PILImage.Image]) -> Optional[str]:
```
ceberam: Since this has become static, what about the following?

```suggestion
    @staticmethod
    def _image_to_hexhash(img: Optional[PILImage.Image]) -> Optional[str]:
```

docling_core/types/doc/document.py
```diff
@@ -4116,16 +4116,10 @@ def _with_pictures_refs(
...
-                    if (
-                        item.image is not None
-                        and isinstance(item.image.uri, AnyUrl)
-                        and item.image.uri.scheme == "data"
-                        and item.image.pil_image is not None
-                    ):
-                        img = item.image.pil_image
-
-                        hexhash = item._image_to_hexhash()
+                        hexhash = item._image_to_hexhash(img)
```
ceberam: If we proceed with the solution above:

```suggestion
                        hexhash = PictureItem._image_to_hexhash(img)
```

nikos-livathinos (approved):

PR #363, opened 2025-08-07
fix path generation in case of relative file paths without artifact directory

geoHeil:
closes https://github.com/geoHeil/docling-core


docling_core/types/doc/document.py
```diff
@@ -4525,23 +4525,30 @@ def save_as_html(
...
-        if artifacts_dir.is_absolute():
+            # Default case: create an '_artifacts' directory alongside the file.
+            final_artifacts_dir = filename.with_name(filename.stem + "_artifacts")
+        else:
+            if isinstance(artifacts_dir, str):
+                artifacts_dir = Path(artifacts_dir)  
+            if artifacts_dir.is_absolute():
+                final_artifacts_dir = artifacts_dir
+            else:
+                final_artifacts_dir = filename.parent / artifacts_dir
```
vku-ibm, line 4546: I'm not sure about this line. Depending on from where this function is called, the use case can be different:
- for example, if called from `save_as_json()` with image `referenced` flag, then one can have a use-case where images are stored completely separately from the json document itself, but path is still relative
- the other example is when both `save_as_json()` and `save_as_markdown()` used, with image `referenced` flag, one could use the same artifact directory to store images and avoid duplication. To clarify, an example would be `filename="../results/json/my_doc.json"` and `filename="../results/md/my_doc.md"` with a shared reference images `artifacts_dir="../results/ref_images/my_doc/"`

vku-ibm: My proposal would be to not modify `artifacts_dir` in any way if value is provided.

PR #381, opened 2025-09-03
feat: rendering of checkboxes in markdown

ErikApption:
This PR is to address https://github.com/docling-project/docling/issues/2167

Checkboxes are labelled correctly by docling but are not rendered in the markdown.

- This PR prepends `[x]` or `[ ]` to any text that is labelled as checkbox through `markdown.py` serializer.
- A new unit test is added and follows other unit tests. The test is based on [this form](https://www.tbs-sct.canada.ca/tbsf-fsct/350-103-eng.asp)
- No regression error on other tests


dolfim-ibm:
Interesting. More a curiosity: Do we know if the [marko](https://github.com/frostming/marko) library we use in Docling to parse markdown will understand these checkboxes?

docling_core/transforms/serializer/markdown.py
```diff
@@ -139,6 +140,10 @@ def serialize(
             text = item.text
             processing_pending = True
 
+        if item.label == DocItemLabel.CHECKBOX_SELECTED:
+            text = f"[x] {text}"
```
PeterStaar-IBM: I think you need `- [x]` and not [x]

docling_core/transforms/serializer/markdown.py
```diff
@@ -139,6 +140,10 @@ def serialize(
...
+        if item.label == DocItemLabel.CHECKBOX_UNSELECTED:
+            text = f"[ ] {text}"
```
PeterStaar-IBM: I think you need `- [ ]` isntead of `[ ]`

PeterStaar-IBM (changes requested):

docling_core/transforms/serializer/markdown.py
ErikApption: yes likely more universal

PeterStaar-IBM:
> > @ErikApption Could you do the fixes we requested?
> 
> Done. Alternatively was wondering about making the checkbox renders as parameters the same way there is a parameter for page break but I think this will be a better ootb experience

ok now you will need to rerun the test to generate the update ground-truth files and after that run the `pre-commit` (`uv run pre-commit --all-files`)

ErikApption:
> > > @ErikApption Could you do the fixes we requested?
> > 
> > 
> > Done. Alternatively was wondering about making the checkbox renders as parameters the same way there is a parameter for page break but I think this will be a better ootb experience
> 
> ok now you will need to rerun the test to generate the update ground-truth files and after that run the `pre-commit` (`uv run pre-commit --all-files`)

Updated the truth files. All unit tests are passing again

PeterStaar-IBM:
> > > > @ErikApption Could you do the fixes we requested?
> > > 
> > > 
> > > Done. Alternatively was wondering about making the checkbox renders as parameters the same way there is a parameter for page break but I think this will be a better ootb experience
> > 
> > 
> > ok now you will need to rerun the test to generate the update ground-truth files and after that run the `pre-commit` (`uv run pre-commit --all-files`)
> 
> Updated the truth files. All unit tests are passing again

yes, but you also need to run the formatting, you should see,?

```
taa@Munlochy docling-core % uv run pre-commit run --all-files
Black....................................................................Passed
isort....................................................................Passed
autoflake................................................................Passed
MyPy.....................................................................Passed
Flake8...................................................................Passed
Pytest...................................................................Passed
Docs.....................................................................Passed
uv-lock..................................................................Passed
```

ErikApption:
> yes, but you also need to run the formatting, you should see,?

Got it! I was missing the `run` twice from your previous pre-commit command

PeterStaar-IBM (approved):
lgtm!

dolfim-ibm (approved):

PeterStaar-IBM (approved):

PR #384, opened 2025-09-09
feat: Introduction of fillable TableCell

maxmnemonic:
Added property `fillable` to `TableCell`, to represent semantics of fillable table cell


docling_core/types/doc/document.py
vagenas, line 1: Can you also bump the (minor) version of `CURRENT_VERSION` in this file?

maxmnemonic: updated to 1.7.0

PR #389, opened 2025-09-22
feat: add summary field and serializer

PeterStaar-IBM:


docling_core/experimental/serializer/markdown_summary.py
```diff
@@ -0,0 +1,293 @@
...
+
+
+class MarkdownSummaryMode(str, Enum):
+    """Display mode for document summary output."""
+
+    OUTLINE = "outline"
+    TABLE_OF_CONTENTS = "table_of_contents"
+
+
+class MarkdownSummaryParams(MarkdownParams):
```
dolfim-ibm, line 42: this will now (because of the inheritance) contain a lot of parameters which refer to the actual "content" serialization.

if they could be applied to the summary, then it is good.
if not, maybe we need a separate fresh params model here.

(could be fixed before moving out of experimental)

docling_core/experimental/serializer/markdown_summary.py
```diff
@@ -0,0 +1,293 @@
...
+
+
+class MarkdownSummarySerializer(MarkdownDocSerializer):
+    """Markdown-specific document summary serializer.
+
+    Inherits MarkdownDocSerializer to reuse Markdown formatting/post-processing
+    and sub-serializers; overrides only the parts selection logic.
+    """
+
+    params: MarkdownSummaryParams = MarkdownSummaryParams()
```
dolfim-ibm, line 66: in case we won't inherit `MarkdownSummaryParams(MarkdownParams)`, we might have to override the __init__ and initialize `super()` with a "fixed" `MarkdownParams` which has the required values (or default)

docling_core/experimental/serializer/markdown_summary.py
PeterStaar-IBM: Yes, I would fix before moving out of experimental

PR #393, opened 2025-09-23
feat: add the latex serializer

PeterStaar-IBM:
Let's first merge this: https://github.com/docling-project/docling-core/pull/402

### DIFF
--- a/docling_core/transforms/serializer/latex.py
+++ b/docling_core/transforms/serializer/latex.py
@@ -656,7 +656,9 @@ class LaTeXDocSerializer(DocSerializer):
         # Build optional preamble
         preamble_lines: list[str] = []
         if params.document_class:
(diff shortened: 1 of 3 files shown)


docling_core/transforms/serializer/latex.py
```diff
@@ -0,0 +1,840 @@
...
+    enable_chart_tables: bool = True
+    include_annotations: bool = True
+
+    indent: int = 2  # spaces for nested lists
+
+    # If not None, emitted where page breaks occur (e.g., "\\newpage")
+    page_break_command: Optional[str] = None
+
+    # Escape LaTeX special characters in text
+    escape_latex: bool = True
```
vagenas, line 81: Would be great to have some test coverage for all introduced params (I think e.g. these two are not covered).

PR #395, opened 2025-09-24
feat: added the Azure Document Intelligence

PeterStaar-IBM:


docling_core/transforms/serializer/azure.py
```diff
@@ -0,0 +1,491 @@
...
+    form_serializer: BaseFormSerializer = _AzureFormSerializer()
+    fallback_serializer: BaseFallbackSerializer = _AzureFallbackSerializer()
+
+    list_serializer: BaseListSerializer = _AzureListSerializer()
+    inline_serializer: BaseInlineSerializer = _AzureInlineSerializer()
+
+    params: AzureParams = AzureParams()
+
+    # Accumulator for the Azure-like output
+    azure: Dict[str, Any] = Field(default_factory=dict)
```
vagenas: 1. I'd rather use a Pydantic model for better managing its fields etc. e.g. `_AzureData`
2. I'd make the field private (and possibly rename to something like `_azure_data`

test/test_azure_serializer.py
vagenas, line 1: I'd include an actual JSON output file in the testing.

docling_core/transforms/serializer/azure.py
```diff
@@ -0,0 +1,491 @@
...
+class AzureParams(CommonParams):
+    """Azure-specific serialization parameters.
+
+    - include_words: whether to export page words (not supported; kept for future).
+    """
+
+    include_words: bool = False
+
+
+class _AzureTextSerializer(BaseModel, BaseTextSerializer):
```
vagenas, line 175: Since the PR only uses these classes to populate the target object (currently named `azure`) fields, it makes little sense for them to subclass `Base*Serializer` — these can just be private methods of `BaseDocSerializer`.

Same applies to all `_Azure*Serializer` classes in this file.

PeterStaar-IBM: Yes, but here I want to follow the route all other serializers have.

test/test_azure_serializer.py
PeterStaar-IBM: done

docling_core/transforms/serializer/azure.py
PeterStaar-IBM: yup, good point!

PR #398, opened 2025-10-03
feat: add code chunking functionality

bridgetmcg:
This PR introduces code chunking functionality to `docling-core`, enabling intelligent parsing and chunking of source code files across multiple programming languages. The implementation leverages tree-sitter for accurate parsing and provides language-specific chunkers for Python, TypeScript, JavaScript, Java, and C.

### Core

- CodeChunker - Base abstract class for code chunking with Tree-sitter integration
- Language-specific chunkers - Specialized implementations for 5 major programming languages
- Smart chunk splitting - Automatic splitting of large functions while preserving context

### Language Support
- Python (PythonFunctionChunker) - Functions, classes, imports, module variables
- TypeScript (TypeScriptFunctionChunker) - Functions, classes, interfaces, imports
- JavaScript (JavaScriptFunctionChunker) - Inherits from TypeScript chunker
- Java (JavaFunctionChunker) - Methods, constructors, classes, enums, interfaces
- C (CFunctionChunker) - Functions, structs, macros, preprocessor definitions

### Testing
- test_code_chunker.py - multi-language, real code samples


vagenas (commented):
Nice! Let me share some first thoughts, mostly on how these capabilities can be packaged and exposed:

I see the PR contains various language-specific chunkers, e.g. Java, Python etc.
My recommendation would be:
- to encapsulate these capabilities under a single component, which would include the language detection inside it, and,
- to ensure composability with existing chunkers, instead of introducing a new chunker, I'd rather provide this as a capability pluggable into the `HierarchicalChunker` (good fit because (1) it follows Item boundaries, so CodeItems can be nicely delegated, and (2) is itself composable into the `HybridChunker`). Let us still define the interface specifics, but it could look like an optional kwarg in `HierarchicalChunker.chunk()`, e.g. `code_chunking_strategy`, adhering to a matching interface.

A more minor comment is regarding `CodeDocMeta`, which I see inherits from `BaseMeta`. Some application code may expect to interact with `DocMeta` (also a `BaseMeta` child), so if possible we should best extend that.
This point goes hand in hand with the interface specifics TBD above.

(For now I would focus on the present PR & the points above — the idea of introducing a code backend as per the [docling repo PR](https://github.com/docling-project/docling/pull/2378) can be discussed at a second step.)

vagenas (commented):
Hi @bridgetmcg, many thanks for the new iteration, incorporating feedback from above!

I only have a couple last points from my side (@dolfim-ibm I don't know if you want to add anything):
1. I see the actual integration meanwhile occurs via `HierarchicalChunker` -> `DefaultCodeChunkingStrategy` -> `CodeChunkingStrategyFactory` which returns a `CodeChunker` (subclass of `BaseChunker`). Since the [chunker](https://docling-project.github.io/docling/concepts/chunking/) primitive should be runnable on any document, it could be confusing to expose a component that operates only on e.g. Python, as a "chunker". Ideally one could just satisfy the newly introduced interface/protocol (and not `BaseChunker` which may require additional points), but perhaps the fastest way to address this point is just to mark the `CodeChunker` class hierarchy "internal" by prepending with `_`. Then it's clear all these classes are implementation internals, and users only need to care about the strategy they can optionally pass to `HierarchicalChunker`.
2. the defined `CodeChunkingStrategy(Protocol)` does not seem to be used anywhere — shouldn't this be somewhere in the typing of field `code_chunking_strategy` within `HierarchicalChunker`?
3. to allow for extensibility we support the notion of customizable [serializers](https://docling-project.github.io/docling/concepts/serialization/); while this perhaps makes little sense in `CodeItem`s, for consistency, in `HierarchicalChunker` we should still best use the result from `doc_serializer` (instead of just `item.text`)

Hope that makes sense. Otherwise we can also have a quick call to clarify & finalize.

docling_core/transforms/chunker/hierarchical_chunker.py
```diff
@@ -116,6 +118,45 @@ def check_version_is_compatible(cls, v: str) -> str:
...
+    chunk_type: Optional[str] = Field(default=None)
+
+
+class CodeChunk(BaseChunk):
+    """Data model for code chunks."""
+
+    meta: CodeDocMeta
+
+
+class ChunkType(str, Enum):
```
dolfim-ibm: to avoid users thinking this is a generic `ChunkType`, we could rename the class to `CodeChunkType`.

dolfim-ibm:
Looking at the discussion and the proposed `code_chunking_strategy`, I'm think we want to consider how this will scale to other types. For example, multiple chunking strategies for tables are a recurrent topic.

bridgetmcg:
@vagenas I believe I addressed your comments! Let me know if not. Many thanks!

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "docling-core"
-version = "2.48.4"  # DO NOT EDIT, updated automatically
+version = "2.43.0"  # DO NOT EDIT, updated automatically
```
vagenas: This should be reverted; the version gets bumped by our CD pipeline upon release.

pyproject.toml
```diff
@@ -43,8 +43,15 @@ dependencies = [
...
-    'typer (>=0.12.5,<0.20.0)',
+    'typer (>=0.12.5,<0.17.0)',
     'latex2mathml (>=3.77.0,<4.0.0)',
+    "tree-sitter==0.23.2",
+    "tree-sitter-python==0.23.6",
+    "tree-sitter-c==0.23.4",
+    "tree-sitter-java==0.23.5",
+    "tree-sitter-javascript==0.23.1",
+    "tree-sitter-typescript==0.23.2",
+
```
vagenas: In pyproject.toml we should provide a range for the versions (otherwise we risk dependency hell, plus users get no upstream updates).
Unless there are specific reasons for different ranges, one could go with this:
```suggestion
    'typer (>=0.12.5,<0.20.0)',
    'latex2mathml (>=3.77.0,<4.0.0)',
    "tree-sitter (>=0.23.2,<1.0.0)",
    "tree-sitter-python (>=0.23.6,<1.0.0)",
    "tree-sitter-c (>=0.23.4,<1.0.0)",
    "tree-sitter-java (>=0.23.5,<1.0.0)",
    "tree-sitter-javascript (>=0.23.1,<1.0.0)",
    "tree-sitter-typescript (>=0.23.2,<1.0.0)",
```

pyproject.toml
vagenas: As written on the comment, this should not be edited; version gets bumped by CD pipeline upon release.
```suggestion
version = "2.48.4"  # DO NOT EDIT, updated automatically
```

docling_core/transforms/chunker/hierarchical_chunker.py
```diff
@@ -116,6 +127,51 @@ def check_version_is_compatible(cls, v: str) -> str:
             return _VERSION
 
 
+class CodeDocMeta(DocMeta):
+    """Data model for code chunk metadata."""
+
+    doc_items: Optional[list[DocItem]] = Field(default=None, alias=_KEY_DOC_ITEMS)
```
vagenas: Let's add the schema name here like in DocMeta to provide a hint for deserialization [1].

That requires exposing the class name on `chunker/__init__.py`.
I would actually also add CodeChunk there for consistency.

[1]
```suggestion
    schema_name: Literal["docling_core.transforms.chunker.CodeDocMeta"] = Field(  # type: ignore[assignment]
        default="docling_core.transforms.chunker.CodeDocMeta",
        alias=_KEY_SCHEMA_NAME,
    )
    doc_items: Optional[list[DocItem]] = Field(default=None, alias=_KEY_DOC_ITEMS)
```

docling_core/transforms/chunker/hierarchical_chunker.py
```diff
@@ -259,3 +363,14 @@ def chunk(
                     ),
                 )
                 yield c
+
+    def _strip_markdown_code_formatting(self, text: str) -> str:
+        """Strip markdown code block formatting from text."""
+        if not text.startswith("```") or not text.endswith("```"):
```
vagenas: These backticks look like a bug, perhaps to best be addressed on a different level.

docling_core/transforms/chunker/hierarchical_chunker.py
```diff
@@ -116,6 +128,55 @@ def check_version_is_compatible(cls, v: str) -> str:
...
+    """Chunk type."""
+
+    FUNCTION = "function"
+    METHOD = "method"
+    PREAMBLE = "preamble"
+    CLASS = "class"
+    CODE_BLOCK = "code_block"
+
+
+class CodeChunkingStrategy(Protocol):
```
vagenas: Can make a BaseCodeChunkingStrategy(ABC) out of this, with abstract method

docling_core/types/doc/labels.py
```diff
@@ -266,14 +266,22 @@ def __str__(self):
...
+        return {
+            cls.PYTHON,
+            cls.JAVA,
+            cls.C,
+            cls.TYPESCRIPT,
+            cls.JAVASCRIPT,
         }
-        return mapping.get(self, None)
+
+    def is_supported_for_chunking(self) -> bool:
```
vagenas: is this not used currently?

docling_core/types/doc/labels.py
vagenas, line 1: To complete this change I would remove the additional Enum (`Language`), e.g. turning it into a simple class, and provide the same functionality in a way that reuses the existing labels from `CodeLanguageLabel`.

bridgetmcg: thanks @vagenas. I addressed this in [this](https://github.com/docling-project/docling-core/pull/398/commits/fc649877f62661b5bb061280b32eadecfdb97546) update.

vagenas (approved):
Hi @bridgetmcg, I did some refactoring & made couple improvements — details in https://github.com/docling-project/docling-core/pull/398/commits/641ea9c9c7d34d2f3fc22f9c04fb42b9789f53ff commit message (also removed the previous change of `doc_items` type in `CodeDocMeta`).

PR #400, opened 2025-10-07
chore: Silencing info logging on "delete_items" by moving it to debug level

maxmnemonic:
Moving log of deleting elements from info level to debug.
This is due to support of rich tables, while constructing them backends (docx, html) manipulate docling document tree, and sometimes need to delete elements, this creates excessive logging.

Note from the user: https://github.com/docling-project/docling/issues/2360#issuecomment-3372942862


dolfim-ibm (approved):
lgtm

PR #408, opened 2025-10-21
feat: add metadata model hierarchy

vagenas:
Main goal has been to establish metadata model that:
- allows both _prescribed_ and _custom_ fields, with a way to distinguish the two
- allows both _common_ and _type-specific_ fields, so that we can further specify the metadata reqs e.g. for PictureItem
- accounts for _serialization_ and _deserialization_, e.g. ideally avoiding serializing additional data if no meta info available

### DIFF
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -1017,7 +1017,7 @@ class SummaryMetaField(BasePrediction):
     text: str
 
 
-# NOTE: should be manually kept in sync with top-level BaseMeta hierarchy fields
+# NOTE: must be manually kept in sync with top-level BaseMeta hierarchy fields
 class MetaFieldName(str, Enum):
     """Standard meta field names."""
 
@@ -1069,7 +1069,7 @@ class PictureClassificationMetaField(_ExtraAllowingModel):
 class MoleculeMetaField(BasePrediction):
     """Molecule metadata field."""
 
-    smiles: str = Field(description="The SMILES representation of the molecule.")
+    smi: str = Field(description="The SMILES representation of the molecule.")
 
 
 class TabularChartMetaField(BasePrediction):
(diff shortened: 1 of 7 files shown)


docling_core/types/doc/document.py
```diff
@@ -941,6 +943,156 @@ class ContentLayer(str, Enum):
...
+class MoleculeMetaField(BasePrediction):
+    """Molecule metadata field."""
+
+    smi: str = Field(description="The SMILES representation of the molecule.")
+
+
+class TabularChartMetaField(BasePrediction):
+    """Tabular chart metadata field."""
+
+    title: str
```
PeterStaar-IBM: let's quickly check on this

docling_core/types/doc/document.py
```diff
@@ -941,6 +943,156 @@ class ContentLayer(str, Enum):
...
+            ):
+                max_conf_pos = i
+                max_conf = pred.confidence
+        return self.predictions[max_conf_pos if max_conf_pos is not None else 0]
+
+
+class MoleculeMetaField(BasePrediction):
+    """Molecule metadata field."""
+
+    smi: str = Field(description="The SMILES representation of the molecule.")
```
PeterStaar-IBM, line 1072: check on convention

docling_core/types/doc/document.py
```diff
@@ -941,6 +943,156 @@ class ContentLayer(str, Enum):
...
+
+class SummaryMetaField(BasePrediction):
+    """Summary data."""
+
+    text: str
+
+
+# NOTE: should be manually kept in sync with top-level BaseMeta hierarchy fields
+class MetaFieldName(str, Enum):
+    """Standard meta field names."""
```
PeterStaar-IBM, line 1022: add docs

docling_core/types/doc/document.py
vagenas: Checked, `smiles` is also fine.

docling_core/types/doc/document.py
vagenas: done

PR #414, opened 2025-11-05
fix: parameterized disable_numparse for MarkdownTableSerializer

lucaselvaggio:
This change introduces the `disable_table_numparse` parameter to `MarkdownParams`, enabling it to be passed through to the `export_to_markdown` function. 
When set to `True`, this flag is forwarded to the `tabulate` function via `disable_numparse`, preventing automatic numeric parsing.

Previously, numeric values containing **trailing zeros** (e.g. "50.000") were automatically parsed and reformatted (e.g. to "50"), resulting in _loss of precision_ in the Markdown table output. With this change, setting `disable_table_numparse=True` preserves the original formatting of numeric strings during Markdown export.

Issue ref: https://github.com/docling-project/docling/issues/1891

By exposing this parameter, the implementation provides greater flexibility for different usage scenarios where numeric formatting preservation is required.


PeterStaar-IBM (dismissed):

cau-git:
It looks like this has been addressed in the meantime, therefore closing. Please re-open if you have new evidence it is still needed.

PR #416, opened 2025-11-07
fix: item numbering in document

glypt:
The PR https://github.com/docling-project/docling/pull/2589 revealed a bug, the caption is having the same number as the table here in https://github.com/docling-project/docling/pull/2589/files#diff-0ff184cc09560c89eb50dc9cf939c40bdf55d372d26823c956b55fb8f6c5e16fR1-R5.

This introduces a global numbering with increment at each item.

### DIFF
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -5668,6 +5668,7 @@ class DoclingDocument(BaseModel):
     ):
         """Export the document to indented text to expose hierarchy."""
         result = []
+        item_counter = 0
 
         def get_text(text: str, max_text_len: int):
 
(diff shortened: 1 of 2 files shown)


ceberam:
Thanks @glypt for your contributions and for suggesting this PR.
However, the issue you described is not an actual bug. The code works as designed.
The indented text serialization has the goal of exposing the document hierarchy in a succinct yet visual manner. In the serialization, we iterate over the document items through the line:

https://github.com/docling-project/docling-core/blob/a54f6f075aaaef5be6048cf3230e14d7b8a59832/docling_core/types/doc/document.py#L5686

The index `i` in the enumeration distinguishes those items and helps us track their order in the Docling document (typically the reading order of the original document).
For `TextItem` objects, we show their indent level and the beginning of the text.
For`TableItem` (a table) and `PictureItem` (a picture), in addition to their indent level, we also show their caption (if it exists), since it may be helpful for identifying that table or picture. It should not be interpreted as a completely distinct node item. 
With this consideration, the example output that you marked as a bug is actually correct:

```
item-0 at level 0: unspecified: group _root_
  item-1 at level 1: section: group sheet: Duck Observations
    item-2 at level 2: caption: Number of freshwater ducks per year
    item-3 at level 2: table with [7x2]
      item-3 at level 3: caption: Number of freshwater ducks per year
```

This shows that the item at index `2` is a caption. Right after that, we find a table at index `3`, which has a certain caption.
The line `item-3 at level 3: caption: Number of freshwater ducks per year` should be read as: _item at index 3 is indented at level 3 and has a caption starting with the following text: Number of freshwater ducks per year_.

If we applied the changes in your PR, the serialization would look like:

```
item-0 at level 0: unspecified: group _root_
  item-1 at level 1: section: group sheet: Duck Observations
    item-2 at level 2: caption: Number of freshwater ducks per year
    item-3 at level 2: table with [7x2]
      item-4 at level 3: caption: Number of freshwater ducks per year
```

and it would therefore _count_ more items than they are in the document, since the caption would be counted as two different items.

Please, let me know if the explanation is reasonable.

PR #419, opened 2025-11-13
fix: DoclingDocument model validator should deal with any raw input

ceberam:
The `transform_to_content_layer` model validator of `DoclingDocument` is of type _before_, since it may need to transform the raw input before the `DoclingDocument` is instantiated.
However, it assumes that the raw input data is a `dict` and this may create issues like described in docling-project/docling#2616.
Since _before_ validators _have to deal with the raw input, which in theory could be any arbitrary object_, a type check has been added.

Resolves https://github.com/docling-project/docling/issues/2616

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Docling Core
 
 [![PyPI version](https://img.shields.io/pypi/v/docling-core)](https://pypi.org/project/docling-core/)
-![Python](https://img.shields.io/badge/python-3.9%20%7C%203.10%20%7C%20%203.11%20%7C%203.12%20%7C%203.13-blue)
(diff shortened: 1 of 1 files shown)


cau-git (approved):

test/test_docling_doc.py
```diff
@@ -1156,6 +1156,14 @@ def test_upgrade_content_layer_from_1_0_0():
     assert doc.version == CURRENT_VERSION
     assert doc.texts[0].content_layer == ContentLayer.FURNITURE
 
+    # test that transform_to_content_layer model validator can handle any data type
+    class ContentOutput(BaseModel):
+        content: str | DoclingDocument
```
vagenas: `Union` needed here for 3.9

ceberam: I have also added a minor update in `README`

cau-git (approved):

Issue #423, opened 2025-11-16
TripletTableSerializer always serializes rich cells in markdown


Vdaleke:
TripletTableSerializer ignores DocumentSerializer.

This is because the text is serialized in the export_to_dataframe method, which doesn't accept a document serializer and then doesn't pass it to the _get_text method for cells.

As a result, the default serializer for Rich Cells, which is Markdown, is used.

PR #425, opened 2025-11-16
fix: rich table triplet serialization

Vdaleke:
Adds `**kwargs` to the `export_to_dataframe` method and passes them to `_get_text` for the table cell.

In `TripletTableSerializer`, the method is called with `doc_serializer`, which solves the problem of incorrect serialization.

Tests for the case have been added.

Resolves https://github.com/docling-project/docling-core/issues/423

### DIFF
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -2172,12 +2172,18 @@ class TableItem(FloatingItem):
 
             return self
 
-    def export_to_dataframe(
+    def export_to_dataframe(self, doc: Optional["DoclingDocument"] = None) -> pd.DataFrame:
+        """Export the table as a Pandas DataFrame."""
+
+        return self._export_to_dataframe_with_options(doc=doc)
+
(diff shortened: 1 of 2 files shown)


docling_core/types/doc/document.py
```diff
@@ -1840,7 +1840,9 @@ def _migrate_annotations_to_meta(self) -> Self:
         return self
 
     def export_to_dataframe(
-        self, doc: Optional["DoclingDocument"] = None
+        self,
+        doc: Optional["DoclingDocument"] = None,
+        **kwargs: Any,
```
ceberam, line 2183: @Vdaleke After discussing it with the team, we prefer not to expose the `kwargs` of a private function (`_get_text`) to the public function `export_to_dataframe`.
Instead, we could define a new optional parameter for the serializer in the function `export_to_dataframe` and then pass it to `_get_text`.

ceberam (changes requested):

PR #430, opened 2025-11-18
feat: convert regions into TableData

maxmnemonic:
Added capability to load TableData.from_regions
Converts regions (lists of BoundingBox): rows, columns, merged cells into table_data structure,
Adds semantics for regions of row_headers, col_headers, row_section.
Tolerates slight overlaps "wobbly" bounding boxes, in some cases missing bounding boxes, etc.

Includes test_regions_to_table.py - that acts as an example


docling_core/types/doc/regions.py
```diff
@@ -0,0 +1,303 @@
...
+    for row_header in row_headers:
+        if bbox.intersection_over_self(row_header) >= 0.5:
+            c_row_header = True
+    for row_section in row_sections:
+        if bbox.intersection_over_self(row_section) >= 0.5:
+            c_row_section = True
+    return c_column_header, c_row_header, c_row_section
+
+
+def _bbox_intersection(a: BoundingBox, b: BoundingBox) -> Optional[BoundingBox]:
```
vagenas: This seems like a useful method to push into the `BoundingBox` class.

docling_core/types/doc/regions.py
```diff
@@ -0,0 +1,303 @@
+#
+# Copyright IBM Corp. 2025 - 2025
+# SPDX-License-Identifier: MIT
+#
+
```
vagenas: This header is not applicable any longer.

docling_core/types/doc/regions.py
```diff
@@ -0,0 +1,303 @@
...
+    for row_section in row_sections:
+        if bbox.intersection_over_self(row_section) >= 0.5:
+            c_row_section = True
+    return c_column_header, c_row_header, c_row_section
+
+
+def _bbox_intersection(a: BoundingBox, b: BoundingBox) -> Optional[BoundingBox]:
+    """Return the intersection of two bounding boxes or ``None`` when disjoint."""
+    if a.coord_origin != b.coord_origin:
+        raise ValueError("BoundingBoxes have different CoordOrigin")
```
vagenas: Wouldn't it be more useful to first normalize/origin-align and then indeed intersect?

docling_core/types/doc/regions.py
```diff
@@ -0,0 +1,303 @@
...
+) -> List[TableCell]:
+    """Returns TableCell. Merged cells are aligned to grid boundaries.
+
+    rows, columns, merges are lists of BoundingBox(l,t,r,b).
+    """
+    rows.sort(key=lambda r: (r.t + r.b) / 2.0)
+    columns.sort(key=lambda c: (c.l + c.r) / 2.0)
+
+    # n_rows, n_cols = len(rows), len(columns)
+
```
vagenas: Can be cleaned up.

docling_core/types/doc/regions.py
```diff
@@ -0,0 +1,303 @@
...
+def regions_to_table(
+    table_bbox: BoundingBox,
+    rows: List[BoundingBox],
+    cols: List[BoundingBox],
+    merges: List[BoundingBox],
+    row_headers: List[BoundingBox] = [],
+    col_headers: List[BoundingBox] = [],
+    row_sections: List[BoundingBox] = [],
+) -> Optional[TableData]:
```
vagenas: Looking at the implementation that seems to be not `Optional`

docling_core/types/doc/regions.py
cau-git: Why is not `BoundingBox.intersection_area_with` used?

maxmnemonic: > intersection_area_with

@cau-git - because we need bounding box of the intersection, not the area

docling_core/types/doc/regions.py
maxmnemonic: @vagenas transforming origin would introduce requirement to know page height in this utility, which is currently unnecessary, as function will work for both types of origin-aligns, and will only complain if two bounding boxes have different origin

Issue #427, opened 2025-11-17
The extract_items_range method generates a document with inconsistent hierarchy


olivierantonelli:
**Bug**

The DoclingDocument.extract_items_range method may create an invalid DoclingDocument when the original document contains tables with RichTableCell cells.

During extract_items_range, GroupItems in the given range are renumbered, but the references to GroupItems in RichTableCell are not updated accordingly which results in an "IndexError: list index out of range" exception.

**Steps to reproduce**

```python
from docling_core.types.doc import DoclingDocument, RichTableCell, TableCell, DocItemLabel, TableData, Formatting

doc = DoclingDocument(name="sample")

list_before_table = doc.add_list_group(name="A list")
doc.add_list_item(text="First item", parent=list_before_table)

table_item = doc.add_table(data=TableData(num_rows=2, num_cols=1))

cell = TableCell(text="R1C1", start_row_offset_idx=0, end_row_offset_idx=1, start_col_offset_idx=0, end_col_offset_idx=1)
doc.add_table_cell(table_item=table_item, cell=cell)

rich_group = doc.add_group(name="rich_cell_group", parent=table_item)
rich_cell = RichTableCell(start_row_offset_idx=1, end_row_offset_idx=2, start_col_offset_idx=0, end_col_offset_idx=1, ref=rich_group.get_ref(),)
doc.add_text(label=DocItemLabel.TEXT,text="R2C1", parent=rich_group, formatting=Formatting(italic=True))
doc.add_table_cell(table_item=table_item, cell=rich_cell)

# Here we create a new Docling document with the TableItem and all his childs
doc_with_only_the_table = doc.extract_items_range(start=doc.tables[0], end=doc.tables[0])

# export_to_markdown raises an "IndexError: list index out of range" exception
doc_with_only_the_table.export_to_markdown()
```

Here, without the export_to_markdown, 'doc_with_only_the_table' references '#/groups/1' which no longer exist:

```python
doc_with_only_the_table.tables[0].data.grid[1][0].ref
```
```text
RefItem(cref='#/groups/1')
```
```python
list(doc_with_only_the_table.iterate_items(with_groups=True))
```
```text
[(GroupItem(self_ref='#/body', parent=None, children=[RefItem(cref='#/tables/0')], content_layer=<ContentLayer.BODY: 'body'>, meta=None, name='_root_', label=<GroupLabel.UNSPECIFIED: 'unspecified'>),
  0),
[...]

PR #438, opened 2025-11-26
fix: correct extract_items_range method for rich tables

capoan:
The purpose of this PR is to solve issue #427 (rich table items extraction issue).

A way to extract rich table items from a DoclingDocument is given in #427 by copying the items as-is and then using _DocIndex.index. Here, we propose another solution:

- We add a "root" parameter to _DocIndex.index to extract only this "root" element (and all his descendants) from a document.
- We use this new "root" parameter of _DocIndex.index in the DoclingDocument method extract_items_range to support extracting a range of items, including rich table items.

### DIFF
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -4293,12 +4293,15 @@ class DoclingDocument(BaseModel):
                 "Start NodeItem must come before or be the same as the end NodeItem in the document structure."
             )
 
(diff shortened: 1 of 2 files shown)


jponf:
Hi, I'm also hitting this issue when extracting rich tables. 

This has been open for 6+ months — is there anything blocking the merge?

PR #447, opened 2025-11-28
feat: Configurable output for doctags

nassarofficial:
Enables toggling document element types (OCR text, tables, pictures, charts, code, formulas) via include_* flags in DocTagsParams. Serializers filter elements based on these flags, allowing selective export by task type.


docling_core/transforms/serializer/doctags.py
```diff
@@ -90,6 +104,61 @@ def _get_delim(params: DocTagsParams) -> str:
     return delim
 
 
+def create_task_filtered_params(task_list: list[str], **kwargs) -> DocTagsParams:
```
vagenas, line 107: This preprocessing code best belongs to where these task type values "ocr", "layout" etc are governed.

dco_fix
vagenas, line 1: File to be removed.

PR #454, opened 2025-12-10
fix: drop malformed bounding boxes

b-hahn:
This change handles the case where bounding boxes predicted by VLMs have a width or a height <=0. Previously, this would cause the pipeline to crash while saving the PIL image. Now, malformed bounding boxes are simply ignored, resulting in an empty provenance field while keeping the extracted text. 

Fixes docling-project/docling#2763

### DIFF
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -5300,7 +5300,7 @@ class DoclingDocument(BaseModel):
             if len(coords) == 4:
                 l, t, r, b = map(float, coords)
                 eps = 1 / 500
-                # Ignore bounding boxes with width or height of <1e-3, including cases where l>r or t>b.
+                # Ignore bounding boxes with width or height of < 0.2% of the image width or height.
                 if r - l < eps or b - t < eps:
                     return None
                 return BoundingBox(l=l / 500, t=t / 500, r=r / 500, b=b / 500)
@@ -5318,13 +5318,12 @@ class DoclingDocument(BaseModel):
             if caption is not None:
                 caption_content = caption.group(1)
                 bbox = extract_bounding_box(caption_content)
-                if bbox is not None:
-                    caption_text = extract_inner_text(caption_content)
-                    caption_item = doc.add_text(
-                        label=DocItemLabel.CAPTION,
-                        text=caption_text,
-                        parent=None,
-                    )
+                caption_text = extract_inner_text(caption_content)
+                caption_item = doc.add_text(
+                    label=DocItemLabel.CAPTION,
+                    text=caption_text,
+                    parent=None,
+                )
             else:
                 caption_item = None
                 bbox = None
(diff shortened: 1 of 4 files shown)


simon376:
I think I found another issue that might relate to this, unfortunately I can't provide a stack trace since it is hard to reproduce and I didn't fully log everything on the server, and I couldn't find this error message in the code yet. Maybe you can have a look at this, too:
`Pipeline VlmPipeline failed: Coordinate 'lower' is less than 'upper'`

docling_core/types/doc/document.py
```diff
@@ -5366,12 +5370,13 @@ def extract_caption(
             if caption is not None:
                 caption_content = caption.group(1)
                 bbox = extract_bounding_box(caption_content)
-                caption_text = extract_inner_text(caption_content)
-                caption_item = doc.add_text(
-                    label=DocItemLabel.CAPTION,
-                    text=caption_text,
-                    parent=None,
-                )
+                if bbox is not None:
```
cau-git: I would remove this condition so captions are created also when the `bbox` is not present, as before.

docling_core/types/doc/document.py
```diff
@@ -5634,17 +5639,20 @@ def _add_text(
...
-                            bbox=common_bbox,
-                            pg_width=pg_width,
-                            pg_height=pg_height,
-                            page_no=page_no,
-                            tag_name=item_tag,
-                            doc_label=tag_to_doclabel.get(item_tag, DocItemLabel.TEXT),
-                            doc=doc,
-                            parent=inline_group,
-                        )
+                        if common_bbox is not None:
```
cau-git: I would remove this condition so that `_add_text` is called also with a `None` bounding box, as before.

cau-git (commented):
@b-hahn @kklein just reflecting the desired solution below. Let's first go only with your change in `extract_bounding_box`, then observe before introducing conditions that could potentially skip valid content.

docling_core/types/doc/document.py
b-hahn: Done.

docling_core/types/doc/document.py
b-hahn: Done.

ceberam (dismissed):
LGTM, it works as expected, thus approved.

I would only add the following suggestion. Turn this variable into a module constant, so that we _remember_ the value choice we made. We can also add some docstrings. Something like:

```python
_DOCTAGS_BBOX_MIN_DIMENSION: Final = 1 / 500
"""Minimum bounding box dimension threshold for DocTags import.

This constant represents the minimum width or height (as a fraction of the image
dimensions) that a bounding box must have to be considered valid during DocTags
import. The value ``1/500 = 0.002 = 0.2%`` means that any bounding box with width or
height less than 0.2% of the image width or height will be discarded.
"""
```

cau-git (dismissed):

ceberam:
@b-hahn we just need your sign-off in the second commit (af71bb5d4ed5949fe68c0f91204fc46849d2e8c2) to merge this PR.

ceberam:
> @b-hahn we just need your sign-off in the second commit ([af71bb5](https://github.com/docling-project/docling-core/commit/af71bb5d4ed5949fe68c0f91204fc46849d2e8c2)) to merge this PR.

@b-hahn ^^^ were you able to have a look at this? ^^^
Also, consider the remark here https://github.com/docling-project/docling-core/pull/454#pullrequestreview-3744013956

realech:
I am running into this issue as well, is there anything blocking the merge? Would love to see it merged, thank you! @ceberam @b-hahn

ceberam:
Thanks @realech for the heads-up.
Unfortunately, since the last discussion, some conflicts have appeared.
@b-hahn would you mind rebasing to `main` and solve the conflicts?

carev01:
I'd like to reported that I tested this fix in a fork, and it effectively fixed the issue reported here:
https://github.com/docling-project/docling/issues/2763 (SystemError: tile cannot extend outside image caused by VLM zero-area predictions (Integer Rounding Edge Case).
It would be great to see this PR merged into main, so I can drop my fork.

Issue #453, opened 2025-12-10
Migrate formatter and linter to ruff


dolfim-ibm:
This repo is the last one in the Docling ecosystem which didn't yet migrate to `ruff` as formatter and linter.

Taking [docling](https://github.com/docling-project/docling), [docling-jobkit](https://github.com/docling-project/docling-jobkit) as example one can easily migrate this repo as well.

The changes will mostly be regarding:

1. The pyproject.toml config
2. The pre-commit.yaml config

PR #456, opened 2025-12-10
style: replace black, isort, flake8 and autoflake with ruff

schwarbf:
This PR aims to close the open issue [453](https://github.com/docling-project/docling-core/issues/453).

I tried to do as little code changes as possible. I did the following: 
- I started taking the ruff config from the `docling` repository
- Using ruff now as a replacement for `black`, `isort`, `autoflake` and `flake8`
- Ran `ruff format` (formatted approx. 30 files)  and `ruff check --fix` (fixed approx. 50 warnings)
- Excluded the `test/data` from ruff
- Added some rules to the `ignore` list that would have affected > 3 changes. Feel free to remove any again to see possible changes.
- Set the line length again to `120` in `pyproject.toml` (was `120` in `.flake8` file, but `88` in `pyproject.toml`)

FYI: When checking which commit added `ruff` to docling, you can see that [5458a88
](https://github.com/docling-project/docling/commit/5458a884649ae0887c1659d9075fc4a8d9f434d9#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711) also changed a significant number of files and added some rules to the `ignore` list.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -161,11 +161,9 @@ select = [
 ignore = [
   "C408",   # Unnecessary `dict()` call (rewrite as a literal)
   "E501",   # Line too long, handled by ruff formatter
-  "E203",   # whitespace-before-punctuation
   "E741",   # Ambiguous variable name: `l`
   "D107",   # "Missing docstring in __init__",
   "F401",   # imported but unused; consider using `importlib.util.find_spec` to test for "
-  "F811",   # "redefinition of the same function"
   "PL",     # Pylint
   "RUF005", # Consider ... instead of concatenation
   "RUF012", # Mutable Class Attributes


docling_core/transforms/chunker/code_chunking/code_chunk.py
```diff
@@ -28,12 +28,12 @@ class CodeDocMeta(DocMeta):
...
-    sha256: Optional[int] = Field(default=None)
-    start_line: Optional[int] = Field(default=None)
-    end_line: Optional[int] = Field(default=None)
-    end_line_signature: Optional[int] = Field(default=None)
+    part_name: str | None = Field(default=None)
+    docstring: str | None = Field(default=None)
+    sha256: int | None = Field(default=None)
+    start_line: int | None = Field(default=None)
+    end_line: int | None = Field(default=None)
+    end_line_signature: int | None = Field(default=None)
```
ceberam: Was this suggested by Ruff?
We still support 3.9 (it may change at some point) and the `|` syntax is only valid from 3.10. We should therefore keep `typing.Optional`

schwarbf: Okay I see. It's weird since I had actually set :

```toml
[tool.ruff]
target-version = "py39"
```

Anyways, I will revert this part.

ceberam: Thanks! However, I've seen that the syntax is still in other places, like in `test/test_page_chunker.py`

schwarbf: Can you please check again? I found 2 remaining spots, where the syntax `dtype | None` was used instead of `Optional[dtype]`.  The respective rule `UP045` is also in the ignore list such that no errors are raised in the future.

cau-git (approved):

PR #457, opened 2025-12-11
feat: idoctags serialization and deserialization matching the iso proposal

PeterStaar-IBM:
# IDocTags Serialization Implementation

## Overview

  Implements bidirectional serialization between DoclingDocument and IDocTags format—a specialized XML-based markup language for structured document representation with geometric and semantic annotations.

## Serialization Features

  Core Capabilities:
  - Two output modes: HUMAN_FRIENDLY (indented, readable) and LLM_FRIENDLY (compact, tokenizer-optimized)
  - Geometric encoding: Bounding boxes quantized to configurable resolution (default: 512×512) via <location> tokens
  - Semantic markup: Rich vocabulary covering titles, headings, text, captions, lists, forms, tables, pictures, code, formulas
  - OTSL table structure: Optimized Table Sequence Language with cell types (fcel/ecel, ched/rhed/corn, lcel/ucel/xcel, nl)
  - Content control: add_content parameter allows structure-only serialization (omits text while preserving tags)
  - XML compliance mode: Optional escaping of special characters (&, <, >) for valid XML output
  - Deserialization: Round-trip support to reconstruct DoclingDocument from IDocTags

## Current Test Coverage

  - Vocabulary helpers (create_closing_token validation, self-closing token handling)
  - Content suppression correctness (captions, table cells, list items)
  - Multi-mode serialization (with/without content, human/LLM-friendly)
  - Metadata serialization and XML escaping
  - Round-trip edge cases

## Outstanding Work (with FIXME's)

  1. Inline groups with multi-provenance (idoctags.py:1093) — Current split-per-provenance logic may need enhancement for inline groups
  2. Checkbox token generation (lines 1174, 1179) — Add dedicated create_selected_token() method to vocabulary
  3. Catch-all label handling (lines 1183-1184) — Refine mapping for EMPTY_VALUE, HANDWRITTEN_TEXT, PARAGRAPH, etc.; EMPTY_VALUE may need FormItem representation
  4. OTSL cell logic verification (lines 1564, 1570) — Validate rowstart/colstart conditions for UCEL/LCEL tokens
  5. TABLE vs DOCUMENT_INDEX distinction (line 1602) — Check label to emit correct floating group type
  6. We still need to do the FormItems
  7. we still need to take care of the `<thread id="int">` and page-breaks. This will likely need some updates to the BaseSerializer. As such, I want to not include it in this PR.
  8. We have a few tests in `./test/test_deserializer_idoctags.py` that do not pass. Currently, they are skipped but we need to make them work.

## Testing

Dump Mode Usage

Serialize DoclingDocuments from HuggingFace datasets to IDocTags format and generate a validation report:

```
python examples/convert_to_idoctags.py --mode dump [--config CONFIG.json] [--limit N]
```

What it does:
  - Loads documents from a HuggingFace dataset (default: docling-project/doclaynet-set-a)
  - Serializes them to IDocTags in multiple variants (LLM/human-friendly, with/without content, XML-compliant)
  - Generates an Excel report (./scratch/idoctags_report.xlsx) tracking success/failure for each serialization mode
  - Optionally limits processing to first N documents with --limit

Config file:

If --config is omitted, a default config (idoctags_dump_config.json) is auto-generated. Key settings: dataset_name, dataset_subset, output_dir, report_path, limit.

Use --write-default-config to generate the config template without running the dump.
  
The result of,

```
uv run python ./examples/convert_to_idoctags.py --mode dump --config ./idoctags_dump_config.json
```

is

```
Wrote report (Excel via pandas) to: scratch/idoctags_report.xlsx
Overview summary:
 - Total processed: 3544
 - Loaded DoclingDocument: 3544
 - Serialized IDocTags (human_friendly, xml_compliant=True, content=True): 3529
 - Serialized IDocTags (human_friendly, xml_compliant=True, content=False): 3544
 - Serialized IDocTags (human_friendly, xml_compliant=False, content=True): 3034
 - Serialized IDocTags (human_friendly, xml_compliant=False, content=False): 3544
 - Serialized IDocTags (llm_friendly, xml_compliant=True, content=True): 3544
 - Serialized IDocTags (llm_friendly, xml_compliant=True, content=False): 3544
 - Serialized IDocTags (llm_friendly, xml_compliant=False, content=True): 3544
 - Serialized IDocTags (llm_friendly, xml_compliant=False, content=False): 3544
 - Serialized HTML: 3541
 ```

### DIFF
--- a/examples/convert_to_idoctags.py
+++ b/examples/convert_to_idoctags.py
@@ -29,7 +29,7 @@ import numpy as np
 
 # In order to download **before** the datasets library, run
 # 
-# HF_HUB_DISABLE_XET=1 hf download --repo-type dataset "docling-project/canva-set-a"
+# HF_HUB_DISABLE_XET=1 hf download --repo-type dataset "{hf-repo-id}"
 #
 
 def update_tokenizer(tokenizer: PreTrainedTokenizerBase, verbose: bool = False) -> PreTrainedTokenizerBase:

### CODE REVIEW

File: docling_core/experimental/idoctags.py [hunk 1]
```diff
@@ -1,199 +1,933 @@
...
+    METADATA = "metadata"
+    GEOMETRIC = "geometric"
+    TEMPORAL = "temporal"
+    SEMANTIC = "semantic"
+    FORMATTING = "formatting"
+    GROUPING = "grouping"
+    STRUCTURAL = "structural"
+    CONTENT = "content"
+    BINARY_DATA = "binary_data"
+    CONTINUATION = "continuation"
```
vagenas, line 251: This seems to be unused, I would therefore remove it.

(The concept is anyway reflected by the categorization of the actual `IDocTagsToken` values further below).

docling_core/experimental/idoctags.py
```diff
@@ -1,199 +1,933 @@
...
+    | 52 | Continuation Tokens | `thread` | Yes | Yes | `id` |
+      Continuation marker for split content; reuse same `id` across parts. |
+    | 53 |  | `h_thread` | Yes | Yes | `id` | Horizontal stitching marker for split tables; reuse same `id`. |
+    | 54 | Binary Data Tokens | `base64` | No | No | — | Embedded binary data (base64). |
+    | 55 |  | `uri` | No | No | — | External resource reference. |
+    | 56 | Content Tokens | `marker` | No | No | — | List/form marker content. |
+    | 57 |  | `facets` | No | No | — | Container for application-specific derived properties. |
+    | 58 | Structural Tokens (Form) | `key` | No | No | — | Form item key (child of `form_item`). |
+    | 59 |  | `value` | No | No | — | Form item value (child of `form_item`). |
+    """
```
vagenas, line 329: I would remove the table from the code itself.

Later, one can just add a link/reference to the actual standard table.

docling_core/experimental/idoctags.py
```diff
@@ -1,199 +1,933 @@
...
+        IDocTagsToken.CORN: True,
+        IDocTagsToken.SROW: True,
+        IDocTagsToken.LCEL: True,
+        IDocTagsToken.UCEL: True,
+        IDocTagsToken.XCEL: True,
+        IDocTagsToken.NL: True,
+        # Continuation markers
+        IDocTagsToken.THREAD: True,
+        IDocTagsToken.H_THREAD: True,
+    }
```
vagenas: In case we keep these low-level "rules" for now: this seems to include only the self-closing ones (i.e. value is always True), so why not just make it a set?

docling_core/experimental/idoctags.py
```diff
@@ -441,79 +1412,1068 @@ def serialize(
...
+
         return create_ser_result(text=text_res, span_source=parts)
+
+    @override
+    def requires_page_break(self):
+        """Return whether page breaks should be emitted for the document."""
+        return self.params.add_page_break
+
+
+class IDocTagsDocDeserializer(BaseModel):
```
vagenas, line 2158: Does this need to be a Pydantic model?

docling_core/experimental/idoctags.py
```diff
@@ -441,79 +1412,1068 @@ def serialize(
...
+            code_text, lang_label = self._extract_code_content_and_language(el)
+            if not code_text.strip():
+                return
+            item = doc.add_code(
+                text=code_text,
+                code_language=lang_label,
+                parent=parent,
+                prov=(prov_list[0] if prov_list else None),
+            )
+            for p in prov_list[1:]:
```
vagenas, line 2297: Why are these appended separately and not the whole list assigned above like `prov=prov_list or None`?

Same applies to 7 occurrences further below.

docling_core/experimental/idoctags.py
```diff
@@ -441,79 +1412,1068 @@ def serialize(
...
+        caption = self._extract_caption(doc=doc, el=el)
+        otsl_el = self._first_child(el, IDocTagsToken.OTSL.value)
+        if otsl_el is None:
+            doc.add_table(data=TableData(), caption=caption, parent=parent)
+            return
+        # Extract table provenance from <otsl> leading <location/> tokens
+        tbl_provs = self._extract_provenance(doc=doc, el=otsl_el)
+        inner = self._inner_xml(otsl_el)
+        # Remove any location tokens from the OTSL content before parsing
+        inner = re.sub(r"<\s*location\b[^>]*/\s*>", "", inner)
```
vagenas: Why use regex when we already have the parsed content in `el`? 

Applies to other regex usage occurrences too.

docling_core/experimental/idoctags.py
```diff
@@ -441,79 +1412,1068 @@ def serialize(
...
+                return
+            item = doc.add_code(
+                text=code_text,
+                code_language=lang_label,
+                parent=parent,
+                prov=(prov_list[0] if prov_list else None),
+            )
+            for p in prov_list[1:]:
+                item.prov.append(p)
+            return
```
vagenas: Using a single explicit `if ... elif ... else` would be more readable than the various `if ... return`s.

docling_core/experimental/idoctags.py
```diff
@@ -441,79 +1412,1068 @@ def serialize(
...
+        parts: list[str] = []
+        for node in el.childNodes:
+            if isinstance(node, Text):
+                if node.data.strip():
+                    parts.append(node.data)
+            elif isinstance(node, Element):
+                nm_child = node.tagName
+                if nm_child == IDocTagsToken.FACETS.value:
+                    facets_text = self._get_text(node).strip()
+                    if "=" in facets_text:
```
vagenas, line 2365: Since we have the parsed elements we shouldn't have to manually parse text again.

docling_core/experimental/idoctags.py
```diff
@@ -1,199 +1,933 @@
...
+    TABLE = "table"
     FORM = "form"
-    EMPTY_VALUE = "empty_value"  # used for empty value fields in fillable forms
+
+
+class IDocTagsVocabulary(BaseModel):
+    """IDocTagsVocabulary."""
+
+    # Allowed attributes per token (defined outside the Enum to satisfy mypy)
+    ALLOWED_ATTRIBUTES: ClassVar[dict[IDocTagsToken, set["IDocTagsAttributeKey"]]] = {
```
vagenas, line 450: Wondering if something more Pydantic-like (e.g. [pydantic-xml](https://github.com/dapper91/pydantic-xml)) could help better organize the various validation rules.

If we keep these low-level rules for now, at least I would make them internal e.g. `_ALLOWED_ATTRIBUTES`.

docling_core/experimental/idoctags.py
PeterStaar-IBM: we have a method,

```
def get_category(cls, token: IDocTagsToken) -> IDocTagsCategory
```

docling_core/experimental/idoctags.py
PeterStaar-IBM: I will update with the latest new table, but it gives a lot of good context for the code generators.

docling_core/experimental/idoctags.py
PeterStaar-IBM: The current `DocSerializer` is also a pydantic class, I would assume we want symmetry between `DocSerializer` and `DocDeserializer`.

docling_core/experimental/idoctags.py
PeterStaar-IBM: done

docling_core/experimental/idoctags.py
PeterStaar-IBM: this pattern exists because of an API design constraint. Let me show you what's happening:

  The add_* methods (like add_text, add_code, etc.) accept a single ProvenanceItem:

  def add_text(
      self,
      ...
      prov: Optional[ProvenanceItem] = None,  # Single item, not a list
      ...
  )

  But the actual item's prov field is a list:

  class BaseItem:
      ...
      prov: List[ProvenanceItem] = []

  Inside the add_* methods, when you pass a prov parameter, it gets appended to the list (docling_core/types/doc/document.py:3253-3254):

  if prov:
      text_item.prov.append(prov)

  So when idoctags.py has multiple provenance items in prov_list, it:
  1. Passes the first one to the constructor (which appends it internally)
  2. Manually appends the remaining items afterward

  The alternative of prov=prov_list or None wouldn't work because the API expects a single ProvenanceItem, not a List[ProvenanceItem].

docling_core/experimental/idoctags.py
PeterStaar-IBM: fixed (regex removed)

docling_core/experimental/idoctags.py
PeterStaar-IBM: this can be refactored later I would say

docling_core/experimental/idoctags.py
PeterStaar-IBM: fixed

docling_core/experimental/idoctags.py
PeterStaar-IBM: cleaned this up and fixed.

docling_core/experimental/idoctags.py
PeterStaar-IBM: fixed

examples/convert_to_idoctags.py
```diff
@@ -0,0 +1,696 @@
...
+    IDocTagsVocabulary,
+    IDocTagsDocSerializer,
+)
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+# In order to download **before** the datasets library, run
+# 
+# HF_HUB_DISABLE_XET=1 hf download --repo-type dataset "docling-project/canva-set-a"
```
cau-git: Let's remove this comment since the dataset is not public.

cau-git (approved):

PR #468, opened 2026-01-08
fix: lazy load transformers in HybridChunker

### LINKED ISSUE #459: Using docling[chunking-openai] raises import RuntimeError
I'm using docling-serve to process the document but wanted to run the chunking locally using OpenAI's tiktoken. For this, I added `docling[chunking-openai]` to my pyproject.toml
However, importing the HybridChunker still fails with:
```
RuntimeError: Extra required by module: 'chunking' by default (or 'chunking-openai' if specifically using OpenAI tokenization); to install, run: `pip install 'docling-core[chunking]'` or `pip install 'docling-core[chunking-openai]'`
```

`docling-core[chunking-openai]` does not install transformers
https://github.com/docling-project/docling-core/blob/5ab8b8c2cc2fbd32104a6cc56d89b9f6126849e1/pyproject.toml#L81-L93

But transformers is imported regardless in https://github.com/docling-project/docling-core/blob/5ab8b8c2cc2fbd32104a6cc56d89b9f6126849e1/docling_core/transforms/chunker/hybrid_chunker.py#L14-L18

ryyhan:
## Description
Resolves #459.

Implements lazy loading for `transformers` and `semchunk` in `HybridChunker`. This prevents `RuntimeError` when using `docling-core` with only the `chunking-openai` extra (which does not depend on `transformers`).

**Note:** This applies the same lazy-loading pattern accepted in `docling` PR #2826(https://github.com/docling-project/docling/pull/2826) (fix: transformers models lazy-loaded).

## Changes
- Moved `semchunk` import to `_split_using_plain_text` method.
- Replaced explicit `isinstance` check for `PreTrainedTokenizerBase` with a module path string check (`tokenizer.__module__`) to avoid importing `transformers`.

## Type of change
- [ ] Bug fix (non-breaking change which fixes an issue)

## Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] New and existing unit tests pass locally with my changes

### DIFF
--- a/docling_core/transforms/chunker/hybrid_chunker.py
+++ b/docling_core/transforms/chunker/hybrid_chunker.py
@@ -7,16 +7,6 @@ from typing import Any, Optional, Union, cast
 
 from pydantic import BaseModel, ConfigDict, Field, computed_field, model_validator
 
-try:
-    import semchunk
-except ImportError:
-    raise RuntimeError(
-        "Extra required by module: 'chunking' by default (or 'chunking-openai' if "
-        "specifically using OpenAI tokenization); to install, run: "
-        "`pip install 'docling-core[chunking]'` or "
-        "`pip install 'docling-core[chunking-openai]'`"
-    )
-
 from docling_core.transforms.chunker import (
     BaseChunk,
     BaseChunker,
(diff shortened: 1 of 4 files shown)


jspast:
@dolfim-ibm Sorry, I know you have already been mentioned in this PR, but after 4 months and no activity I assume it got lost somewhere. This PR makes it possible to use `HybridChunker` without necessarily importing `transformers`. Please take a look when possible. Thanks in advance.

ceberam:
Thanks @ryyhan and @jspast for the heads-up. Could you please resolve the existing conflict?

ceberam:
@ryyhan I wanted to push forward this PR, so I rebased to the latest `main` and pushed everything again. The PR introduced lazy loading for transformers in HybridChunker to avoid a RuntimeError when the library is used with only the chunking-openai extra. However, I have seen some inconsistencies among the modules `hybrid_chunker.py`, `huggingface.py`, and `openai.py`, so I created a new commit to ensure that all three modules are aligned to the pattern established by the docling backends (msword_backend.py, mspowerpoint_backend.py, etc.):
- Module-level `_XXX_AVAILABLE: bool = False` and `_XXX_IMPORT_ERROR: ImportError | None = None` flags.
- A try/except `ImportError` as e block that silently captures the error rather than re-raising it.
- A `_INSTALL_HINT` string with a human-friendly install instruction.
- The actual raise `ImportError(_INSTALL_HINT)` from `_XXX_IMPORT_ERROR` deferred to the point of use: inside `model_validator(mode="before")` for Pydantic model instantiation, and inside `from_pretrained()` for the class-method constructor. For `semchunk`, the guard sits inside `segment()` immediately before the call.

The net result: importing any of these modules is always safe regardless of which optional extras are installed. The ImportError with a clear install hint is only raised when user code actually tries to instantiate the class that requires the missing package.

ceberam (dismissed):

PeterStaar-IBM (dismissed):
lgtm!

dolfim-ibm (dismissed):

vagenas (dismissed):

ceberam (approved):

cau-git (approved):

PR #470, opened 2026-01-08
fix: disable numparse in tabulate to preserve trailing zeros

### LINKED ISSUE #350: Issue with loss of trailing zeros in formatted numbers during PDF parsing
Loss of trailing zeros are happening in Docling, i followed the issue and found it's happening due to Docling-core's `export_to_markdown()` method (see attached screenshot).

Image

Docling issue link: https://github.com/docling-project/docling/issues/1891

To fix the bug need to pass `disable_numparse=True` in `try` code block but currently it's in the `except`... i wanted to fix/suggest an update but hesitated if there is a case i am not aware that requires current approach.

Looking forward for maintainer's opinion... TIA.

ryyhan:
Resolves #350

Fixes the loss of trailing zeros in Markdown table export by disabling automatic number parsing in `tabulate`.

Verified with reproduction script.


docling_core/transforms/serializer/markdown.py
cau-git: Now the try and except blocks call `tabulate` with exactly the same parameters, which makes no sense any more. 
It looks like the original intention was to try using `tabulate` _with_ number parsing, and disable it as fallback when a `ValueError` occurs.

cau-git (commented):
@ryyhan Thanks for attempting a fix at #350. We must however ensure it has no unwanted side effects in other cases than the ones observed in that issue. The original logic was for sure created with some intention.

The branch may also need an rebase to `main`, I see some unrelated stuff like the `add_comment` in the diff, which is already present on `main`.

docling_core/transforms/serializer/markdown.py
ryyhan: Hi @cau-git,

Thank you for the review.

1. On Redundancy: You are correct; my changes made the try...except block redundant as both paths now use disable_numparse=True. My intention was to strictly enforce the preservation of trailing zeros (Issue #350), which tabulate strips by default.

2. On Side Effects: I have verified the side effects. Disabling number parsing preserves the exact text (e.g., "1.20") but changes the column alignment from right-aligned (numeric) to left-aligned (text).

Default: 1.2 (Right aligned, data loss)
Proposed: 1.20 (Left aligned, high fidelity)
I believe preserving the document's original text fidelity is more important than the heuristic alignment provided by tabulate. If you agree, I will proceed with this trade-off.

3. Next Steps: I will rebase the branch to main to remove the unrelated add_comment changes. I will also simplify the code to remove the now-redundant try...except block, leaving just the single tabulate call with disable_numparse=True.

Does this approach sound good to you?

cau-git: This sounds reasonable to me. Let's see where it reflects in the test ground truth after reproducing that one. To do so please also run `DOCLING_GEN_TEST_DATA=1 uv run pytest -s` then commit the changes.

ryyhan: Hi @cau-git,

Thanks again for the guidance.

I have:
1.  **Rebased** the branch on the latest `main` to clean up the history.
2.  **Simplified** the implementation in [markdown.py](cci:7://file:///Users/danishkhan/Downloads/codes/contributions/docling-core/docling_core/transforms/serializer/markdown.py:0:0-0:0) (removed the redundant `try...except`).
3.  **Updated the test ground truth** by running the requested command.

The tests are passing locally. I've pushed the changes.
Ready for another look!

lulmer:
Hello everyone, I am also facing this issue, was planing to PR and discovered this PR that solve the problem perfectly. 
Since I see it is open since january, is there anything that needs help so we can merge it ? 

Thank you in advance

cau-git:
@ryyhan @lulmer This was addressed in the meantime. Closing this PR.

PR #473, opened 2026-01-09
fix(deps): switch to tree-sitter-java-orchard and expand typer compatibility

### LINKED ISSUE #429: Replace tree-sitter-java with tree-sitter-java-orchard
In downstream we are having issues with tree-sitter-java package. It does not build from source, because the source distribution is missing files. @wetneb recommended to move to the fork `tree-sitter-java-orchard`, which is actively maintained and has additional fixes.

https://github.com/tree-sitter/tree-sitter-java/issues/219#issuecomment-3542400429

dolfim-ibm:
Resolves #429

### DIFF
--- a/docling_core/transforms/chunker/code_chunking/_utils.py
+++ b/docling_core/transforms/chunker/code_chunking/_utils.py
@@ -1,12 +1,8 @@
 """Utility functions and classes for code language detection and processing."""
 
+import sys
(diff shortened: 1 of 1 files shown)


vagenas (approved):
Let's revert this workaround as soon as we drop 3.9 support ;-)

PR #476, opened 2026-01-13
feat(IDocTags): add fine-grained content serialization filtering

vagenas:
Replacing https://github.com/docling-project/docling-core/pull/447 in the context of the updated IDocTags serializer.


test/data/doc/content_block_specific.gt.idt.xml
```diff
@@ -0,0 +1,164 @@
+<doctag version="1.0.0">
+  <list ordered="false">
+    <list_text>item of leading list</list_text>
+  </list>
+  <title>Title of the Document</title>
+  <text>Author 1
```
PeterStaar-IBM, line 6: Do we need a `<br>` here in theory?

PR #485, opened 2026-01-21
fix(serialization): escape pipe symbol in single cell md serialization

ceberam:
This PR fixes the markdown serialization of pipe characters (|) in a table single cell.
In markdown, pipe characters appearing in table cells need to be replaced by the HTML character code `&#124;`.

This PR is intended to support PR docling-project/docling#2904  , which addresses the issue docling-project/docling#2880 

Known limitation: in case of `RichTableCell` objects, the markdown serializer needs a prior refactoring to properly serialize nested tables (currently not supported).


dolfim-ibm (approved):
lgtm

PR #501, opened 2026-02-02
fix(serialization): add 'traverse_pictures' parameter to serializers

ceberam:
Add `traverse_pictures` parameter to enable serialization of text inside `PictureItem` objects.

Resolves https://github.com/docling-project/docling/issues/1094

### Problem
When chunking or serializing documents, TextItem objects that are children of PictureItem objects were not being included in the output. This was because the `_iterate_items` function has a `traverse_pictures` parameter to control this behavior, but:
1. The parameter was not exposed in `CommonParams` (similar to how `include_formatting` is exposed)
2. The `get_parts` method was not passing the parameter through to `_iterate_items`

This meant that text content embedded within pictures (e.g., OCR text, captions within picture boundaries) was silently excluded from serialization and chunking operations.

This may be a desired behavior to avoid serializing OCR text, but in other use cases, like information retrieval, we may want to capture all the text parsed under picture items.

### Solution
Added a new `traverse_pictures` parameter to `CommonParams` class that:
- Defaults to `False` to maintain backward compatibility
- Can be set to `True` to include text items that are children of PictureItems
- Follows the same pattern as existing parameters like `compact_tables` in MarkdownParams.

The parameter is now properly propagated through the serialization chain via the `get_parts` method.

### Changes
1. **`docling_core/transforms/serializer/common.py`**:
   - Added `traverse_pictures: bool` field to `CommonParams` class with default value `False`
   - Updated `get_parts` method to pass `params.traverse_pictures` to `_iterate_items`

2. **`test/test_serialization.py`**:
   - Added `test_md_traverse_pictures()` test to verify the parameter works correctly
   - Tests both default behavior (text inside pictures excluded) and enabled behavior (text inside pictures included)

### Usage Example
Users can now control this behavior by creating a custom serializer provider:

```python
from docling_core.transforms.chunker import PageChunker
from docling_core.transforms.chunker.hierarchical_chunker import (
    ChunkingSerializerProvider,
    ChunkingDocSerializer,
)
from docling_core.transforms.serializer.markdown import MarkdownParams

class TraversePicturesSerializerProvider(ChunkingSerializerProvider):
    def get_serializer(self, doc):
        params = MarkdownParams(
            traverse_pictures=True,  # Enable traversing into pictures
            image_mode=ImageRefMode.PLACEHOLDER,
            image_placeholder="",
            escape_underscores=False,
            escape_html=False,
        )
        return ChunkingDocSerializer(doc=doc, params=params)

# Use with PageChunker
page_chunker = PageChunker(
    serializer_provider=TraversePicturesSerializerProvider()
)
doc_chunks = list(page_chunker.chunk(doc))
```

### Testing
- ✅ All existing tests pass (39 serialization tests, 3 chunker tests)
- ✅ New test `test_md_traverse_pictures` validates both enabled and disabled states
- ✅ Verified with real-world document that previously missing text is now included when parameter is enabled

### Backward Compatibility
This change is fully backward compatible:
- Default value is `False`, preserving existing behavior
- Only affects output when explicitly set to `True`
- No changes to existing APIs or method signatures

### DIFF
--- a/test/test_serialization.py
+++ b/test/test_serialization.py
@@ -13,6 +13,7 @@ from docling_core.transforms.serializer.html import (
 from docling_core.transforms.serializer.markdown import (
     MarkdownDocSerializer,
     MarkdownParams,
+    MarkdownTableSerializer,
     OrigListItemMarkerMode,
 )
 from docling_core.transforms.serializer.webvtt import WebVTTDocSerializer
@@ -21,8 +22,11 @@ from docling_core.types.doc.base import ImageRefMode
 from docling_core.types.doc.document import (
     DescriptionAnnotation,
     DoclingDocument,
+    PictureItem,
+    RefItem,
     TableCell,
     TableData,
+    TextItem,
 )
 from docling_core.types.doc.labels import DocItemLabel
 
@@ -352,7 +356,6 @@ def test_md_pipe_in_table():
 
 def test_md_compact_table():
     """Test compact table format removes padding and uses minimal separators."""
-    from docling_core.transforms.serializer.markdown import MarkdownTableSerializer
 
     # Test the _compact_table method directly
     padded_table = """| item   | qty   | description           |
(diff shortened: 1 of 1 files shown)


CivisQi4M:
This fix doesn't work when using HierarchialChunker (and HybridChunker as a consequence), since the iteration through document's items is done via `dl_doc.iterate_items(with_groups=True):` (see line 155) without any reference to serializer params

PR #513, opened 2026-02-12
chore: add support for `pandas==3.0.0`

wanadzhar913:
Hi there, this PR adds support for the newly released `pandas==3.0.0`. However: **it only supports python 3.11** (as per [here](https://pandas.pydata.org/docs/whatsnew/v3.0.0.html#increased-minimum-version-for-python)) which results in discarding support for python 3.10.

A full list of backward incompatible API changes can be found [here](https://pandas.pydata.org/docs/whatsnew/v3.0.0.html#backwards-incompatible-api-changes).

Would love to know what the core maintainers think about whether to support or not.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,8 +43,7 @@ dependencies = [
     'pydantic (>=2.6.0,<3.0.0,!=2.10.0,!=2.10.1,!=2.10.2)',
     'jsonref (>=1.1.0,<2.0.0)',
     'tabulate (>=0.9.0,<0.10.0)',
-    'pandas (>=2.1.4,<3.0.0); python_version < "3.11"',
-    'pandas (>=2.1.4,<4.0.0); python_version >= "3.11"',
+    'pandas (>=2.1.4,<4.0.0)',
     'pillow (>=10.0.0,<13.0.0)',
     'pyyaml (>=5.1,<7.0.0)',
     'typing-extensions (>=4.12.2,<5.0.0)',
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -43,7 +43,8 @@ dependencies = [
     'pydantic (>=2.6.0,<3.0.0,!=2.10.0,!=2.10.1,!=2.10.2)',
     'jsonref (>=1.1.0,<2.0.0)',
     'tabulate (>=0.9.0,<0.10.0)',
-    'pandas (>=2.1.4,<3.0.0)',
+    'pandas (>=2.1.4,<3.0.0); python_version < "3.11"',
+    'pandas (>=2.1.4,<4.0.0); python_version >= "3.11"',
```
dolfim-ibm: I think we should simply broaden the range.
The correct pinning based on python versions is usually needed for the lock file and should be done in the `[dependency-groups]` section.

```suggestion
    'pandas (>=2.1.4,<4.0.0)',
```

pyproject.toml
```diff
@@ -93,7 +94,8 @@ examples = [
     "datasets>=4.0.0",
     "matplotlib>=3.7.0",
     "openpyxl>=3.1.5",
-    "pandas>=2.3.3",
+    'pandas (>=2.3.3,<3.0.0); python_version < "3.11"',
+    'pandas (>=2.3.3,<4.0.0); python_version >= "3.11"',
```
dolfim-ibm: remove it completely from here

```suggestion
```

pyproject.toml
dolfim-ibm: Done in [8daaa2e](https://github.com/docling-project/docling-core/pull/513/commits/8daaa2ee7ef6f17fea06946fca4d9bf8f2a5e422)

pyproject.toml
dolfim-ibm: Done in [8daaa2e](https://github.com/docling-project/docling-core/pull/513/commits/8daaa2ee7ef6f17fea06946fca4d9bf8f2a5e422)

PR #514, opened 2026-02-13
fix: updated caption references in add_node_items

lukesan48:
### Description
Fixes a bug where a `TableItem` caption reference was not correctly updated during document-to-document node transfers. The index was carried over exactly instead of updating.

When using `add_node_items` to copy elements (like a Table) from a source document to a destination document, the caption text was being copied, but the parent element's internal `captions` list still pointed to the index in the original documen. This can result in broken references or `IndexError` in the new document.

### Related Issues
Attempt to fix docling-project/docling#2298

### Changes
- Modified `_append_item_copies` to recursively resolve, copy, and re-index captions for FloatingItems.
- Used `getattr`/`setattr` pattern to ensure compatibility with `NodeItem` base class and satisfy MyPy type checking.

### Verification
Verified with a reproduction script extracting a table from a large document:
- **Before:** New Table Caption Ref pointed to out-of-bounds index (original doc index).
- **After:** New Table Caption Ref correctly points to `#/texts/0` (new doc index).


docling_core/types/doc/document.py
```diff
@@ -4562,6 +4562,19 @@ def _append_item_copies(
         for item in node_items:
             item_copy = item.model_copy(deep=True)
 
+            # handling captions for floating items (Tables/Pictures)
+            captions = getattr(item, "captions", None)
+            if captions:
```
dolfim-ibm: instead of using `getattr` and `hasattr`, we prefer using the `isinstance(item, ...)`

lukesan48: Thanks for the review! I've updated the logic to use `isinstance(item, FloatingItem)` as requested.

I also added a new test case `(test_add_node_items_updates_captions)` to the test suite. 

Ready for another look when you have a moment!

ceberam (commented):
Thanks @lukesan48 for catching this bug!
The fix solves the issue 🏆 

However, I realize that `FloatingItem` has also other fields that would lead to the same issue: `references`, `footnotes`, and maybe `comments` from its parent too.

It would be helpful if you could address those cases in this PR. Otherwise, we can merge it as it is and create a new one.

docling_core/types/doc/document.py
```diff
@@ -4562,6 +4562,30 @@ def _append_item_copies(
         for item in node_items:
             item_copy = item.model_copy(deep=True)
 
+            # handle DocItem pointers (comments)
+            if isinstance(item, DocItem):
+                if item.comments:
+                    if isinstance(item_copy, DocItem):
+                        item_copy.comments = self._copy_and_reindex_refs(item.comments, doc=doc, parent_ref=parent_ref)
```
ceberam, line 4569: For readability, can we keep it more compact? (this one and the other similar patterns) 
 
```suggestion
            if isinstance(item, DocItem) and item.comments and isinstance(item_copy, DocItem):
                item_copy.comments = self._copy_and_reindex_refs(item.comments, doc=doc, parent_ref=parent_ref)
```

docling_core/types/doc/document.py
```diff
@@ -4578,6 +4602,31 @@ def _append_item_copies(
...
         return new_refs
 
+    def _copy_and_reindex_refs(self, ref_list: list[Any], doc: "DoclingDocument", parent_ref: RefItem) -> list[Any]:
+        """Helper to copy referenced items and return their new indices
+
+        :param ref_list: list[Any]: The list of references (e.g., captions, footnotes, comments) to be copied
+        :param doc: "DoclingDocument": The document from which the NodeItems are taken
+        :param parent_ref: RefItem: The reference of the parent item in the current document where copies will be appended to
+
+        :returns: list[Any]: A new list of references pointing to the newly appended items in the current document
```
ceberam, line 4612: Even though we have not been consistent in the past, we want to stick to the google docstring conventions at least on new code, as we specify it on [pyproject.toml](https://github.com/docling-project/docling-core/blob/main/pyproject.toml#L183)

docling_core/types/doc/document.py
```diff
@@ -4578,6 +4602,31 @@ def _append_item_copies(
 
         return new_refs
 
+    def _copy_and_reindex_refs(self, ref_list: list[Any], doc: "DoclingDocument", parent_ref: RefItem) -> list[Any]:
```
ceberam, line 4605: Fine grain comment: can we be more precise with the type hints? Would `list[Any]` rather be `list[NodeItem]` ?

ceberam (changes requested):
Thanks @lukesan48 for addressing those extra cases.
I just had some minor style comments that would be good to tackle too.

docling_core/types/doc/document.py
lukesan48: Because `item.comments` expects a `list[FineRef]` while other fields use `list[RefItem]`, MyPy throws a list invariance error if I type-hint the helper method strictly as `list[RefItem]` or `list[FineRef]`. Would you prefer I keep the type hint as `list[Any]` to handle both cases or split this into two separate helper functions?

lukesan48: @ceberam Just a gentle ping on this when you have a moment.

If list[Any] is a bit too loose for the project's typing standards, I can implement a generic TypeVar (e.g., T_Ref = TypeVar("T_Ref", bound="RefItem") alongside Sequence[T_Ref]). This approach should satisfy MyPy's strict list invariance rules while keeping the exact typing intact for both FineRef and RefItem outputs.

Let me know if you'd like me to push that update, or if you're comfortable moving forward with it as-is!

PR #519, opened 2026-02-17
feat: introduce field data model incl. Doclang serialization

vagenas:
Open points:
- [ ] DoclingDocument version bump
- Doclang deserialization
- Serialization to Markdown / HTML /...

### DIFF
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -6033,16 +6033,6 @@ class DoclingDocument(BaseModel):
 
         return ser_res.text
 
-    def export_to_doclang(self) -> str:
-        """Export to Doclang."""
-        from docling_core.experimental.doclang import DoclangDocSerializer, DoclangParams
-
-        serializer = DoclangDocSerializer(
-            doc=self,
-            params=DoclangParams(),
-        )
-        return serializer.serialize().text
-
     @staticmethod
     def load_from_doctags(  # noqa: C901
         doctag_document: DocTagsDocument, document_name: str = "Document"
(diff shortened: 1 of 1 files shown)


docling_core/types/doc/document.py
```diff
@@ -6163,6 +6938,8 @@ class _DocIndex(BaseModel):
         tables: list[TableItem] = []
         key_value_items: list[KeyValueItem] = []
         form_items: list[FormItem] = []
+        field_regions: list[FieldRegionItem] = []
```
PeterStaar-IBM, line 6962: should we make those private? (just to give us a bit of breathing room later on)

docling_core/types/doc/document.py
```diff
@@ -6163,6 +6938,8 @@ class _DocIndex(BaseModel):
...
+        field_items: list[FieldItem] = []
```
PeterStaar-IBM, line 6963: should we make those private? (just to give us a bit of breathing room later on)

docling_core/types/doc/document.py
vagenas: Pydantic private attributes in principle imply exclusion from JSON dump/load & even if "hacked" into JSON, it would defeat the purpose of "private" (although actually helpful in JSON for debugging/maintenance).

(An alternative could be subclassing DoclingDocument in the experimental package & adding all delta there — but also that should be short-lived to prevent maintenance overhead).

docling_core/types/doc/document.py
vagenas: Answered above.

PR #525, opened 2026-02-23
fix: prevent hang in export_to_markdown() on nested RichTableCells

### LINKED ISSUE #524: export_to_markdown() hangs on documents with nested RichTableCells
## Bug

`export_to_markdown()` hangs indefinitely on documents where a `RichTableCell` references a nested `TableItem`.

## Root cause

`MarkdownTableSerializer.serialize()` calls `doc_serializer.serialize()` for each `RichTableCell`'s content. When that content is itself a table, the call re-enters `MarkdownTableSerializer`, which passes already-formatted markdown strings to `tabulate()`. Those strings grow exponentially with each level of nesting — eventually exhausting memory or hanging indefinitely.

## Affected versions

Introduced in v2.46.0 (commit 1d04154) when `RichTableCell` markdown serialization was first added.

## Reproduction

Any document with a `RichTableCell` whose `ref` resolves to a `TableItem` — for example, a Wikipedia page with a taxobox or classification table produced by the docling HTML backend.

ivan-liminary:
Fixes #524

## Summary

When a `RichTableCell` references a nested `TableItem`, `MarkdownTableSerializer` recursively called `doc_serializer.serialize()`, causing `tabulate()` to receive already-formatted markdown tables as cell content. Those strings were `&#124;`-encoded and re-processed at each level, causing exponential string growth and an indefinite hang (observed on Wikipedia pages with taxobox/classification tables).

## Fix

Add two module-level helpers to `markdown.py`:

- **`_cell_content_has_table(item, doc)`** — walks the item's subtree and returns `True` if any node is a `TableItem`. Used before calling `serialize()` to detect the problematic case.
- **`_mark_subtree_visited(item, doc, visited)`** — adds all nodes in the subtree to the shared `visited` set. Needed because skipping `serialize()` means the normal bookkeeping that prevents double-emission never runs.

In `MarkdownTableSerializer.serialize()`, when a cell's content contains a nested table: mark the subtree visited and fall back to `col.text` (the precomputed plain-text value). For cells without nested tables, `doc_serializer.serialize()` is called as before, so inline Markdown (italic, bold, hyperlinks, etc.) is preserved.

## Tests

- `test_md_nested_rich_table_no_hang` — builds a depth-5 nested structure and asserts `export_to_markdown()` completes within 5 seconds.
- `test_md_rich_table` — asserts the nested table does not appear as a duplicate top-level item.

Bug introduced in v2.46.0 (commit 1d04154) when `RichTableCell` markdown serialization was first added.

### DIFF
--- a/docling_core/transforms/serializer/markdown.py
+++ b/docling_core/transforms/serializer/markdown.py
@@ -69,47 +69,40 @@ from docling_core.types.doc.document import (
 )
 
 
-def _cell_content_has_table(item: Any, doc: "DoclingDocument") -> bool:
-    """Return True if *item* is, or has a descendant that is, a TableItem.
-
-    Used by MarkdownTableSerializer to decide whether a RichTableCell's
-    content should be serialized recursively or flattened into plain text.
-    Markdown table cells cannot contain nested block-level tables, so when
-    the rich content includes one we collect text via
-    ``_collect_subtree_text`` rather than producing pipe-heavy output that
-    breaks the outer table.
-    """
+def _cell_content_has_table(item: NodeItem, doc: DoclingDocument) -> bool:
+    """Return True if *item* is, or has a descendant that is, a TableItem."""
     if isinstance(item, TableItem):
         return True
-    for child_ref in getattr(item, "children", []):
-        if _cell_content_has_table(child_ref.resolve(doc=doc), doc):
-            return True
+    elif isinstance(item, NodeItem):
+        for child_ref in item.children:
+            if _cell_content_has_table(child_ref.resolve(doc=doc), doc):
+                return True
     return False
 
 
-def _mark_subtree_visited(item: Any, doc: "DoclingDocument", visited: set[str]) -> None:
+def _mark_subtree_visited(
+    item: NodeItem,
+    doc: DoclingDocument,
+    visited: set[str],
+) -> None:
     """Recursively add *item* and all its descendants to *visited*.
 
-    When a RichTableCell's content is skipped (because it contains a nested
-    table), the content is never passed through the normal serialize() path
-    that would mark items visited.  Calling this instead keeps the visited
-    set consistent so the document serializer does not emit those items a
-    second time at the top level.
+    When a nested table inside a RichTableCell is flattened, its items are
+    never passed through the normal serialize() path that would mark them
+    visited.  Calling this keeps the visited set consistent so the document
+    serializer does not emit those items again at the top level.
     """
-    if (ref := getattr(item, "self_ref", None)) is not None:
-        visited.add(ref)
-    for child_ref in getattr(item, "children", []):
-        _mark_subtree_visited(child_ref.resolve(doc=doc), doc, visited)
+    if isinstance(item, NodeItem):
+        visited.add(item.self_ref)
+        for child_ref in item.children:
+            _mark_subtree_visited(child_ref.resolve(doc=doc), doc, visited)
 
 
-def _collect_subtree_text(item: Any, doc: "DoclingDocument") -> str:
+def _collect_subtree_text(item: NodeItem, doc: DoclingDocument) -> str:
     """Collect all text from *item*'s subtree, flattening nested tables.
 
-    When a RichTableCell contains a nested TableItem the markdown serializer
-    cannot render it as a table inside the outer cell (no markdown spec
-    supports nested tables).  This helper walks the subtree and returns a
-    space-joined string of every piece of text it finds so that the content
-    is preserved in a flat, readable form.
+    Returns a space-joined string of every piece of text found so that the
+    content of a nested table is preserved in a flat, readable form.
 
     For TableItems the text is pulled from ``data.grid`` cells directly;
     children are *not* recursed into because they duplicate the grid content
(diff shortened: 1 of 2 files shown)


test/data/doc/rich_table.gt.md
```diff
@@ -1,8 +1,8 @@
...
 
-| cell 0,0                                               | cell 0,1                                                                                                                                                                         |
-|--------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| cell 1,0                                               | *text in italic*                                                                                                                                                                 |
-| - list item 1 - list item 2                            | cell 2,1                                                                                                                                                                         |
-| cell 3,0                                               | | inner cell 0,0   | inner cell 0,1   | inner cell 0,2   | |------------------|------------------|------------------| | inner cell 1,0   | inner cell 1,1   | inner cell 1,2   | |
```
ceberam, line 7: @ivan-liminary without entering yet into the implementation details, it seems that the inner table does not get serialized anymore.

ivan-liminary: Yes, I just pushed a fix that preserves the inner table content as flattened text. Nested tables aren't supported by any markdown spec, so the cell content can't be rendered as a proper table inside the outer table -- but the text from all inner cells is now preserved instead of being dropped. The previous behavior produced invalid pipe-heavy markdown that also triggered exponential string growth (the OOM hang this PR fixes).

ceberam: Thanks @ivan-liminary for the explanation. We've seen that some people use HTML for nested tables. What do you think about it? I would favor this flattened solution by the way.

ceberam (changes requested):
@ivan-liminary could you please fix the type check (Mypy) error?

test/data/doc/rich_table.gt.md
ivan-liminary: Agreed -- flattened text is the safer default since it produces valid markdown everywhere. Embedding HTML <table> inside a markdown cell works on GitHub but isn't portable across all renderers. If there's demand for it, that could be added as an opt-in option

docling_core/transforms/serializer/markdown.py
```diff
@@ -69,6 +69,74 @@
 )
 
 
+def _cell_content_has_table(item: Any, doc: "DoclingDocument") -> bool:
```
ceberam: You don't need to use string annotations (`"DoclingDocument"`) in `_cell_content_has_table` and `_mark_subtree_visited` signatures. The class is imported anyway.

docling_core/transforms/serializer/markdown.py
```diff
@@ -69,6 +69,74 @@
...
+    the rich content includes one we collect text via
+    ``_collect_subtree_text`` rather than producing pipe-heavy output that
+    breaks the outer table.
+    """
+    if isinstance(item, TableItem):
+        return True
+    for child_ref in getattr(item, "children", []):
+        if _cell_content_has_table(child_ref.resolve(doc=doc), doc):
+            return True
+    return False
```
ceberam, line 80: We prefer to do type checks and access pydantic model fields with the dot notation (the recommended method by pydantic) rather than using `getattr`.

```suggestion
    if isinstance(item, TableItem):
        return True
    elif isinstance(item, NodeItem):
        for child_ref in item.children:
            if _cell_content_has_table(child_ref.resolve(doc=doc), doc):
                return True
    return False
```

docling_core/transforms/serializer/markdown.py
```diff
@@ -69,6 +69,74 @@
...
+    When a RichTableCell's content is skipped (because it contains a nested
+    table), the content is never passed through the normal serialize() path
+    that would mark items visited.  Calling this instead keeps the visited
+    set consistent so the document serializer does not emit those items a
+    second time at the top level.
+    """
+    if (ref := getattr(item, "self_ref", None)) is not None:
+        visited.add(ref)
+    for child_ref in getattr(item, "children", []):
+        _mark_subtree_visited(child_ref.resolve(doc=doc), doc, visited)
```
ceberam: Same comment as above

docling_core/transforms/serializer/markdown.py
```diff
@@ -69,6 +69,74 @@
...
+            for cell in row:
+                if cell.text:
+                    parts.append(cell.text)
+        return " ".join(parts)
+
+    if isinstance(item, TextItem) and item.text:
+        parts.append(item.text)
+
+    for child_ref in getattr(item, "children", []):
+        child = child_ref.resolve(doc=doc)
```
ceberam: See comment above

docling_core/transforms/serializer/markdown.py
```diff
@@ -419,19 +487,30 @@ def serialize(
...
+                for col in row:
+                    if isinstance(col, RichTableCell):
+                        ref_item = col.ref.resolve(doc=doc)
+                        if _cell_content_has_table(ref_item, doc):
+                            # Markdown table cells cannot contain nested block-level
+                            # tables.  Collect all text from the subtree in flat form
+                            # so the content is preserved without pipe-heavy output
+                            # that would break the outer table.  Mark the skipped
+                            # subtree as visited so the document serializer does not
+                            # emit those items again at top level.
```
ceberam: Please, remove this comment for brevity. The intent is already thoroughly described in the helper functions.

PR #527, opened 2026-02-25
feat: table aware chunking

odelliab:
**Summary**
This PR adds table header duplication functionality to the HybridChunker and introduces a new LineBasedTokenChunker for improved table chunking. When tables are split across multiple chunks, the table headers are now automatically repeated in each chunk to maintain context.

**Changes Made**
**Core Features**

1. Table Header Duplication: Added duplicate_table_header parameter to HybridChunker (default: True)- 
- When enabled, table headers are repeated in each chunk when a table is split
- Uses the new LineBasedTokenTextChunker for table-specific chunking logic
2. - New LineBasedTokenChunker: Created a specialized chunker for line-based text splitting
- Supports prefix text that is prepended to each chunk (e.g., table headers)
- Handles token-aware splitting with word boundary preservation
- Includes comprehensive validation and error handling

**Implementation Details**

- Modified HybridChunker._split_using_plain_text() to accept a doc_serializer parameter (as headers identification depends on the serializer)
- Added new segment() method to handle table vs. non-table content differently
- Integrated get_header_and_body_lines() from table serializers to extract headers

**New Files**
docling_core/transforms/chunker/line_chunker.py: New line-based chunker implementation

test/test_line_chunker.py: Comprehensive test suite (308 lines) covering:

- Prefix handling
- Long line splitting
- Token limit enforcement
- Word boundary preservation
- Document chunking
- Edge cases (empty documents, long content, etc.)

**Testing**
✅ Added comprehensive unit tests for LineBasedTokenTextChunker (15+ test cases)
✅ Added integration tests for table header duplication in test_hybrid_chunker.py
✅ Verified table headers are repeated when tables span multiple chunks
✅ Confirmed all body lines from tables appear in the chunked output
✅ Tested with real document data (2408.09869v3 with 996 texts, 5 tables, 13 pictures)
✅ Generated 84 chunks with 27 containing tables

**Additional Notes**
This enhancement significantly improves the usability of chunked table data by ensuring each chunk maintains the necessary context (table headers) for standalone interpretation. This is particularly valuable for:

- RAG (Retrieval-Augmented Generation) applications
- Embedding generation where each chunk needs to be self-contained
- Search and retrieval systems that work with individual chunks

Currently, for get_header_and_body_lines(), a naive default implementation and an implementation for markdown table serializer exist. We should add implementation for other serializers (e.g. html).

The implementation is backward compatible - the duplicate_table_header parameter defaults to True but can be disabled if the old behavior is desired.

### DIFF
--- a/docling_core/transforms/chunker/hybrid_chunker.py
+++ b/docling_core/transforms/chunker/hybrid_chunker.py
@@ -7,13 +7,13 @@ from typing import Any, Optional, Union
 
 from pydantic import BaseModel, ConfigDict, Field, computed_field, model_validator
 
-from docling_core.transforms.chunker.base import _get_default_tokenizer
 from docling_core.transforms.chunker.hierarchical_chunker import (
     ChunkingDocSerializer,
     ChunkingSerializerProvider,
 )
 from docling_core.transforms.chunker.line_chunker import LineBasedTokenChunker
 from docling_core.transforms.chunker.tokenizer.base import BaseTokenizer
+from docling_core.transforms.chunker.tokenizer.huggingface import get_default_tokenizer
 from docling_core.transforms.serializer.base import BaseDocSerializer
 from docling_core.types.doc.document import SectionHeaderItem, TableItem, TitleItem
 
@@ -56,7 +56,7 @@ class HybridChunker(BaseChunker):
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
-    tokenizer: BaseTokenizer = Field(default_factory=_get_default_tokenizer)
+    tokenizer: BaseTokenizer = Field(default_factory=get_default_tokenizer)
     repeat_table_header: bool = True
     merge_peers: bool = True
 
@@ -90,7 +90,7 @@ class HybridChunker(BaseChunker):
                         max_tokens=max_tokens,
                     )
                 elif tokenizer is None or isinstance(tokenizer, PreTrainedTokenizerBase):
-                    kwargs = {"tokenizer": tokenizer or _get_default_tokenizer().tokenizer}
+                    kwargs = {"tokenizer": tokenizer or get_default_tokenizer().tokenizer}
                     if max_tokens is not None:
                         kwargs["max_tokens"] = max_tokens
                     data["tokenizer"] = HuggingFaceTokenizer(**kwargs)
--- a/docling_core/transforms/chunker/line_chunker.py
+++ b/docling_core/transforms/chunker/line_chunker.py
@@ -1,16 +1,16 @@
 import warnings
 from collections.abc import Iterator
 from functools import cached_property
-from typing import Annotated, Any, Optional
+from typing import Annotated, Any
 
 from pydantic import ConfigDict, Field, computed_field
 
 from docling_core.transforms.chunker import BaseChunk, BaseChunker, DocChunk, DocMeta
-from docling_core.transforms.chunker.base import _get_default_tokenizer
 from docling_core.transforms.chunker.hierarchical_chunker import (
     ChunkingSerializerProvider,
 )
 from docling_core.transforms.chunker.tokenizer.base import BaseTokenizer
+from docling_core.transforms.chunker.tokenizer.huggingface import get_default_tokenizer
 from docling_core.transforms.serializer.base import (
     BaseSerializerProvider,
 )
@@ -31,7 +31,7 @@ class LineBasedTokenChunker(BaseChunker):
     tokenizer: Annotated[
         BaseTokenizer,
         Field(
-            default_factory=_get_default_tokenizer,
+            default_factory=get_default_tokenizer,
             description="The tokenizer to use; either instantiated object or name or path of respective pretrained model",
         ),
     ]
--- a/docling_core/transforms/chunker/base.py
+++ b/docling_core/transforms/chunker/base.py
@@ -13,14 +13,6 @@ from docling_core.types.doc import DoclingDocument as DLDocument
 DFLT_DELIM = "\n"
 
 
-def _get_default_tokenizer():
-    from docling_core.transforms.chunker.tokenizer.huggingface import (
-        HuggingFaceTokenizer,
-    )
-
-    return HuggingFaceTokenizer.from_pretrained(model_name="sentence-transformers/all-MiniLM-L6-v2")
-
-
 class BaseMeta(BaseModel):
     """Chunk metadata base class."""
 
--- a/docling_core/transforms/chunker/tokenizer/huggingface.py
+++ b/docling_core/transforms/chunker/tokenizer/huggingface.py
@@ -68,3 +68,9 @@ class HuggingFaceTokenizer(BaseTokenizer):
     def get_tokenizer(self):
         """Get underlying tokenizer object."""
         return self.tokenizer
+
+
+def get_default_tokenizer():
+    """Get default tokenizer instance."""
+
+    return HuggingFaceTokenizer.from_pretrained(model_name="sentence-transformers/all-MiniLM-L6-v2")


docling_core/transforms/chunker/hybrid_chunker.py
```diff
@@ -61,6 +62,7 @@ class HybridChunker(BaseChunker):
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
     tokenizer: BaseTokenizer = Field(default_factory=_get_default_tokenizer)
+    duplicate_table_header: bool = True
```
ceberam: Just wondering if the name could be more self explanatory. It may be misunderstood. What about `repeat_table_headers` (to convey the message that headers appear again across chunks).

Could you also add a description in the docstring? (by the way, see another comment below regarding our current docstrings in chunkers and serializers).

docling_core/transforms/chunker/hybrid_chunker.py
```diff
@@ -233,12 +236,34 @@ def _split_using_plain_text(
...
+        if (
+            self.duplicate_table_header
+            and isinstance(doc_serializer, ChunkingDocSerializer)
+            and len(doc_chunk.meta.doc_items) == 1
+            and isinstance(doc_chunk.meta.doc_items[0], TableItem)
+        ):
+            header_lines, body_lines = doc_serializer.table_serializer.get_header_and_body_lines(
+                table_text=doc_chunk.text
+            )
+            from docling_core.transforms.chunker.line_chunker import LineBasedTokenChunker
```
ceberam: This import inside a method is unconventional. I guess you were trying to avoid a circular import between `line_chunker.py` and `hybrid_chunker.py`.
I would then propose to move `_get_default_tokenizer()` to another module, like `base.py` and make the function public (for consistency).

docling_core/transforms/chunker/line_chunker.py
```diff
@@ -0,0 +1,193 @@
...
+
+class LineBasedTokenChunker(BaseChunker):
+    r"""Chunker doing tokenization-aware chunking of document text. Chunk contains full lines.
+
+    Args:
+        tokenizer: The tokenizer to use; either instantiated object or name or path of
+            respective pretrained model
+        max_tokens: The maximum number of tokens per chunk. If not set, limit is
+            resolved from the tokenizer
+        prefix: a text that should appear at the beginning of each chunks, default is an empty string
```
ceberam: Using `Args:` for class attributes is not correct according to Google style for python docstrings (it's for arguments in functions). I know that we have this pattern in other chunkers and in serializers, but we would prefer not to propagate mistakes 🙂
For pydantic models we are adopting a pattern based on the `typing.Annotated`. So we could do something like this:


```suggestion
    """Tokenization-aware chunker that preserves line boundaries.
    
    This chunker serializes the document content into text and attempts to keep lines
    intact within chunks. It only splits a line if it exceeds the maximum token limit on
    its own. This is particularly useful for structured content like tables, code, or logs
    where line boundaries are semantically important.
    """
```

and then add the documentation of each pydantic model field through annotations. See an example below.

docling_core/transforms/chunker/line_chunker.py
```diff
@@ -0,0 +1,193 @@
...
+    """
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+    tokenizer: BaseTokenizer = Field(default_factory=_get_default_tokenizer)
+    prefix: str = ""
```
ceberam: Please, provide documentation of the fields using the `Annotated` pattern.

```suggestion
    prefix: Annotated[
        str,
        Field(
            description="Text that appears at the beginning of each chunk. Useful for adding context like table headers"
        ),
    ] = ""
```

docling_core/transforms/chunker/line_chunker.py
```diff
@@ -0,0 +1,193 @@
...
+    prefix_len: int = Field(default=0, init=False)
```
ceberam: We prefer the more pythonic and self-contained use of decorators `computed_field` and `cached_property`, like in `docling_core/transforms/serializer/common.py`

```suggestion
    @computed_field  # type: ignore[misc]
    @cached_property
    _prefix_len(self) -> int:
       // add computation here
       pass
```

test/test_line_chunker.py
```diff
@@ -0,0 +1,308 @@
...
+    """Test text chunking with a prefix."""
+    prefix = "Context: "
+    chunker = LineBasedTokenChunker(
+        tokenizer=HuggingFaceTokenizer(
+            tokenizer=INNER_TOKENIZER,
+            max_tokens=MAX_TOKENS,
+        ),
+        prefix=prefix,
+    )
+
```
ceberam: Small observation here and in other lines of this test file: we try to follow the PEP8 recommendations and avoid trailing whitespaces.
We should at some point add some minimal style checks on tests.

```suggestion

```

docling_core/transforms/chunker/line_chunker.py
```diff
@@ -0,0 +1,193 @@
...
+        if token_limit <= 0 or not text:
+            return "", text
+
+        # if the whole text already fits, return as is.
+        if self.tokenizer.count_tokens(text) <= token_limit:
+            return text, ""
+
+        # Binary search over character indices [0, len(text)]
+        lo, hi = 0, len(text)
+        best_idx: Optional[int] = None
```
ceberam: Very minor observation: we try to use the union operator `|` in the new code, since it's the preferred style in python 3.10+


```suggestion
        best_idx: int | None = None
```

docling_core/transforms/chunker/line_chunker.py
```diff
@@ -0,0 +1,193 @@
...
+            # Even the first character exceeds the limit (e.g., tokenizer behavior).
+            # Return nothing in head, everything in tail.
+            return "", text
+
+        # Optionally adjust to a previous whitespace boundary without violating the limit
+        if prefer_word_boundary:
+            # Search backwards from best_idx to find whitespace; keep within token limit.
+
+            last_space_index = text[:best_idx].rfind(" ")
+            if last_space_index > 0:
```
ceberam: Shouldn't it be `>=`?

```suggestion
            if last_space_index >= 0:
```

docling_core/transforms/chunker/line_chunker.py
```diff
@@ -0,0 +1,193 @@
...
+                # Add the taken part
+                current += "\n" + take
+                current_len += self.tokenizer.count_tokens(take)
+
+                # flush the current chunk (full)
+                chunks.append(current)
+                current = self.prefix
+                current_len = self.prefix_len
+
+            # end while for this line
```
ceberam, line 150: I would remove comments that are not much informative

```suggestion
```

docling_core/transforms/serializer/base.py
```diff
@@ -78,6 +78,26 @@ def serialize(
...
+    ) -> tuple[list[str], list[str]]:
+        """Get header lines and body lines from the table.
+
+        Returns:
+            A tuple of (header_lines, body_lines) where header_lines is a list
+            of strings representing table headers and body_lines is a list of
+            strings representing table body rows.
+
+        Default implementation returns empty header lines and all content in body lines.
+        """
```
ceberam, line 104: I would make the docstrings more complete (and Google-styled):

```suggestion
        """Get header lines and body lines from the table.

        Splits a serialized table into header and body sections. This is useful
        for repeating headers when a table spans multiple chunks.

        Args:
            table_text: The serialized table text to split.

        Returns:
            A tuple of (header_lines, body_lines) where header_lines is a list
            of strings representing table headers and body_lines is a list of
            strings representing table body rows.

        Note:
            Default implementation returns empty header lines and all content 
            in body lines. Subclasses should override to provide format-specific
            splitting logic.
        """
```

Open question: there is here an implicit logical assumption that this function should work on top of the implementation of the serialization (`serialize`), which is specific to the class implementing `BaseTableSerializer`. The table serializer would serialize a table to text, and then `get_header_and_body_lines` would split the header and the body lines from that text. I'm just wondering if it would make sense to replace the argument type from `str` to `TableItem` and call the `serialize` function inside that method, to focus on the scope of the table serializers. Not sure about it.

docling_core/transforms/chunker/hybrid_chunker.py
odelliab: changed the name and added a description

docling_core/transforms/chunker/hybrid_chunker.py
odelliab: done

docling_core/transforms/chunker/line_chunker.py
odelliab: fixed

docling_core/transforms/serializer/base.py
odelliab: I was wondering about this myself, but if I remember correctly this would require changing the segment() method interface to a more cumbersome one.

PR #532, opened 2026-03-05
feat: split html table to headers and body

odelliab:
implement get_header_and_body_lines() for HtmlTableSerializer

### DIFF
--- a/test/test_hybrid_chunker.py
+++ b/test/test_hybrid_chunker.py
@@ -632,33 +632,7 @@ def test_html_parser_error_handling():
     
     serializer = chunker.serializer_provider.get_serializer(dl_doc)
     
-    # Test 1: Valid HTML should parse correctly
-    valid_html = "<table><tr><th>Header</th></tr><tr><td>Data</td></tr></table>"
-    header_lines, body_lines = serializer.table_serializer.get_header_and_body_lines(table_text=valid_html)
-    assert len(header_lines) > 0, "Valid HTML should produce header lines"
-    assert len(body_lines) > 0, "Valid HTML should produce body lines"
-    
-    # Test 2: Malformed HTML should be handled gracefully (fallback to string splitting)
-    malformed_html = "<table><tr><th>Header</th><tr><td>Data</td></table>"  # Missing closing tags
-    try:
-        header_lines, body_lines = serializer.table_serializer.get_header_and_body_lines(table_text=malformed_html)
-        # Should not raise an exception - error is caught and handled
-        assert True, "Malformed HTML should be handled without raising exception"
-        assert len(header_lines)==0, "Should return empty header lines for malformed HTML"
-        assert len(body_lines)==0, "Should return empty body lines for malformed HTML"
-    except Exception as e:
-        pytest.fail(f"HTML parsing error should be caught, but got: {e}")
-    
-    # Test 3: Empty or invalid content should be handled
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -50,6 +50,7 @@ dependencies = [
     'typer (>=0.12.5,<0.22.0)',
     'latex2mathml (>=3.77.0,<4.0.0)',
     "defusedxml (>=0.7.1, <0.8.0)",
+    "bs4>=0.0.2",
```
ceberam: We should use the original library and set a version major limit. In `docling` library we have:

```
"beautifulsoup4 (>=4.12.3,<5.0.0)"
```

docling_core/transforms/serializer/html.py
```diff
@@ -386,6 +387,67 @@ def serialize(
...
+        Returns:
+            A tuple of (header_lines, body_lines) where a row is considered a header row
+            if it contains at least one non-empty <th> cell and all <td> cells are empty.           rows with header cells and body_lines contains rows with data cells.
+        """
+        # Find the position of the first <tr> and last </tr>
+        first_tr_pos = table_text.find("<tr")
+        last_tr_end_pos = table_text.rfind("</tr>")
+
+        if first_tr_pos == -1 or last_tr_end_pos == -1:
+            raise ValueError("No table rows found in the provided content")
```
ceberam: Maybe instead of throwing an error and stopping the parsing in this situation, we could gracefully log a warning message and return an empty tuple (`[], []`)

docling_core/transforms/serializer/html.py
```diff
@@ -386,6 +387,67 @@ def serialize(
...
+        rows_content = table_text[first_tr_pos:last_tr_end_pos]
+        footer_content = table_text[last_tr_end_pos:].strip()
+
+        headings_list = []
+        if header_content:
+            headings_list.append(header_content)
+        data_list = []
+
+        # Parse rows_content with BeautifulSoup
+        soup = BeautifulSoup(rows_content, "html.parser")
```
ceberam: I would wrap it with a `try` and `except` to prevent BeautifulSoup from breaking the whole conversion in case of parsing exceptions.

docling_core/transforms/serializer/html.py
```diff
@@ -386,6 +387,67 @@ def serialize(
...
+        table_text: str,
+        **kwargs: Any,
+    ) -> tuple[list[str], list[str]]:
+        """Get header lines and body lines from the HTML table.
+
+
+        Returns:
+            A tuple of (header_lines, body_lines) where a row is considered a header row
+            if it contains at least one non-empty <th> cell and all <td> cells are empty.           rows with header cells and body_lines contains rows with data cells.
+        """
```
ceberam, line 438: Typo in the docstrings. What about:


```suggestion
        """Get header lines and body lines from the HTML table.

        Returns:
            A tuple of (header_lines, body_lines) where a row is considered a header row if it contains at least one
                non-empty <th> cell and all <td> cells are empty.
        """
```

pyproject.toml
odelliab: done

docling_core/transforms/serializer/html.py
odelliab: done

PeterStaar-IBM (approved):

PR #536, opened 2026-03-08
feat: handle wide table outliers with LineBasedTokenChunker

### LINKED ISSUE #535: Handling wide tables with Hybrid and Line Chunkers
The `LineBasedTokenChunker`, added in #527, allowed the Hybrid chunker to have an option of `repeat_table_header`. This option, when enabled, persists the header of a table item in every chunk that contains the rows of the table item.

The following outliers in the above process have behaviors that could be addressed differently.
- When a prefix that contains more tokens than the `max_tokens` specified is sent to `LineBasedTokenChunker`, a warning is thrown mentioning the fact that the prefix is too long, and then the prefix is set to `""`. When used in combination with `HybridChunker`, the prefix is set to the table header. If the header happens to be too large w.r.t. the specified max token limit, it gets completely omitted, and the resulting chunks do not have the table header at all
- When the header fits in the chunk, but a line of the table does not fit into the chunk, the line is broken down to fit it into the remaining space of the chunk, and overflowed to the next chunk (which also contains the header). In wide tables, when the header is just below the token count, the remaining token space is too narrow to store any meaningful table data. This causes a large number of poor quality chunks. 
    - For example, in a wide table of 30 rows, if the header takes up 500 tokens of the available 512 token limit, 12 tokens remain as available space. When rows are chunked, each row could create 40+ chunks with row information in the last 12 tokens of every chunk. (30 rows could result in 1200 chunks, meaning the header is repeated 1200 times in the chunk text)

z404:
## Summary

This PR addresses edge cases in table header handling when using the `repeat_table_header` feature introduced in #527. It adds a new `omit_header_on_overflow` parameter to both `HybridChunker` and `LineBasedTokenChunker` to handle scenarios where table headers are too large relative to the token limit.

### Implementation Details

**LineBasedTokenChunker enhancements:**
- Modified `prefix_chunks` computed field to split oversized prefixes into multiple chunks instead of discarding them
- Updated `prefix_len` to return `0` when prefix needs splitting (instead of discarding the prefix)
- Enhanced `chunk_text()` method with overflow detection logic:
  - Checks if a line would fit without prefix when it doesn't fit with prefix
  - Omits prefix for that line when `omit_prefix_on_overflow=True`
  - Omits prefix for overflow chunks from split lines when `omit_prefix_on_overflow=True`
  - Issues a warning once when prefix is actually omitted

**HybridChunker integration:**
- Passes `omit_header_on_overflow` parameter to `LineBasedTokenChunker` in `segment()` method

### Modified Files

- `docling_core/transforms/chunker/line_chunker.py`: Enhanced prefix handling and overflow logic
- `docling_core/transforms/chunker/hybrid_chunker.py`: Added `omit_header_on_overflow` parameter
- `test/test_line_chunker.py`: Added comprehensive test coverage

### Testing

✅ Added comprehensive test coverage in `test_line_chunker.py` (11 new test cases):
-  Ensure correct prefix behavior across scenarios including always being applied by default, being omitted only when the overflow‑omission option is enabled, and handling the no‑prefix case cleanly.
- Validate overflow and line‑splitting behavior by confirming that oversized lines are properly split, overflow chunks follow the correct prefix rules, and mixed small/large line scenarios behave consistently.
- Verify warning logic by ensuring warnings are emitted for overflow
- Confirms long‑prefix handling by verifying that long prefixes are split into chunks instead of discarded.

## Benefits

This enhancement provides more flexible handling of large table headers:

1. **Prevents data loss**: Large headers are now split into chunks instead of being completely discarded
2. **Reduces chunk fragmentation**: With `omit_header_on_overflow=True`, wide tables with large headers generate fewer, higher-quality chunks
3. **Maintains backward compatibility**: Default behavior (`omit_header_on_overflow=False`) preserves existing functionality
4. **User control**: Applications can choose the appropriate strategy based on their use case:
   - RAG systems prioritizing context might keep `omit_header_on_overflow=False`
   - Systems prioritizing chunk quality might set `omit_header_on_overflow=True`

## Usage Example

```python
from docling_core.transforms.chunker import HybridChunker

# New behavior - omit headers when they would cause overflow
chunker = HybridChunker(
    repeat_table_header=True,
    omit_header_on_overflow=True  # prevents excessive fragmentation
)
```

## Related Issues

Closes #535.

### DIFF
--- a/docling_core/transforms/chunker/line_chunker.py
+++ b/docling_core/transforms/chunker/line_chunker.py
@@ -25,25 +25,12 @@ class LineBasedTokenChunker(BaseChunker):
     its own. This is particularly useful for structured content like tables, code, or logs
     where line boundaries are semantically important.
 
-    Parameters:
-        tokenizer: The tokenizer to use for counting tokens. Can be an instantiated
-            tokenizer object or a model name/path for a pretrained model.
-
-        prefix: Optional text that appears at the beginning of each chunk. This is
-            useful for adding context like table headers, column names, or metadata
-            that should be repeated in each chunk for standalone readability.
-
-        omit_prefix_on_overflow: Controls prefix behavior when a line is too large
-            to fit with the prefix but would fit without it. When False (default),
-            the line will be split across multiple chunks, each including the prefix.
-            When True, the prefix is omitted for that specific line/chunk to prevent
-            splitting, which preserves line integrity but may result in inconsistent
-            chunk formatting. This is particularly useful when individual lines are
-            close to the token limit and splitting them would break their semantic
-            meaning (e.g., table rows, code lines, log entries).
-
-        serializer_provider: Provider for document serialization during chunking.
-            Defaults to ChunkingSerializerProvider.
+    The chunker supports adding a repeated prefix to each chunk, which can be useful for
+    keeping context like headers or metadata, while also offering flexibility when unusually
+    long lines appear. If a line is too large to fit alongside the prefix, the chunker
+    can either split the line across multiple prefixed chunks to maintain a consistent
+    format, or temporarily drop the prefix for that line to preserve the line's
+    integrity.
 
     Note:
         If the prefix itself exceeds max_tokens, it will be split into multiple


docling_core/transforms/chunker/line_chunker.py
```diff
@@ -54,13 +59,44 @@ class LineBasedTokenChunker(BaseChunker):
...
+        if not self.prefix:
+            return []
+
         token_count = self.tokenizer.count_tokens(self.prefix)
         if token_count >= self.max_tokens:
             warnings.warn(
-                f"Chunks prefix: {self.prefix} is too long for chunk size {self.max_tokens} and will be ignored"
+                f"Chunks prefix is too long ({token_count} tokens) for chunk size {self.max_tokens}. "
+                f"It will be split into multiple chunks and only included in the first chunk(s). "
+                f"Consider increasing max_tokens to accommodate the full prefix in each chunk."
```
ceberam, line 86: You may want to consider adding a suggestion like: _Consider increasing max_tokens or setting omit_prefix_on_overflow=False if consistent formatting is required."_

z404: This warning can trigger even when omit_prefix_on_overflow=False, because this is the warning that occurs when the prefix is too big. Setting max tokens to a higher number is the only way this warning can be prevented. omit_prefix_on_overflow=False will still trigger this warning (as intended)

PR #540, opened 2026-03-09
fix: convert headings to plain text in table cells

mikemikimike:
## Description

According to markdown specs, headings are not allowed inside tables. This fix converts markdown headings to plain text when rendering content inside table cells.

## Fix

In MarkdownTextSerializer.serialize(), added in_table_cell parameter. The _format_heading() method now checks this flag and returns plain text instead of markdown headings when rendering inside table cells.

## Issue

Fixes docling-project/docling#2722

### DIFF
--- a/docling_core/transforms/serializer/markdown.py
+++ b/docling_core/transforms/serializer/markdown.py
@@ -196,6 +196,7 @@ class MarkdownTextSerializer(BaseModel, BaseTextSerializer):
         doc_serializer: BaseDocSerializer,
         doc: DoclingDocument,
         is_inline_scope: bool = False,
+        in_table_cell: bool = False,
         visited: Optional[set[str]] = None,  # refs of visited items
         **kwargs: Any,
     ) -> SerializationResult:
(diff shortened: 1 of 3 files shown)


ceberam:
Thanks @mikemikimike for suggesting this PR.
There were some issues that I have observed:
- `PlainTextTextSerializer._format_heading` required `in_table_cell: bool = False` in its signature
- A regression test was missing
- It would be good to ad docstrings to the related functions.

I have moved ahead and apply those fixes, so that we can merge this PR as soon as possible.

ceberam (approved):

mikemikimike:
Thanks for catching these issues and applying the fixes directly.

I verified the current PR head `8f2001ddd56c35765ab67bbd3945e286d2f110f6` on `fix/2722-table-heading`: it includes the `in_table_cell` default, regression coverage, and the related docstrings. The PR is mergeable and approved, and the package/check matrix is passing. No additional change is needed from my branch.

cau-git (approved):

PR #549, opened 2026-03-16
feat: DocChunk expansion

### LINKED ISSUE #542: DocChunk expansion
When chunking documents for retrieval, typically there is a limit on chunk size (due to embedding models limitations).
However, post retrieval tasks (such as re-ranking or generation) do not share this limit and actually benefit from larger context.
Take for example a long table, which is chunked, but the generator that is required to answer a question regarding the table, is likely to provide better answer if given the whole table.

A naive solution for that is numbering the chunks, and expand a retrieved chunk by retrieving also its adjacent chunks. However, this solution does not take into account the document structure, so we may end up with information which is not cohesive (e.g texts from different sections of the document).

The requested feature is a method, that given a DocChunk, expands it to another DocChunk with a cohesive content (e.g. subsection, table, ...).

odelliab:
## Overview
This PR introduces chunk expansion functionality to the `DocChunk` class, enabling chunks to be expanded to include complete document objects or full pages. It also includes a comprehensive test suite with 15 test cases.

Resolves #542

### 1. Chunk Expansion Methods
Three new methods added to `DocChunk` class:

#### `_get_top_containing_objects(doc: DoclingDocument) -> list[DocItem] | None`
- Traverses the document tree upward from chunk items to find top-level containing objects
- Maintains reading order as in the original document
- Returns `None` if no top objects found

#### `expand_to_object(dl_doc: DoclingDocument, serializer: BaseDocSerializer) -> DocChunk`
- Expands chunk to include complete top-level document objects (sections, tables, lists)
- Prevents partial object inclusion in chunks
- Returns original chunk if expansion fails

#### `expand_to_page(doc: DoclingDocument, serializer: BaseDocSerializer) -> DocChunk`
- Expands chunk to include all content from the pages it spans
- Useful for maintaining page-level context
- Returns original chunk if expansion not possible

### 2. Comprehensive Test Suite
Created `test/test_doc_chunk_expansion.py` with 15 test cases covering:

# Dependencies
No new dependencies added.

#Note:
In the current implementation the expanded chunk has the same headings as the original  chunk. 
This may need revisiting.


docling_core/transforms/chunker/doc_chunk.py
```diff
@@ -87,3 +92,71 @@ class DocChunk(BaseChunk):
...
+            return self
+
+        meta = copy(self.meta)
+        meta.doc_items = doc_items
+        return DocChunk(
+            text=content,
+            meta=meta,
+        )
+
+    def expand_to_page(self, doc: DoclingDocument, serializer: BaseDocSerializer) -> DocChunk | None:
```
ceberam: Small type annotation detail:

```suggestion
    def expand_to_page(self, doc: DoclingDocument, serializer: BaseDocSerializer) -> DocChunk:
```

docling_core/transforms/chunker/doc_chunk.py
```diff
@@ -87,3 +92,71 @@ class DocChunk(BaseChunk):
...
+                    # Extract doc_items from serialization result
+                    doc_items.append(top_object)
+
+                except Exception as e:
+                    _logger.warning(f"error in extacting text of {top_object}: {e}")
+        if len(content.strip()) == 0:
+            _logger.warning(f"expansion of {self} did not yield any text")
+            return self
+
+        meta = copy(self.meta)
```
ceberam: Just wondering: is the shallow copy sufficient? Maybe `deepcopy()` would be safer in case there are nested mutable objects.

docling_core/transforms/chunker/doc_chunk.py
```diff
@@ -87,3 +92,71 @@ class DocChunk(BaseChunk):
...
+        top_objects = self.get_top_containing_objects(dl_doc)
+        if not top_objects:
+            _logger.warning(f"error in getting top objects of {self}")
+            return self
+
+        content = ""
+        doc_items = []
+
+        for top_object in top_objects:
+            if isinstance(top_object, ListGroup | InlineGroup | DocItem):
```
ceberam: At this point `top_object` will always be a `DocItem`. Is this a wrong type hint or maybe you were expecting to serialize groups? In that case you may need to review `get_top_containing_objects`.

docling_core/transforms/chunker/doc_chunk.py
```diff
@@ -87,3 +92,71 @@ class DocChunk(BaseChunk):
...
+        )
+
+    def expand_to_page(self, doc: DoclingDocument, serializer: BaseDocSerializer) -> DocChunk | None:
+        page_ids = [i.page_no for item in self.meta.doc_items for i in item.prov]
+        ser_params: CommonParams | None = getattr(serializer, "params", None)
+        if len(doc.pages) == 0 or page_ids is None or len(page_ids) == 0 or not ser_params:
+            _logger.warning(f"cannot expand to page the following chunk: {self}")
+            return self
+
+        ser_params.pages = set(page_ids)
```
ceberam: I don't think it is a good idea to mutate the serializer's parameters, which could cause issues if the serializer is reused somewhere else.

docling_core/transforms/chunker/doc_chunk.py
```diff
@@ -87,3 +92,71 @@ class DocChunk(BaseChunk):
...
+        for top_object in top_objects:
+            if isinstance(top_object, ListGroup | InlineGroup | DocItem):
+                try:
+                    ser_res = serializer.serialize(item=top_object)
+                    content += ser_res.text + "\n"
+                    # Extract doc_items from serialization result
+                    doc_items.append(top_object)
+
+                except Exception as e:
+                    _logger.warning(f"error in extacting text of {top_object}: {e}")
```
ceberam: Small typo:

```suggestion
                    _logger.warning(f"error in extracting text of {top_object}: {e}")
```

docling_core/transforms/chunker/doc_chunk.py
```diff
@@ -87,3 +92,71 @@ class DocChunk(BaseChunk):
...
+        meta = copy(self.meta)
+        meta.doc_items = doc_items
+        return DocChunk(
+            text=content,
+            meta=meta,
+        )
+
+    def expand_to_page(self, doc: DoclingDocument, serializer: BaseDocSerializer) -> DocChunk | None:
+        page_ids = [i.page_no for item in self.meta.doc_items for i in item.prov]
+        ser_params: CommonParams | None = getattr(serializer, "params", None)
```
ceberam: Not a big fan of using `getattr` instead of the dot notation in pydantic models (for several reasons...).
If instead of a `BaseDocSerializer` you define `serializer` as a `DocSerializer` (it's its only direct implementation anyway), you can already access `params` with the dot notation.

test/test_doc_chunk_expansion.py
```diff
@@ -0,0 +1,512 @@
...
+    
+        
+    def test_expand_to_page_preserves_metadata(self, sample_doc_with_pages):
+        """Test that page expansion preserves metadata."""
+        chunker = HybridChunker(
+            tokenizer=HuggingFaceTokenizer.from_pretrained(
+                model_name=EMBED_MODEL_ID,
+                max_tokens=MAX_TOKENS,
+            ),
+        )
```
ceberam: Can we reuse the chunker across tests for efficiency?

docling_core/transforms/chunker/doc_chunk.py
```diff
@@ -87,3 +92,71 @@ class DocChunk(BaseChunk):
...
+        if len(doc.pages) == 0 or page_ids is None or len(page_ids) == 0 or not ser_params:
```
ceberam: `doc.pages` will be empty for all Docling documents generated by backends that don't support pagination (`supports_pagination()==False`). This is the case of some declarative backend parsers like the HTML and the Markdown parsers.
I'm OK with _doing nothing_ in this case. But I would rather return a `warnings.warn()` message instead of logging that message, since this is information directed to the user.

docling_core/transforms/chunker/doc_chunk.py
```diff
@@ -87,3 +92,71 @@ class DocChunk(BaseChunk):
     """Data model for document chunks."""
 
     meta: DocMeta
+
+    def get_top_containing_objects(self, doc: DoclingDocument) -> list[DocItem] | None:
```
ceberam: Just asking. Do we want to expose this function as public? Or is it just a helper for `expand_to_object`? If it is the second case, please mark prefix it with `_` to mark it as a private method.

docling_core/transforms/chunker/doc_chunk.py
```diff
@@ -87,3 +92,71 @@ class DocChunk(BaseChunk):
...
+            objects[obj.self_ref] = obj
+
+        # maintain the reading order as in the original document
+        doc_body_refs = [ref.cref for ref in doc.body.children]
+        doc_ordered_refs = [ref for ref in doc_body_refs if ref in objects]
+        if len(doc_ordered_refs) > 0:
+            return [objects[ref] for ref in doc_ordered_refs]
+        return None
+
+    def expand_to_object(self, dl_doc: DoclingDocument, serializer: BaseDocSerializer) -> DocChunk:
```
ceberam: Could we add some docstring documentation? You could reuse the method description on this PR. Please use the google conventions.

ceberam (changes requested):
@odelliab Thanks for suggesting this implementation!
I left some technical comments, but I also wanted to share some high level comments:

- ~Please, sync your fork, rebase to `main`, and resolve the current conflicts.~ All good! It was a temporary issue with Github's CI/CD.
- Can we recap the use case to build those expansion functions? I understood that we want to expand the context for the answer generation in RAG from the chunks obtained in the retrieval phase. If this is the case, do we really need to return a chunk with those functions? Or text would do the job? Because I see some drawbacks in this implementation choice:
  - we serialize again a document (or part of the document) that was already serialized, since the starting point is a chunk (presumably obtained by a chunker with a serializer)
  - the `expand_to_page` and `expand_to_object` trigger a new serialization that could be different to the one that generated the original chunk. What would be the consequences? What happens if, for instance, we run those functions with an HTML serializer on a chunk created using the default serializer (`ChunkingDocSerializer` with a markdown serialization)?
  - if the given chunk was generated by a HybridChunker that respects a maximum token length, wouldn’t it be logical that the new chunk from `expand_to_object` also respects that condition?
  - in general, we detach the chunk with the chunker that originated it.
- Isn’t it enough the chunk and the Docling document for those expansions? Why do we need a serializer? We can locate the necessary items and then concatenate the text as we do it in `BaseChunker.contextualize` function. But maybe I’m missing a use case that would require building a new chunk.
- I would use the word `item` instead of `object` in the new method names, since that’s the terminology we used in Docling document (`TextItem`, `DocItem`, `ListItem`, …)
- Just wondering if we really nee 15 tests (thinking of keeping the `pre-commit` checks agile). Also, I would expect that we really check the content of the expanded chunks, to verify that they cover the expected doc items. For instance, I don’t see a test that checks that the output of `expand_to_page` contains all the doc items of the page or that the `text` field is exactly what we expect to get.
  - Can we reuse artifacts across the tests, like the HybridChunker?
- We could also let other users know about the potential of these expansions for RAG, by adding a small section in https://docling-project.github.io/docling/examples/advanced_chunking_and_serialization/ (in another PR on **docling** library).

docling_core/transforms/chunker/doc_chunk.py
odelliab: added

docling_core/transforms/chunker/doc_chunk.py
odelliab: Not sure why is this the case. A DocItem can be part of a GroupItem that resides in doc.body. No ?

See for instance test\data\chunker\0_inp_dl_doc.json. #/texts/113 -> #/groups/5 -> body

docling_core/transforms/chunker/doc_chunk.py
odelliab: right

docling_core/transforms/chunker/doc_chunk.py
odelliab: changed to DocSerializer

docling_core/transforms/chunker/doc_chunk.py
odelliab: changed to warnings

test/test_doc_chunk_expansion.py
odelliab: done

docling_core/transforms/chunker/doc_chunk.py
odelliab: Good call ! , used deepcopy for the serializer as well

docling_core/transforms/chunker/doc_chunk.py
odelliab: done

ceberam:
@odelliab We have discussed this PR with @vagenas and we found out that:
- chunk expansion is a feature that clearly brings value in Docling for supporting downstream applications such as RAG
- the method `expand_to_item` has a specific purpose but we think other users would be interested in other expansion rules (like down in the tree instead of upwards), therefore we think that we should not compromise the signature of `DocChunk` with specific methods. Instead, we could create a new abstraction for expanding chunks, with 2 implementations corresponding to the ones you proposed. If later on we need other expansion methods, we can simply add new implementations of the abstraction.
- we prefer to keep `DocChunk` as a pure data model class and put transformations related to chunking in `transforms/chunker`.

I have pushed the implementation of the refactoring suggestion with a new commit 1be5787. Please, check and provide comments or suggestions.

ceberam (approved):

odelliab:
I checked the latest commit and I have no corrections.
I think the code is much better now, thanks

vagenas (approved):

PR #552, opened 2026-03-16
feat: add artifacts_dir parameter to export_to_markdown and export_to_html

olijacklu:
Related to docling-project/docling#3094

## Summary
- Added `artifacts_dir` parameter to `export_to_markdown()` and `export_to_html()`
- When provided with `ImageRefMode.REFERENCED`, images are automatically saved to the specified directory
- Reuses existing `_with_pictures_refs()` method

## Usage
```python
markdown = doc.export_to_markdown(
    image_mode=ImageRefMode.REFERENCED,
    artifacts_dir="./images"
)
```

## Future feature
I also reference this feature in my issue but I haven't added it for now (let me know if you think it could be worth it):
- `image_path_prefix` parameter to allow custom paths in the output (e.g., for CDN URLs or web server paths)


PeterStaar-IBM (dismissed):

docling_core/types/doc/document.py
```diff
@@ -5794,6 +5794,7 @@ def export_to_markdown(
         include_annotations: bool = True,
         mark_annotations: bool = False,
         compact_tables: bool = False,
+        artifacts_dir: Optional[Union[str, Path]] = None,
```
PeterStaar-IBM: let's keep the same name `image_dir`

docling_core/types/doc/document.py
```diff
@@ -6078,6 +6091,7 @@ def export_to_html(
         included_content_layers: Optional[set[ContentLayer]] = None,
         split_page_view: bool = False,
         include_annotations: bool = True,
+        artifacts_dir: Optional[Union[str, Path]] = None,
```
PeterStaar-IBM: let's keep the same name `image_dir`

PeterStaar-IBM (changes requested):

PeterStaar-IBM (dismissed):

Jyotiraditya21-bug:
Hi @olijacklu @PeterStaar-IBM! 👋

I've offered to help revive/expand this PR in [docling-project/docling#3094](https://github.com/docling-project/docling/issues/3094#issuecomment-5151658188) by contributing the missing test coverage and adding `image_path_prefix` support. 

Let me know if you would like me to submit a PR to your branch or if I should open a new PR linking back here.

siddharthgaur1:
The only red check here is `codecov/patch` — everything else has been green since 08-01.

I've opened olijacklu/docling-core#1 against this branch with tests for the `image_dir` guard, so the approval here isn't lost to a fresh PR. Five tests covering both sides of both guards: the REFERENCED path for `export_to_markdown` and `export_to_html` (picture lands in `image_dir`, returned string references it), a `str` `image_dir` since the signature is `Optional[Union[str, Path]]`, and EMBEDDED/PLACEHOLDER with `image_dir` set to pin the untaken branch. Confirmed under `coverage` that every added statement in both guards is executed.

@olijacklu, merge it into your branch and this should go green. @Jyotiraditya21-bug offered the same tests on 08-01 — if yours are further along, use those instead; I only care that this stops being blocked.

Fixes docling-project/docling#3094 for the string-returning exports.

ceberam (changes requested):
Hi @olijacklu, thank you for the contribution! The feature is useful and the core idea of reusing `_with_pictures_refs` is the right approach. There are a few things to address before this can be merged.

- Rebase onto `main`. The branch has diverged by more than 100 commits and has a merge conflict, even if Github shows everything is green. Basically, this is because `main` now enforces some extra Ruff rules.
- `image_path_prefix` is missing and the feature is broken without it. I know that you intentionally left it aside, but I think we definitely need it if we add `image_dir`. `export_to_*` methods return a string in memory, there is no output file path, so there is no anchor from which to compute relative URIs. `_with_pictures_refs` is called without a `reference_path`, which means the image URIs embedded in the output will be absolute filesystem paths, making the output non-portable on any machine. By the way, I would call it `image_uri_prefix` instead of `image_path_prefix`.
- Silent no-op when `image_mode` is not `REFERENCED`. If a caller passes `image_dir` but forgets to set `image_mode=ImageRefMode.REFERENCED` (the default is `PLACEHOLDER`), the directory is silently ignored and no images are saved. What about a guard like this:
  
  ```python
  if image_dir is not None and image_mode != ImageRefMode.REFERENCED:
      raise ValueError(
          "`image_dir` requires `image_mode=ImageRefMode.REFERENCED`"
      )
  ```
- No tests added. At minimum, I would add tests for:
  - `export_to_markdown(image_mode=ImageRefMode.REFERENCED, image_dir=tmp_path)` saves a file and the returned string contains the expected URI.
  - Same for `export_to_html`.
  - Error path: passing `image_dir` without `REFERENCED` raises `ValueError`.
  Keep in mind that tests now go to `tests/` folder, like in any other Docling repo.
- Minor style comments:
  - the `doc = self` is unnecessary. A simple conditional expression is cleaner:
    ```python
    doc = (
        self._with_pictures_refs(image_dir=Path(image_dir), page_no=page_no)
        if image_dir is not None
        else self
    )
    ```
  - inline comments: I would remove `# Handle image saving when image_dir is provided` and save some bytes since the code is self-evident.
  - `image_dir` is placed differently in the two signatures (after `compact_tables` in `export_to_markdown`, last in `export_to_html`). Please make them consistent.

olijacklu:
Hi @ceberam !
Thank you for the feedback on this, I have rebased the branch onto main and made some further small changes based on your comments. Please let me know if this is what you had in mind (especially for `image_uri_prefix`).

ceberam (changes requested):
@olijacklu Thanks for addressing all the points!

Just a small detail. The standalone `tests/test_image_dir_export.py` file is not wrong, but four tests do not justify a dedicated module. `tests/test_serialization.py` is the right home, and specifically, the tests should be split across it rather than kept together in one block, each placed next to its thematic neighbors.

The two happy-path tests (`test_export_to_markdown_referenced_saves_images`, `test_export_to_html_referenced_saves_images`) belong after the `test_referenced_image_*` cluster at the end of the file, since they exercise the same `REFERENCED` mode but via the new `image_dir`/`image_uri_prefix` parameters. The two `ValueError` guard tests can sit immediately before them.

olijacklu:
@ceberam thanks for the feedback! I've moved the tests to the serialization file.😊

ceberam (approved):
Excellent! Thanks again @olijacklu  for the contribution 🚀

PeterStaar-IBM (approved):
lgtm!

PR #562, opened 2026-03-22
perf: Use OpenCV over PIL for PNG encoding in `ImageRef.from_pil`

maxdswain:
**Overview**
The `ImageRef.from_pil` class method is used widely in docling's codebase. It is often used several times per page when parsing documents in the `docling_parse.pdf_parser.PdfDocument._to_bitmap_resources_from_decoder` method. From my profiling, I found that it took up ~45% of processing time when doing a `DocumentConverter` conversion with all AI models disabled. This led me looking into how it's performance can be improved.

The function uses pillow to encode the image to a png, which is notoriously slow. So I swapped it out with opencv, improving the performance of this function by ~55% for this simple test case:

```python
import timeit

setup = """
from PIL import Image as PILImage
from docling_core.types.doc import ImageRef
"""

stmt = """
fig_image = PILImage.new(mode="RGBA", size=(200, 400), color=(0, 0, 0))
image = ImageRef.from_pil(image=fig_image, dpi=72)
"""

result = timeit.timeit(stmt=stmt, setup=setup, number=1000)
print(f"1000 runs: {result:.4f}s")
print(f"Per call:  {result:.4f}ms")
# Before: 1000 runs: 3.9219s, Per call:  3.9219ms
# After: 1000 runs: 1.7429s, Per call:  1.7429ms
```

When using these changes in the main docling repo, it reduced by conversion time from 14.2 to 9.21 (~35%) when disabling all AI models.

One caveat is that I did add an extra dependency `opencv-python-headless`, however this is already a dependency in the main docling repo's `uv.lock`.


dolfim-ibm (commented):
I think we are better without the extra conflicting dependencies.

I propose we remove the public dependencies and add a dev dependency on the headless version.

maxdswain:
Thanks for the feedback @dolfim-ibm, I've made the changes you've suggested.

maxdswain:
@cau-git @dolfim-ibm I should addressed all of your concerns, is this ready to be merged?

cau-git:
@maxdswain I still need to test this in a setup with docling to see if any of this may cause differences in the test outputs there. I already confirmed it speeds up things as expected.

cau-git (dismissed):

cau-git (approved):

cau-git:
@maxdswain thanks for the patience, I made commits to align it with the latest upstream `main` branch and had to add some fixes and test updates for it to pass.

PeterStaar-IBM (approved):
lgtm!

ceberam (approved):
lgtm

PR #569, opened 2026-03-26
feat(markdown): add footnote serialization support

ShrillHarrier:
This PR is related to the [Improved Footnote Serialization in MarkdownDocSerializer](https://github.com/docling-project/docling/issues/3128). 

It is a Feature Request submitted by [simonschoe](https://github.com/simonschoe) in [docling-project](https://github.com/docling-project/docling).

The features added include serializing a footnote in the form `[^{Identifier}]: {Description}`. This is done for Table and Picture items, as footnotes are linked to those.

In general, footnotes in `.md` files should look like:

```
[^5]: https://github.com/tesseract-ocr/tesseract
[^6]: https://github.com/VikParuchuri/surya
[^7]: https://github.com/lukas-blecher/LaTeX-OCR
```

Resolves docling-project/docling#3128

### Tests Added:

- test_table_with_footnotes_markdown()
- test_picture_with_footnotes_markdown()
- test_table_export_to_markdown_with_footnotes()


docling_core/transforms/serializer/markdown.py
```diff
@@ -842,6 +857,36 @@ class MarkdownDocSerializer(DocSerializer):
...
+    @override
+    def serialize_footnotes(
+        self,
+        item: FloatingItem,
+        **kwargs: Any,
+    ) -> SerializationResult:
+        params: MarkdownParams = self.params.merge_with_patch(patch=kwargs)
+        results: list[SerializationResult] = []
+        if DocItemLabel.FOOTNOTE in params.labels:
+            results = []
```
ceberam, line 869: This line is redundant

```suggestion
```

docling_core/transforms/serializer/markdown.py
```diff
@@ -842,6 +857,36 @@ class MarkdownDocSerializer(DocSerializer):
...
+            for footnote in item.footnotes:
+                if isinstance(ftn := footnote.resolve(self.doc), TextItem):
+                    parts = ftn.text.split(" ", 1)
```
ceberam, line 872: The footnote parsing logic assumes a specific format (the identifier and the footnote text). This format is not clearly represented or documented. It would be good that the format is explicit, validated, and clearly documented. We should keep in mind that for the markdown footnote to work, _identifiers can be numbers or words, but they can’t contain spaces or tabs._.
In addition, I don't think that we keep the footnote references with the correct formatting (a caret and an identifier inside brackets, e.g., `[^1]`).

If I try to serialize a DoclingDocument from the test dataset, you'll see that the footnote _1 see huggingface.co/ds4sd/docling-models/_ (doc item with reference `#/texts/29`) is not properly serialized as in the markdown specification.

```python

with open("test/data/doc/2408.09869v3_enriched.json") as handler:
  content = handler.read()
doc = DoclingDocument.model_validate_json(content)
doc.export_to_markdown(image_mode=ImageRefMode.PLACEHOLDER)
```

test/test_markdown_footnotes.py
```diff
@@ -0,0 +1,137 @@
...
+from docling_core.types.doc import DocItemLabel, DoclingDocument, TableCell, TableData
+from docling_core.types.doc.document import (
+    CodeItem,
+    FieldHeadingItem,
+    FieldValueItem,
+    FormulaItem,
+    ListItem,
+    SectionHeaderItem,
+    TextItem,
+    TitleItem,
```
ceberam, line 11: Many of these imports are never used. Please, remove unused imports.

ceberam (changes requested):
Thanks @ShrillHarrier for suggesting this PR. Please, see my comments.
In general:
- Adding new tests with a programmatic example is fine but it is much more illustrative to show the impact of the new feature in a serialization with a _ground truth_ data file. This allows us to check how the output markdown file gets rendered in applications like Github or VSC. Please, check how this is done in other test modules.
- To keep the repository consistent, I would suggest that you add the tests in the module `test/test_serialization.py`, together with the other tests of the markdown serialization, instead of creating a separate module.
- There are some DoclingDocument files (`.json`) that do not serialize as expected. Please, check some of them and ensure that the markdown serialization generates the right footnotes hook and text. For instance, `test/data/doc/2408.09869v3_enriched.json` has a footnote. If you regenerate the ground truth files (by running the tests with env variable `DOCLING_GEN_TEST_DATA=1 `), I would expect that the markdown serialization gets updated with the new footnote serialization.

docling_core/transforms/serializer/markdown.py
ShrillHarrier: Will remove. Thanks.

docling_core/transforms/serializer/markdown.py
ShrillHarrier: Hi, ah I see. Okay, I will refine the serialization and retest on the DoclingDocument files.

test/test_markdown_footnotes.py
ShrillHarrier: Will remove.

ceberam:
@ShrillHarrier please let me know when the PR is ready for another review.
I would also encourage you to rebase the branch to the latest `main` and force-push your commits. We have recently done some changes in the CI/CD checks and it would be good to have them in your PR.

ShrillHarrier:
Hello @ceberam,

I have cleaned up the functionality and code in this PR. Since review I have completed the following:

- Migrated tests to `test/test_serialization.py` and make two of them compare to ground truth files
- For the .json -> .md serialization test (`test_md_footnotes_json`) added a small gt file for page 2 of the original json specifications
- Added validation to ensure identifiers are not blank, contain a space, or contain a `\t` character
- Included documentation comments to explicitly denote rules for parsing and identify footnotes
- Removed unused imports
- Rebased with the upstream repo and pushed

ceberam:
@ShrillHarrier getting back to this after an absence...could you please resolve the conflicts with `main`?

ShrillHarrier:
Hello @ceberam, I have merged with `main`. I have also increased code coverage to ensure codecov/patch passes. Please let me know if additional changes are required. Thanks.

ShrillHarrier:
Hello @ceberam I have re-merged with `main` to keep the branch updated with the `docling-project`.

docling_core/transforms/serializer/markdown.py
```diff
@@ -188,6 +188,39 @@ class MarkdownParams(CommonParams):
...
+        For formatting footnote definitions, we need validate, then split the text into two parts:
+        - [identifier] and [definition]
+
+        So, if we have something like "this is a footnote", we split it into:
+        - "this" and "is a footnote" -> "[^this]:" and "is a footnote" -> "[^this]: is a footnote"
+
+        Validation:
+        - Footnote identifiers must be not include spaces or tabs, word (before a space) and strip it.
+        - If there is only one string before a space, that is an identifier, and the description is empty (valid).
+        """
```
ceberam: Please, consider the following: 
- Use the google style for docstrings
- The method name suggests pure formatting, but it also performs validation and raises exceptions. Consider renaming it (e.g., to `_validate_and_format_footnote`)
- Fix grammatical errors (for instance, _we need validate_ should be _we need to validate_)

docling_core/transforms/serializer/markdown.py
```diff
@@ -188,6 +188,39 @@ class MarkdownParams(CommonParams):
...
+        if not identifier.strip():
+            raise ValueError("Footnote identifier cannot be empty")
+
+        if "\t" in identifier or " " in identifier:
+            raise ValueError(f"Footnote identifier '{identifier}' contains invalid characters")
+
+        if len(parts) == 2:
+            return f"[^{identifier}]: {parts[1]}"
+        else:
+            return f"[^{identifier}]:"
```
ceberam: The `identifier` is not escaped or validated for special Markdown characters like ], [, or :. 
While unlikely, if an identifier contains these characters, it could break the footnote syntax.

docling_core/transforms/serializer/markdown.py
```diff
@@ -853,6 +903,27 @@ class MarkdownDocSerializer(DocSerializer):
...
 
+    @override
+    def serialize_footnotes(
+        self,
+        item: FloatingItem,
+        **kwargs: Any,
+    ) -> SerializationResult:
+        params: MarkdownParams = self.params.merge_with_patch(patch=kwargs)
+        results: list[SerializationResult] = []
+        if DocItemLabel.FOOTNOTE in params.labels:
```
ceberam: Nano-improvement 🙂

```suggestion
        if DocItemLabel.FOOTNOTE in params.labels:
            results: list[SerializationResult] = []
```

docling_core/transforms/serializer/markdown.py
```diff
@@ -188,6 +188,39 @@ class MarkdownParams(CommonParams):
...
+
+        parts = text.split(" ", 1)
+        identifier = parts[0]
+
+        # Validate identifier is not empty and doesn't contain tabs or whitespace
+        if not identifier.strip():
+            raise ValueError("Footnote identifier cannot be empty")
+
+        if "\t" in identifier or " " in identifier:
+            raise ValueError(f"Footnote identifier '{identifier}' contains invalid characters")
```
ceberam: The validation only checks for spaces and tabs in the identifier, but doesn't validate against newlines (\n, \r). Markdown footnote identifiers should not contain newlines as they would break the footnote syntax.


```suggestion
        if any(char in identifier for char in {"\t", " ", "\n", "\r"}):
            raise ValueError(f"Footnote identifier '{identifier}' contains invalid characters")
```

docling_core/transforms/serializer/markdown.py
```diff
@@ -853,6 +903,27 @@ class MarkdownDocSerializer(DocSerializer):
...
+        item: FloatingItem,
+        **kwargs: Any,
+    ) -> SerializationResult:
+        params: MarkdownParams = self.params.merge_with_patch(patch=kwargs)
+        results: list[SerializationResult] = []
+        if DocItemLabel.FOOTNOTE in params.labels:
+            for footnote in item.footnotes:
+                if isinstance(ftn := footnote.resolve(self.doc), TextItem):
+                    formatted_text = MarkdownTextSerializer._format_footnote_text(ftn.text)
+                    results.append(create_ser_result(text=formatted_text, span_source=ftn))
```
ceberam: If `footnote.resolve(self.doc)` returns None or a non-`TextItem`, the footnote is silently skipped. This could hide data integrity issues. Consider logging a warning for debugging purposes.

test/test_serialization.py
```diff
@@ -613,6 +613,77 @@ def test_md_traverse_pictures():
...
+            image_placeholder="<!-- image -->",
+        ),
+    )
+    actual = ser.serialize().text
+    verify(exp_file=src.with_suffix(".yaml.md"), actual=actual)
+
+def test_md_footnotes_json():
+    src = Path("./test/data/doc/2408.09869v3_enriched.json")
+    doc = DoclingDocument.load_from_json(src)
+
```
ceberam: Please, apply the Ruff linting on the test file (we need to fix the `pre-commit` to prevent this cases though...). For instance: `Blank line contains whitespace` on this line.

```suggestion

```

ceberam (changes requested):
Thanks again @ShrillHarrier for all the efforts on this PR.

I left some minor technical comments.

However, my biggest concern is the proper rendering of those footnotes. The footnote itself looks well serialized according to the markdown specifications. However, this formatting (e.g., `[^2]: Example`) seems to require an anchor in the text. If there is no anchor, the footnote simply does not get displayed in some applications. Unfortunately, the `DoclingDocument` examples we have from PDFs, which were parsed from PDFs, do not keep a proper link between the anchored text (the _reference_) and the footnote itself. The changes from this PR would actually mask it. For instance, in `test/data/doc/2206.01062.yaml.paged.md`, the markdown serialization from this PR generates the text:

```text
To ensure that future benchmarks in the document-layout analysis community can be easily compared, we have split up DocLayNet into pre-defined train-, test- and validation-sets. In this way, we can avoid spurious variations in the evaluation scores due to random splitting in train-, test- and validation-sets. We also ensured that less frequent labels are represented in train and test sets in equal proportions.

[^2]: e.g. AAPL from https://www.annualreports.com/

Table 1 shows the overall frequency and distribution of the labels among the different sets. Importantly, we ensure that subsets are only split on full-document boundaries. This avoids that pages of the same document are spread over train, test and validation set, which can give an undesired evaluation advantage to models and lead to overestimation of their prediction accuracy. We will show the impact of this decision in Section 5.
```

Github renders this section like this:

Image

While before the changes, we could see this (not ideal but at least visible):

Image

### How to move forward

My idea would be to apply the new footnote formatting **only** if we see an anchor, so that we follow the Markdown syntax strictly and we are sure that Markdown editors will understand it.

The question is how the anchored text should be linked with the footnote. I would then apply the same pattern as the captions with the `PictureItem` objects.

- Check if a footnote is contained in a `FloatingItem` (through its `footnotes` field)
- Use that floating item to store the anchor and render it through the Markdown specs: a caret and an identifier inside brackets (e.g., `[^2]`)
- Render the footnote as you have already done: with a caret and the number inside brackets with a colon and text (`[^2]: My footnote.`).

It could help to create a sample `.json` document that includes a reference (anchor) in a paragraph and the footnote with the _right_ structure from above.

Later on, we could enhance the declarative backends in **docling** repository to follow this formatting (e.g., JATS, USPTO, Markdown,...) in a separate PR. Actually it would be helpful to have a round trip example with footnotes, parsing from and exporting to markdown (`.md` -> `.json` -> `.md`) to fully test this functionality.

Hope this helps and feel free to add any thoughts here.

Matthew-Panizza:
Hi @ceberam,

I have adjusted the code to only show references if an anchor exists. We use the footnotes field in `FloatingItems` to determine with certainty that a footnote exists. We then render the footnote anchor below the `FloatingItem` These changes should address the following requirements/steps: 

- Check if a footnote is contained in a `FloatingItem`. 
- Use that floating item to store the anchor and render it through the Markdown specs (this renders the identifier [^2] after the footnote/image, then the entire footnote after so it appears at the bottom of the page).
- Includes a new `.json` spec that converts `.md` file.
- Added extra tests to ensure anchored footnotes are rendered and unanchored ones are not.
- Also, addressed all PR comments.

With the current implementation, we only render the anchors in attached to `FloatingItems`, and not text. I will post an example of the generated `footnote_with_anchor_example.json`.

Matthew-Panizza:
This document demonstrates footnote anchoring across multiple element types.

## Observations

Our analysis relies on a well-established methodology. Regional revenue data was collected across two consecutive quarters from a standardised reporting framework used by all participating business units. The figures presented in Table 1 summarise gross revenue by region before tax adjustments, allowing direct comparison between the North and South territories over the reference period.

The North region consistently outperformed expectations driven by strong consumer demand and favourable currency movements, while the South region recorded more modest growth constrained by supply chain disruptions in the second quarter.

Table 1: Quarterly revenue by region.

| Region   | Q1 Revenue   | Q2 Revenue   |
|----------|--------------|--------------|
| North    | 1.2M         | 1.4M         |
| South    | 0.9M         | 1.1M         |

[^1]

[^1]: Revenue figures are in USD and exclude tax.

[^2]

[^2]: Figures are unaudited estimates.

The revenue growth observed in both regions aligns with the broader market trends reported in the same period. Variance between Q1 and Q2 is within acceptable tolerance bounds and does not indicate systemic reporting issues.

## Pipeline Architecture

The pipeline architecture is illustrated below. The diagram captures the primary processing stages from document ingestion through to structured output generation.

Figure 1: Pipeline overview diagram.



[^3]

[^3]: Diagram reproduced with permission from the original authors.

The diagram above illustrates the end-to-end flow of documents through the system. Each stage operates independently, enabling modular replacement of individual components without disrupting the overall pipeline. Further details are available in the appendix.

simonschoe:
@ShrillHarrier @Matthew-Panizza @ceberam this PR slept for a long time, any chance for a revival? 😇

ceberam:
@simonschoe thanks for the heads-up! Indeed there's been some effort on this task, so we should give an extra push and merge it this week. I'll reply with a new assessment asap.

simonschoe:
@ceberam thanks for the ongoing initiative, awesome!

simonschoe:
@ceberam sorry for inquiring again, still on your radar? 😋

ceberam:
> @ceberam sorry for inquiring again, still on your radar? 😋

Next on my list!

lawrenceakka:
The updated fixtures in this PR show body footnotes being dropped rather than serialised.

`MarkdownTextSerializer.serialize` returns empty text for every item labelled footnote, and the text is re-emitted only through a parent's serialize_footnotes. A footnote referenced from a paragraph rather than from a table or picture has no parent FloatingItem, so nothing emits it at all.

The effect is visible in the golden files:

- test/data/doc/2206.01062.yaml.md loses three footnotes, with no [^n]: definition replacing them.
- test/data/chunker/0_out_chunks.json and 0b_out_chunks.json each lose a footnote chunk entirely, so the text disappears from chunking as well as from Markdown export.

ceberam:
Thanks @ShrillHarrier  for the work on Markdown footnote serialization and your recent merge to `main`. Also thanks @lawrenceakka for spotting the regression introduced with the PR.

I would like to narrow the scope of this PR before merging. The current DoclingDocument model does not yet represent a true inline footnote reference from a paragraph (or another text item) to a footnote definition. I plan to address that separately by extending the data model with an explicit footnote-reference relationship, likely on `TextItem` or `DocItem`, including the information needed to place an inline anchor correctly.

For this PR, please keep the implementation focused on footnotes associated with floating items (`TableItem` and `PictureItem`) through `FloatingItem.footnotes`. In that case the serializer has an explicit owning item and can emit the synthesized anchor and definition, for example:

```markdown
[^1]

[^1]: Explanation.
```

Please also ensure that non-floating-item footnotes are not silently dropped. At present, `MarkdownTextSerializer.serialize()` returns an empty result for every `DocItemLabel.FOOTNOTE`, while `MarkdownDocSerializer.serialize_footnotes()` only emits footnotes reached through a floating item's `footnotes` field. As a result, body-level footnotes such as those in `tests/data/doc/2206.01062.yaml` disappear from Markdown output because they are not referenced by a table or picture.

Until the data model can represent an inline anchor, these standalone/body-level footnotes should be preserved as ordinary text at their document-tree position rather than emitted as dangling Markdown definitions. In other words:

- floating-item footnote with an explicit `FloatingItem.footnotes` reference: serialize with the Markdown anchor and definition;
- non-floating-item or body-level `FOOTNOTE` item: serialize its text normally, without synthesizing an anchor;
- do not emit a `[^id]: ...` definition when no corresponding anchor can be represented.

Could you please update the implementation and tests accordingly? In particular, I would expect a regression test proving that a `FOOTNOTE` child of `body` remains visible in the Markdown output, while the existing anchored table/picture tests continue to verify the floating-item behavior. Please also update the affected ground-truth files to reflect preservation rather than removal of body-level footnote text.

The explicit inline footnote data-model extension and support for paragraph/text-item anchors should remain outside the scope of this PR.

FYI @simonschoe , I'll discuss the datamodel extension next week with the Docling team and keep you in the loop

lawrenceakka:
> I would like to narrow the scope of this PR before merging. The current DoclingDocument model does not yet represent a true inline footnote reference from a paragraph (or another text item) to a footnote definition. I plan to address that separately by extending the data model with an explicit footnote-reference relationship, likely on `TextItem` or `DocItem`, including the information needed to place an inline anchor correctly.
> 
....

> 
> The explicit inline footnote data-model extension and support for paragraph/text-item anchors should remain outside the scope of this PR.
> 
> FYI @simonschoe , I'll discuss the datamodel extension next week with the Docling team and keep you in the loop


@ceberam See #757 😄

mfurkanakinci:
I tested PR head `8f31b1cc4364dd163d69c474aea7f683187c3412` with small synthetic document examples. The body-level footnote preservation is working, but there is another text-loss case for footnotes explicitly referenced by a table or picture.

For example, an attached footnote containing `[1] İlgili hüküm için bkz. m. 5.` is rejected by `_validate_and_format_footnote`, logged as malformed, and omitted entirely. The same happens with a tab or newline separating the marker from the text. The normal document traversal already excludes these items through `_footnotes_of_some_item`, so there is no later opportunity to emit their text.

Minimal reproduction:

```python
from docling_core.transforms.serializer.markdown import MarkdownDocSerializer
from docling_core.types.doc.labels import DocItemLabel

doc = DoclingDocument(name="synthetic_legal_footnote")
picture = doc.add_picture()
note = doc.add_text(
    label=DocItemLabel.FOOTNOTE,
    text="[1] İlgili hüküm için bkz. m. 5.",
    parent=picture,
)
picture.footnotes.append(note.get_ref())
print(MarkdownDocSerializer(doc=doc).serialize().text)
```

Current output contains only `<!-- image -->`. I would expect the unsupported marker and its text to remain visible as ordinary text, without synthesizing an anchor.

I prepared a small [supplemental fix](https://github.com/mfurkanakinci/docling-core/commit/eaf6fed49fed2f0452851839a4ee31dcad1452a1) ([patch](https://github.com/mfurkanakinci/docling-core/commit/eaf6fed49fed2f0452851839a4ee31dcad1452a1.patch)), based directly on this PR's current head, that falls back to the configured text serializer on validation failure. It also removes the parent-based suppression in `MarkdownTextSerializer`: the document serializer already handles suppression of explicitly owned footnotes, and the extra guard prevents the fallback from working.

Eight regression cases cover tables and pictures, bracketed markers, tabs, newlines, and colon markers. They also check text appears once, source items remain available, valid sibling footnotes still receive anchors, HTML is escaped, and label filtering still applies. All eight fail on the current PR head and pass with the patch; the complete serialization module passes all 115 tests. The full repository pre-commit checks also pass, including MyPy, both complete test suites, documentation generation, and lockfile checks.

This keeps the existing validation rules and stays within the floating-item scope of this PR.

ShrillHarrier:
Thank you @mfurkanakinci for the supplemental fix. Now, this PR covers the following functionality:

- Footnotes attached to a floating item that cannot be serialized as a Markdown anchor are not dropped, they are serialized as plain text.
- Footnotes that are valid or invalid but have no anchor are not modified, they are preserved as plain text.
- Footnotes that are valid and have a floating item are serialized properly, with `[^id]: definition` or just `[^id]:`.

Given these specs, this PR is ready to be reviewed again. I have merged to main and fixed merge conflict. @ceberam if you get a chance, can you take a look at the PR? Thanks.

PR #581, opened 2026-04-09
perf: contextualize function() skip excluded fields

jordanpagewhite:
This change proposes that we use model_dump to skip/avoid serializing fields that are in excluded_embed. This should reduce unnecesary computation and is also more precisely what I believe we are intending to do here in that we do not need to unnecessarily filter the results below in the for-loop.


jordanpagewhite:
After further testing, I think that this is an extremely tiny performance improvement in the context of the greater chunking workflow. However, I still think that it could be worth avoiding the serialization of these excluded_embed fields as it seems unnecessary in the scope of this function.

ceberam:
Thanks @jordanpagewhite for spotting this performance improvement, even if it's extremely tiny 🙂  
And apologies for the late response. We value your contribution to **docling-core** ! 

I have resolve the existing conflict with `main` and added a guard against aliases. It should be ready for merging. 🚀

ceberam (dismissed):

ceberam (approved):

cau-git (approved):

PR #585, opened 2026-04-10
feat(document): add artifacts_dir param to export_to_markdown

Smeet23:
## Summary

- Adds an optional `artifacts_dir: Optional[Path]` parameter to `DoclingDocument.export_to_markdown()`
- When `image_mode=ImageRefMode.REFERENCED` and `artifacts_dir` is provided, images are automatically saved to that directory and the returned markdown contains relative paths referencing them
- When `artifacts_dir` is `None` (default) or `image_mode` is not `REFERENCED`, behaviour is unchanged — fully backwards-compatible

**Before:**
```python
# Manual image saving required before calling export_to_markdown
for item, _ in doc.iterate_items():
    if isinstance(item, PictureItem):
        img = item.get_image(doc=doc)
        img.save(f"./images/image_{i}.png")
        item.image.uri = Path(f"./images/image_{i}.png")

md = doc.export_to_markdown(image_mode=ImageRefMode.REFERENCED)
```

**After:**
```python
md = doc.export_to_markdown(
    image_mode=ImageRefMode.REFERENCED,
    artifacts_dir=Path("./images"),
)
```

## Relation to upstream issue

Resolves docling-project/docling#3094

## Test plan

- [x] `test_export_to_markdown_with_artifacts_dir` — verifies images are saved and referenced in markdown
- [x] `test_export_to_markdown_referenced_without_artifacts_dir` — verifies fallback when `artifacts_dir` is omitted
- [x] `test_export_to_markdown_artifacts_dir_ignored_for_non_referenced` — verifies no side-effects for `PLACEHOLDER`/`EMBEDDED` modes
- [x] Existing `test_save_to_disk`, `test_construct_doc`, `test_save_pictures` all still pass

### DIFF
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -5903,6 +5903,7 @@ class DoclingDocument(BaseModel):
         image_placeholder: str = "<!-- image -->",
         enable_chart_tables: bool = True,
         image_mode: ImageRefMode = ImageRefMode.PLACEHOLDER,
+        artifacts_dir: Optional[Path] = None,
         indent: int = 4,
         text_width: int = -1,
         page_no: Optional[int] = None,
@@ -5947,6 +5948,12 @@ class DoclingDocument(BaseModel):
         :param image_mode: The mode to use for including images in the
             markdown. (Default value = ImageRefMode.PLACEHOLDER).
         :type image_mode: ImageRefMode = ImageRefMode.PLACEHOLDER
+        :param artifacts_dir: Directory where images are saved when
+            image_mode is ImageRefMode.REFERENCED. The markdown output will
+            contain relative paths pointing into this directory. Ignored when
+            image_mode is not ImageRefMode.REFERENCED.
+            (Default value = None).
+        :type artifacts_dir: Optional[Path] = None
         :param indent: The indent in spaces of the nested lists.
             (Default value = 4).
         :type indent: int = 4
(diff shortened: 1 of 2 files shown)


Smeet23:
Hi @PeterStaar-IBM and @dolfim-ibm — could you take a look at this PR when you get a chance? Mergify requires 2 approvals since test data was updated. DCO and all other checks are passing. Would really appreciate a review!

PeterStaar-IBM (dismissed):

Smeet23:
Hi @PeterStaar-IBM — thanks for the approval! I've pushed a small follow-up fix: `ruff` reformatted one line in `document.py` that was too long — that was the only CI failure. DCO ✅ is passing. Could you re-approve the new commit when you get a chance so Mergify can proceed? Happy to address any other feedback. Thanks!

Smeet23:
Hi @PeterStaar-IBM and @dolfim-ibm — just a gentle ping on this one. DCO ✅ is passing and all checks are green. Would appreciate a re-approval so Mergify can proceed. Happy to address any feedback. Thanks!

cau-git (changes requested):
@Smeet23 can you please explain in which case you need this? The design principle is that `export_to_*` methods must stay pure, they have no business writing any files, only return the result in memory. For markdown that means there can't be references. You can use `save_as_markdown` for that, which already supports an artifact dir.

Smeet23:
Thanks @cau-git, that's a fair point and I agree with the principle.

The use case (from docling-project/docling#3094) was getting the markdown back **in memory as a string** while having the referenced images written to disk — without the manual "iterate items, save each image, rewrite the URIs" boilerplate before calling `export_to_markdown`. So `artifacts_dir` was a convenience to fold that into one call.

But you're right that this makes `export_to_markdown` impure, and that `save_as_markdown(..., artifacts_dir=...)` already covers writing markdown + artifacts to disk. The only thin gap is "return the string rather than write the `.md` file," which isn't worth breaking the purity guarantee of `export_to_*`.

Happy to **close this PR** unless you'd prefer I instead add a short docs example showing `save_as_markdown` for this use case. Either works for me — just let me know.

cau-git:
@Smeet23 Doing a short docs example for the use case would be welcome! I suggest doing that on a fresh PR, and close this one.

Smeet23:
Sounds good @cau-git — closing this. I'll open a separate PR with a short docs example showing `save_as_markdown(..., artifacts_dir=...)` for this use case. Thanks for the guidance!

PR #586, opened 2026-04-10
fix: consider rich table cell refs in tree operations

vagenas:
## Fix: Consider rich table cell refs during tree operations

This commit fixes an issue where tree operations weren't handling references from `RichTableCell` objects, causing broken document hierarchies.

### Changes:

1. **Updated `_shift_up`**: When removing a node, updates any table cell references to point to the node's child instead (errors if the node has multiple or no children).
1. **Updated `_shift_down`**: Updates table cell references to point to the new wrapper node before moving the subtree.
1. **Updated `_move_subtree`**: Prevents moving nodes that are referenced by table cells.
1. **Improved `fix_misplaced_list_items`** auto-healing: Uses updated `_shift_down` and `_move_subtree` instead of deleting and recreating items, which preserves references correctly.

### Tests:
- Added 3 test cases for rich table cell reference handling
- Updated expected output for misplaced list items test

### DIFF
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -3453,6 +3453,9 @@ class DoclingDocument(BaseModel):
             node = child_ref.resolve(self)
             self._update_breadth_first_with_lookup(node=node, refs_to_be_deleted=refs_to_be_deleted, lookup=lookup)
 
+    def _find_referencing_table_cells(self, table: TableItem, ref: str) -> list[RichTableCell]:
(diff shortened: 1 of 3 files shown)


PeterStaar-IBM (approved):

ceberam (changes requested):
@vagenas there is a conflict with `test_docling_doc.py`

In the meantime...Besides nodes referenced by rich table cells, what about items referenced by a `GraphCell` or those referenced by a `FloatingItem` through `captions`, `references`, or `footnotes` ?

loickouao:
Do you have any updates regarding this issue? I find it very interesting

PR #589, opened 2026-04-15
feat: add _to_referenced_images and _to_embedded_images to DoclingDocument

OfekDanny:
## Summary

Closes #3094

Adds two named transformation methods on `DoclingDocument` that follow the same pattern as `_hierarchize()` and `_flatten()` — they operate on a deep copy and leave the source document unchanged, as suggested by @PeterStaar-IBM in docling-project/docling#3288.

### `_to_referenced_images(image_dir, image_path_prefix="")`
Saves each `PictureItem` image as `image_001.png`, `image_002.png`, … into `image_dir` and updates the stored URI to `image_path_prefix + filename`. This makes it easy to call `export_to_markdown(image_mode=ImageRefMode.REFERENCED)` with full control over where images land and what prefix appears in the output — without writing the whole document to disk first.

### `_to_embedded_images()`
Returns a deep copy where every image referenced through a file URI is converted to inline base64 form (delegates to `_with_embedded_pictures()`).

### Usage

```python
from pathlib import Path

# Save images to disk, get markdown with relative references
doc_ref = doc._to_referenced_images(
    image_dir=Path("output/images"),
    image_path_prefix="images/",
)
md = doc_ref.export_to_markdown(image_mode=ImageRefMode.REFERENCED)

# Get a fully self-contained document with embedded images
doc_emb = doc._to_embedded_images()
md = doc_emb.export_to_markdown(image_mode=ImageRefMode.EMBEDDED)
```

## Tests

Three new tests added to `test/test_docling_doc.py`:

- `test_to_referenced_images` — verifies images are written with sequential filenames (`image_001.png`) and URIs are set to `prefix + filename`; checks source document is not mutated
- `test_to_referenced_images_markdown_output` — verifies the full pipeline: transform → `export_to_markdown` emits the correct image reference
- `test_to_embedded_images` — verifies the method returns an independent deep copy

## Design notes

- Both methods follow the `_hierarchize` / `_flatten` pattern (private helper, returns a new `DoclingDocument`)
- `_to_embedded_images` is a thin, named wrapper over `_with_embedded_pictures()` to make the intent explicit at call sites
- No existing behaviour is changed; the two lower-level helpers (`_with_pictures_refs`, `_with_embedded_pictures`) remain in place

Signed-off-by: OfekDanny <ofekdanny@gmail.com>

### DIFF
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -5703,9 +5703,10 @@ class DoclingDocument(BaseModel):
         """Create a copy of this document with images saved as referenced files.
 
         Creates a deep copy of the document where each :class:`PictureItem`'s
-        image is saved as a PNG file inside *image_dir* and the stored URI is
-        updated to ``image_path_prefix + <filename>`` so that the exported
-        markdown (or any other format) emits the desired relative reference.
+        image and each page image is saved as a PNG file inside *image_dir* and
+        the stored URI is updated to ``image_path_prefix + <filename>`` so that
+        the exported markdown (or any other format) emits the desired relative
+        reference.
 
         This is the recommended way to obtain a document ready for
         ``export_to_markdown(image_mode=ImageRefMode.REFERENCED)`` when you
@@ -5729,8 +5730,8 @@ class DoclingDocument(BaseModel):
         :param image_path_prefix: String prepended to each image filename in
             the document's URI field (e.g. ``"images/"``).  Defaults to ``""``.
         :type image_path_prefix: str
-        :returns: A new :class:`DoclingDocument` whose picture URIs point to
-            the saved files.
+        :returns: A new :class:`DoclingDocument` whose picture and page image
+            URIs point to the saved files.
         :rtype: DoclingDocument
         """
         result: DoclingDocument = copy.deepcopy(self)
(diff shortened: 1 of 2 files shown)


docling_core/types/doc/document.py
```diff
@@ -5695,6 +5695,89 @@ def _with_pictures_refs(
...
+                    filename = f"image_{img_count + 1:03d}.png"
+                    loc_path = image_dir / filename
+                    img.save(loc_path)
+                    uri_path = Path(image_path_prefix + filename)
+                    if item.image is None:
+                        scale = img.size[0] / item.prov[0].bbox.width
+                        item.image = ImageRef.from_pil(image=img, dpi=round(72 * scale))
+                    item.image.uri = uri_path
+                img_count += 1
+
```
PeterStaar-IBM, line 5753: @OfekDanny I think it might make sense here to also add the saving of the page-images.

PeterStaar-IBM (changes requested):

docling_core/types/doc/document.py
OfekDanny: Done! I've updated `_to_referenced_images` to also iterate over `result.pages` and save each `PageItem.image` as `page_NNN.png`, updating its URI to `image_path_prefix + filename`. Added a dedicated test (`test_to_referenced_images_saves_page_images`) that builds a document with a page image, verifies the file is saved, and checks that the source document is not mutated.

PR #518, opened 2026-02-16
fix(chunker): propagate 'traverse_pictures' parameter to chunker

ceberam:
This is a fix for issue https://github.com/docling-project/docling/issues/1094
In a previous PR https://github.com/docling-project/docling-core/pull/501 , the issue seemed resolved, but it was not addressing the `Hierarchical` and `Hybrid` chunkers as pointed out in comment https://github.com/docling-project/docling-core/pull/501#issuecomment-3864893335

This PR propagates `traverse_pictures` parameter from the `serializer_provider` to the `HierarchicalChunker`, to ensure that chunks include any `TextItem` children from `PictureItem` if the parameter is set to `True`

PR #593, opened 2026-04-21
fix(chunker): forward content layers from serializer params to doc traversal

### LINKED ISSUE #592: HybridChunker ignores layers in serialization params
`HybridChunker` ignores layers in serialization params because the inner `HierarchicalChunker` does not pass them to `iterate_items` method:
https://github.com/docling-project/docling-core/blob/0425dc0c038b687c6fab7cda30da803b114d0515/docling_core/transforms/chunker/hierarchical_chunker.py#L171-L174

Related to https://github.com/docling-project/docling/issues/1325

Vdaleke:
## Summary

`HybridChunker` (and its underlying `HierarchicalChunker`) ignored the `layers`
parameter configured in serializer params. Items on non-default content layers
(e.g. `ContentLayer.FURNITURE`) were never included in chunks, even when the user
explicitly configured `MarkdownParams(layers={ContentLayer.BODY, ContentLayer.FURNITURE})`.

## Root cause

`HierarchicalChunker.chunk` already extracted `traverse_pictures` from the
serializer params and forwarded it to `dl_doc.iterate_items`, but did not do the
same for `params.layers`. As a result, `iterate_items` always used the default
content layers regardless of what was configured on the serializer, making the
`layers` parameter effectively a no-op for both chunkers.

## Changes

- **`hierarchical_chunker.py`**: extract `params.layers` from the serializer (when
  it is a `DocSerializer`) and pass it as `included_content_layers` to
  `dl_doc.iterate_items`, following the same pattern already used for
  `traverse_pictures`.
- **`hybrid_chunker.py`**: remove the redundant `doc_serializer=my_doc_ser` kwarg
  that was being passed to `_inner_chunker.chunk`. This kwarg was silently ignored
  by `HierarchicalChunker.chunk` — the inner chunker already builds its own
  serializer from the shared `serializer_provider`.

## Tests

Added `test_chunk_multiple_content_layers` to both `test_hierarchical_chunker.py`
and `test_hybrid_chunker.py`. Each test:

1. Asserts that the default chunker excludes furniture-layer items (`PAGE_HEADER`,
   `PAGE_FOOTER`).
2. Asserts that a chunker configured with `layers={BODY, FURNITURE}` correctly
   includes those items in the output.
3. Validates the full chunk structure against a JSON ground-truth snapshot.

Resolves #592

### DIFF
--- a/docling_core/transforms/chunker/hierarchical_chunker.py
+++ b/docling_core/transforms/chunker/hierarchical_chunker.py
@@ -209,9 +209,12 @@ class HierarchicalChunker(BaseChunker):
         ser_res = create_ser_result()
         excluded_refs = my_doc_ser.get_excluded_refs(**kwargs)
         traverse_pictures = my_doc_ser.params.traverse_pictures if isinstance(my_doc_ser, DocSerializer) else False
+        included_content_layers = my_doc_ser.params.layers if isinstance(my_doc_ser, DocSerializer) else None
+
         for item, level in dl_doc.iterate_items(
             with_groups=True,
             traverse_pictures=traverse_pictures,
+            included_content_layers=included_content_layers,
         ):
             if item.self_ref in excluded_refs:
                 continue
--- a/docling_core/transforms/chunker/hybrid_chunker.py
+++ b/docling_core/transforms/chunker/hybrid_chunker.py
@@ -329,7 +329,6 @@ class HybridChunker(BaseChunker):
         res: Iterable[DocChunk]
         res = self._inner_chunker.chunk(
             dl_doc=dl_doc,
-            doc_serializer=my_doc_ser,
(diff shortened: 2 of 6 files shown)


Vdaleke:
Hello @ceberam, this PR is similar to yours #518, could you review it?

Vdaleke:
@dolfim-ibm, @ceberam really need this feature, could you assign someone to this PR and issue?

ceberam:
Thanks @Vdaleke for spotting this bug and providing a patch.
And apologies for the late response. We value your contribution to **docling-core** 🚀

I have resolve the existing conflict and approved it.

ceberam (approved):

ceberam (approved):

cau-git (approved):

PR #596, opened 2026-04-26
fix(chunking): handle empty TripletTableSerializer output without consuming children

Vjc5h3nt:
## Summary
Fixes docling-project/docling#3335.

The chunkers were returning 0 chunks for some DOCX files because of a issue in `TripletTableSerializer`. When a table ended up producing empty text (header-only tables, or layout/wrapper tables whose cells are `RichTableCell` refs into body content), the serializer was still adding those child refs to the shared `visited` set on its way through `_export_to_dataframe_with_options`. The downstream walk then thought the body had already been chunked and skipped it, so `HierarchicalChunker` and `HybridChunker` would return nothing.
## What changed
All in `docling_core/transforms/chunker/hierarchical_chunker.py`:
- The dataframe export now runs against a local copy of `visited`. Its contents are merged back into the shared set only if the table actually produces non-empty text. Empty tables no longer consume their siblings' refs.
- Header-only tables (0 data rows but non-empty columns) now emit the column headers as text instead of being dropped silently.
- Single-row single-column tables emit the cell text directly. Previously this branch built an empty triplet list.
- Added a small `_flatten_table_text` helper as a last-resort fallback for tables whose triplet form still collapses to empty text.
- Stopped appending empty `SerializationResult`s to `parts`.
## Tests
Six regression tests in `test/test_hierarchical_chunker.py` and `test/test_hybrid_chunker.py`:
- `test_triplet_table_serializer_header_only` - header-only multi-column table
- `test_triplet_table_serializer_single_row_single_column` - header-only single-cell table
- `test_chunker_header_only_table_produces_chunks` - `HierarchicalChunker` on a doc with a header-only table
- `test_triplet_table_serializer_falls_back_to_single_cell_text` - single-cell `RichTableCell` fallback
- `test_triplet_table_serializer_does_not_consume_empty_children` - confirms `visited` isolation when text is empty
- `test_chunk_single_cell_rich_table` - `HierarchicalChunker` and `HybridChunker` on a single-cell rich table

## Verification
Reproduced the issue's headline symptom synthetically (layout-wrapper table whose `RichTableCell` references body items):

|  | `main` | this branch |
|---|---|---|
| `table_serialize_text_len` | 0 | 328 |
| `visited_after / total_items` | 7 / 7 | 7 / 7 |
| `hierarchical_chunk_count` | **0** | **1** (all body content preserved) |

Also verified end-to-end on a generated DOCX with mixed table types (header-only, normal data, single-column):

|  | `main` | this branch |
|---|---|---|
| tables that produced any text | 2 of 3 | 3 of 3 |
| `hierarchical_chunk_count` | 24 | 25 (+1 recovered table) |

The header-only table that previously serialized to empty now emits `'Field ID. Description. Unit. Source'`. No regression on normal data tables , the 5-column data table (text_len 1268) and the single-column table (text_len 145) produce identical output character-for-character on both branches.

## Test plan
- [x] `uv run pre-commit run --all-files` — all hooks passed (ruff format, ruff lint, mypy, pytest, docs, uv-lock)
- [x] `pytest test/` - 450 passed
- [x] `ruff check` on changed files — clean
- [x] 6 new regression tests added, all pass

### DIFF
--- a/docling_core/transforms/chunker/hierarchical_chunker.py
+++ b/docling_core/transforms/chunker/hierarchical_chunker.py
@@ -6,6 +6,7 @@ import logging
 from collections.abc import Iterator
 from typing import Annotated, Any, Optional, Union
 
+import pandas as pd
 from pydantic import ConfigDict, Field
 from typing_extensions import override
 
@@ -45,6 +46,25 @@ _logger = logging.getLogger(__name__)
 class TripletTableSerializer(BaseTableSerializer):
     """Triplet-based table item serializer."""
 
+    @staticmethod
+    def _flatten_table_text(table_df: pd.DataFrame) -> str:
+        """Last-resort fallback that turns a table into plain text.
+
+        The preferred output of this serializer is the triplet form
+        'row, column = value'. When that comes out empty - for example on
+        single-cell RichTableCell layout tables where the header and row
+        labels are blank - we fall back to this helper so the table still
+        contributes something chunkable and does not end up consuming its
+        sibling refs.
+
+        Cells are read row by row, blanks are skipped, and the remaining
+        values are joined with '. '. So a table like [['A', ''], ['B', 'C']]
+        becomes 'A. B. C'.
+        """
+        return ". ".join(
+            text for row in table_df.itertuples(index=False, name=None) for value in row if (text := str(value).strip())
+        )
+
     @override
     def serialize(
         self,
@@ -56,6 +76,7 @@ class TripletTableSerializer(BaseTableSerializer):
     ) -> SerializationResult:
         """Serializes the passed item."""
         parts: list[SerializationResult] = []
+        shared_visited = kwargs.get("visited")
 
         cap_res = doc_serializer.serialize_captions(
             item=item,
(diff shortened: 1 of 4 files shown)


docling_core/transforms/chunker/hierarchical_chunker.py
```diff
@@ -45,6 +45,13 @@
 class TripletTableSerializer(BaseTableSerializer):
     """Triplet-based table item serializer."""
 
+    @staticmethod
+    def _flatten_table_text(table_df) -> str:
+        """Flatten non-empty cell values into a chunkable text fallback."""
+        return ". ".join(
+            text for row in table_df.itertuples(index=False, name=None) for value in row if (text := str(value).strip())
+        )
```
ceberam: The docstring could benefit from more detail, about when it's used (as a fallback) and what format it produces. This would help future maintainers understand the serialization strategy hierarchy. In addition, please complete the type annotation of the method signature.

ceberam (changes requested):
Thanks @Vjc5h3nt for your contribution. The PR addresses the issue. I only have some remarks regarding the regression tests. They are not critical but they can help keeping the library tidy and easier to maintain by new developers, as well as reduce the burden of the CI/CD checks.
- Instead of repeating the function `_build_single_cell_rich_table_do` in the 2 test modules, place this function in `test_utils.py` and reuse it.
- Gather all the new checks related to this issue in `test_hierarchical_chunker.py` in a single test function, instead of creating multiple ones
- In tests, try to minimize the inline comments, since most of them are redundant (you added good explanations in the test function docstrings).
- omit references to the Git issue (_Regression test for ... issue 3335_), since it is not necessary and we do not want to maintain this type of references in the code (the Git issues and the CHANGELOG are already enough for this purpose).

ceberam (approved):
🚀

Vjc5h3nt:
Hi @cau-git , @dolfim-ibm ,  just a gentle reminder regarding this PR. It has been pending review for some time now. Could you please take a look when you have a moment?

dolfim-ibm (approved):

PR #601, opened 2026-05-04
fix(serializer): stop duplicating spanning cells in markdown/dataframe output

scottf007:
## Summary

The markdown table serializer and the legacy markdown / dataframe export paths iterate `data.grid` and emit each grid cell's text. When a `TableCell` has `col_span > 1` or `row_span > 1`, the same cell reference occupies every grid position it covers, so iterating produces N copies of the cell's text in the output instead of one.

The underlying `TableCell` data is correct (TableFormer emits one cell per merged region with `col_span` / `row_span` set). Only the serialized presentation duplicates.

## Visible symptoms

- A `col_span=4` header like \`KRYTERIA FORMALNE\` rendered as four repeated cells instead of one cell + three empties.
- A `row_span=2` cell like \`Energy reduction\` rendered twice in consecutive rows of the markdown table.
- `export_to_dataframe()` concatenated the parent header text with each child header (e.g. \`KRYTERIA FORMALNE.Lp.\`, \`KRYTERIA FORMALNE.TAK\`).

## Fix

In all three code paths:

- `MarkdownTableSerializer.serialize` (modern path)
- `TableItem.export_to_markdown` (legacy path)
- `TableItem.export_to_dataframe`

emit cell text only at the cell's origin position (`start_row_offset_idx`, `start_col_offset_idx`). Spanned positions render as empty string. Cells with `col_span=1` and `row_span=1` are unaffected because their single grid position equals their origin.

## Test plan

- [x] \`pytest test/test_docling_doc.py\` — 67 pass after fixture regen
- [x] Full non-chunker test suite — 376 pass, 2 xfailed, no regressions
- [x] Manual run via \`docling\` on three PDFs that exercise spans:
  - CVS Group 2025 annual report (TCFD table, page 35) — \`Energy reduction\` rowspan now renders once
  - Polish gov "Czyste Powietrze" — \`KRYTERIA FORMALNE\` colspan=4 now renders once
  - iShares dividend statement — \`Net Cash for Reinvestment#\` colspan=2 now renders once
- [x] Regenerated 6 affected groundtruth fixtures (\`*.yaml.md\`, \`*.paged.md\`, \`constructed_doc*.md.gt\`). Diffs are precisely the removed cell-content duplication, nothing else.

## Issues addressed

Closes docling-project/docling#2862 — \`export_to_dataframe\` produces duplicated cells when tables contain spanning cells. The dataframe path is fixed by this change.

## Notes

- \`data.grid\` semantics are unchanged — the same \`TableCell\` reference still occupies multiple grid positions for spans. Only the serializers' use of it changes.
- HTML serializer wasn't modified; HTML supports native \`colspan\` / \`rowspan\` so the duplication issue may not apply there. Please flag if it does.

### DIFF
--- a/docling_core/transforms/serializer/markdown.py
+++ b/docling_core/transforms/serializer/markdown.py
@@ -532,9 +532,20 @@ class MarkdownTableSerializer(BaseTableSerializer):
                     res_parts.append(ann_res)
 
             rows = []
-            for row in item.data.grid:
+            for r, row in enumerate(item.data.grid):
                 rendered_row = []
-                for col in row:
+                for c, col in enumerate(row):
+                    # Spanning cells (col_span > 1 or row_span > 1) appear at
+                    # every grid position they cover. Emit cell text only at
+                    # the cell's origin position; spanned positions render as
+                    # empty so a colspan=4 cell isn't repeated 4 times.
+                    is_origin = (
+                        r == col.start_row_offset_idx
+                        and c == col.start_col_offset_idx
+                    )
+                    if not is_origin:
+                        rendered_row.append("")
+                        continue
                     if isinstance(col, RichTableCell):
                         ref_item = col.ref.resolve(doc=doc)
                         inner_kwargs = {**kwargs, "_nested_in_table": True}
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -2233,19 +2233,39 @@ class TableItem(FloatingItem):
             else:
                 break
 
-        # Create the column names from all col_headers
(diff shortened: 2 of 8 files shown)


ceberam:
@scottf007 Thanks for contributing to Docling

The Markdown serialization flattens the spanned cells by design. The Markdown exporter cannot preserve the spans since Markdown tables have no native span syntax. To better support downstream applications like RAG, we decided to repeat the values in all the spanned / merged cells. So the behavior is intentional and has been actively refined in previous PRs.

Please, check this documentation section for more details:
https://docling-project.github.io/docling/concepts/serialization/#table-cell-spans

PR #602, opened 2026-05-05
fix(serializer): collapse multi-row column headers into one markdown header

scottf007:
## Summary

`MarkdownTableSerializer` hardcoded `headers=rows[0]` and `rows[1:]`. When TableFormer correctly marks multiple leading rows as `column_header=True` (e.g. \"Cash per Security\" + \"(\$)\" on successive grid rows), only the first rendered as the markdown header and the continuation leaked into the body as a spurious data row.

Mirrors `_export_to_dataframe_with_options` (document.py:2219-2245): count leading grid rows containing any `column_header` cell, concatenate per column for the header, use the rest as body. Spanning siblings render empty so a `colspan=N` header isn't concatenated N times.

Falls back to \"first row is header\" when no `column_header` cells are marked — preserves prior behaviour for that path.

## Empirical result

10-PDF corpus, byte-comparison of rendered markdown:

- **IOZ_2025_04_17.pdf**: 2-row header collapses into 1. Body unchanged.
- **9 other fixtures**: byte-identical.

Related: docling-project/docling#2985.

### DIFF
--- a/docling_core/transforms/serializer/markdown.py
+++ b/docling_core/transforms/serializer/markdown.py
@@ -549,18 +549,53 @@ class MarkdownTableSerializer(BaseTableSerializer):
                     rendered_row.append(cell_text.replace("\n", " ").replace("|", "&#124;"))
                 rows.append(rendered_row)
             if len(rows) > 0:
+                # Count leading rows that contain any column_header cell so
+                # multi-line wrapped headers ("Cash per Security" + "($)" on
+                # successive grid rows) collapse into one markdown header
+                # row instead of leaking the continuation into the body.
+                # Mirrors `_export_to_dataframe_with_options` in
+                # docling_core/types/doc/document.py:2219-2245. Falls back to
+                # "first row is header" if upstream did not mark any header
+                # cells, preserving prior behaviour for that case.
+                num_header_rows = 0
+                for grid_row in item.data.grid:
+                    if any(
+                        getattr(cell, "column_header", False) for cell in grid_row
+                    ):
+                        num_header_rows += 1
+                    else:
+                        break
+                if num_header_rows == 0:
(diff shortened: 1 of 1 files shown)


cau-git:
@scottf007 Thanks for this proposal, having headers correctly separated from table bodies would clearly be an improvement. However, this behaviour is exactly intended in the default serializer: "Spanning siblings render empty so a colspan=N header isn't concatenated N times." so we would like to keep it. Do you want to update the PR with this?

scottf007:
Thanks @cau-git — appreciate you taking a look.

Quick clarification so I update the right thing: the PR already preserves the "spanning siblings render empty so a `colspan=N` header isn't concatenated N times" behaviour — that line in the description is describing what the patch keeps, not something it changes. The only behavioural change is that **multiple leading rows marked `column_header=True` now collapse into a single markdown header row** instead of the continuation rows leaking into the table body (e.g. a column title that wraps onto two visual lines like "Cash per Security" + "($)"). On the 10-PDF corpus only that one fixture changed; the other 9 were byte-identical.

So I'm not sure which part you'd like me to keep vs. change — happy to adjust, just want to make sure I understand. Is the concern that the multi-row collapse itself shouldn't be the default, or something about how the concatenation is done?

Separately: I've fixed the DCO check (the previous remediation commit referenced the wrong SHA).

Some broader feedback while I'm here, in case it's useful: the cases that bit me most were multi-column tables with sparse/empty columns where cells lose their x-position alignment, content occasionally getting dropped from the parse entirely, and flaky header detection. Alignment guards (inter-line/word/column distance heuristics) as a sanity check over the neural output might help — I know that's messy across hundreds of table shapes. Either way I think docling's a great project and want to help where I can.

PR #608, opened 2026-05-11
feat: add DOCLINGCORE_ALLOWED_PRIVATE_IPS setting to allow private IPs

### LINKED ISSUE #599: Allow configurable IP allowlist for _is_safe_url() to support private network access
## Problem

The `_is_safe_url()` function in `docling_core/utils/file.py` (introduced in PR #591) blocks all private, loopback, and link-local IP addresses as SSRF protection. While this is a sensible default for public-facing deployments, it breaks legitimate use cases where documents are fetched from internal/private network hosts, including:

- Local development and testing (`http://localhost`, `http://127.0.0.1`)
- Corporate intranet document servers (`http://192.168.x.x`, `http://10.x.x.x`)
- Container-to-container communication in Docker/Kubernetes
- On-premise document repositories

There is currently no way to opt out of or customize this behavior.

## Potential Solution

Add an `allowed_private_ips` setting to `CoreSettings` in `docling_core/utils/settings.py` that accepts a list of IP addresses and/or CIDR ranges. These would bypass the SSRF check in `_is_safe_url()`, while keeping the default behavior secure (empty list = full SSRF protection).

**Configuration via environment variable:**

```bash
DOCLINGCORE_ALLOWED_PRIVATE_IPS='["192.168.1.0/24", "10.0.0.5", "127.0.0.1"]'
```

**Implementation outline:**

- Add `allowed_private_ips: list[str] = []` to `CoreSettings`
- In `_is_safe_url()`, check the resolved IP against the allowlist before applying the default block — if the IP falls within any allowed network/address, return `True`
- Support both individual IPs and CIDR notation via `ipaddress.ip_network(..., strict=False)`

This follows the existing `CoreSettings` / `pydantic-settings` pattern already used for `allow_image_file_uri` and `max_image_decoded_size`.

## Defaults

The default remains secure — an empty allowlist means the current SSRF protection is unchanged. Users must explicitly opt in to allowing private IPs.

### LINKED ISSUE #600: `docling-core` rejects Azure Blob SAS URLs when the storage account resolves through a Private Endpoint
## Summary

After updating to `docling-core==2.74.1` (via `docling==2.89.0`), Docling rejects Azure Blob SAS URLs before attempting the download when the Blob hostname resolves to a private IP through Azure Private Link.

This appears to come from the URL safety validation in `docling_core.utils.file.resolve_source_to_stream()`, which rejects any URL whose hostname resolves to a non-global IP.

## Environment

- `docling==2.89.0`
- `docling-core==2.74.1`
- Python `3.13`
- Windows `11` locally
- Azure Blob Storage SAS URL
- Storage account uses Azure Private Endpoint / `privatelink.blob.core.windows.net`

## Actual behavior

Calling `DocumentConverter.convert()` with a valid Azure Blob SAS URL fails with:

```text
ValueError: URL is not allowed: https://<storage-account>.blob.core.windows.net/<container>/<document>.pdf?se=<expiry>&sp=r&sv=<api-version>&sr=b&sig=<redacted>
```

## Expected behavior

Docling should allow the URL, or provide a supported opt-out / allowlist mechanism for trusted private hosts used behind Azure Private Endpoints.

The SAS URL is valid and intentionally points to a private Azure Storage endpoint. From the application point of view, this is a legitimate remote document source.

## Minimal reproduction

```python

converter = DocumentConverter()
converter.convert(
    "https://<storage-account>.blob.core.windows.net/<container>/<document>.pdf?se=<expiry>&sp=r&sv=<api-version>&sr=b&sig=<redacted>"
)
```

This fails inside `docling_core.utils.file.resolve_source_to_stream()` before any fetch is attempted.

## DNS evidence

In our environment, the Azure Blob host resolves like this:

```text
<storage-account>.blob.core.windows.net             CNAME
<storage-account>.privatelink.blob.core.windows.net A 10.x.x.x
```

And a minimal check shows the resolved IP is private:

```python
import socket
import ipaddress
from urllib.parse import urlparse

url = "https://<storage-account>.blob.core.windows.net/<container>/<document>.pdf?se=<expiry>&sp=r&sv=<api-version>&sr=b&sig=<redacted>"
host = urlparse(url).hostname
ip = socket.gethostbyname(host)

print(host)  # <storage-account>.blob.core.windows.net
print(ip)    # 10.x.x.x
print(ipaddress.ip_address(ip).is_private)  # True
```

## Relevant code path

The rejection appears to come from this logic:

```python
http_url = TypeAdapter(AnyHttpUrl).validate_python(source)
url_str = str(http_url)

if not _is_safe_url(url_str):
    raise ValueError(f"URL is not allowed: {url_str}")
```

`_is_safe_url()` resolves the hostname and requires the resulting IP to be globally routable, which blocks Azure Storage accounts accessed through Private Link.

## Why this is a problem

This makes Docling incompatible with a common enterprise Azure setup where Blob Storage is intentionally reachable only through private networking.

The URL is not untrusted in our case; it is generated by our application and protected with a SAS token. The current behavior blocks legitimate private infrastructure along with SSRF-style targets.

## Questions

1. Is this strict private-IP rejection intentional for all remote sources?
2. Would you accept a configuration-based override, allowlist, or callback for trusted hosts?

## Full stack trace

```text
  File ".../app/messaging/consumer.py", line 159, in _process_message
    await self.message_handler(message_value)
  File ".../entrypoints/conversion_worker/lifespan.py", line 24, in _handle_message
    await asyncio.get_running_loop().run_in_executor(None, process_file_markdown, event.data.model_dump())
  File ".../Lib/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
  File ".../.venv/Lib/site-packages/sentry_sdk/integrations/threading.py", line 179, in wrapped_fn
    return fn(*args, **kwargs)
  File ".../app/services/conversion.py", line 99, in process_file_markdown
    markdown_content = convert_to_markdown_isolated(file_path=url, fast=fast)
  File ".../entrypoints/conversion_worker/docling_isolated.py", line 359, in convert_to_markdown_isolated
    return _get_docling_runtime_manager().convert_to_markdown(file_path=file_path, fast=fast)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../entrypoints/conversion_worker/docling_isolated.py", line 303, in convert_to_markdown
    markdown_content = client.convert_to_markdown(file_path=file_path, fast=fast)
  File ".../app/clients/docling.py", line 396, in convert_to_markdown
    result = converter.convert(file_path)
  File ".../.venv/Lib/site-packages/pydantic/_internal/_validate_call.py", line 40, in wrapper_function
    return wrapper(*args, **kwargs)
  File ".../.venv/Lib/site-packages/pydantic/_internal/_validate_call.py", line 137, in __call__
    res = self.__pydantic_validator__.validate_python(pydantic_core.ArgsKwargs(args, kwargs))
  File ".../.venv/Lib/site-packages/docling/document_converter.py", line 388, in convert
    return next(all_res)
  File ".../.venv/Lib/site-packages/docling/document_converter.py", line 447, in convert_all
    for conv_res in conv_res_iter:
                    ^^^^^^^^^^^^^
  File ".../.venv/Lib/site-packages/docling/document_converter.py", line 542, in _convert
    for input_batch in chunkify(
                       ~~~~~~~~^
        conv_input.docs(self.format_to_options),
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        settings.perf.doc_batch_size,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ):
    ^
  File ".../.venv/Lib/site-packages/docling/utils/utils.py", line 15, in chunkify
    for first in iterator:
                 ^^^^^^^^
  File ".../.venv/Lib/site-packages/docling/datamodel/document.py", line 452, in docs
    resolve_source_to_stream(item, self.headers)
  File ".../.venv/Lib/site-packages/docling_core/utils/file.py", line 119, in resolve_source_to_stream
    raise ValueError(f"URL is not allowed: {url_str}")
ValueError: URL is not allowed: https://<storage-account>.blob.core.windows.net/<container>/<document>.pdf?se=<expiry>&sp=r&sv=<api-version>&sr=b&sig=<redacted>
```

COTHSC:
Closes #599

Adds an optional "allowed_private_ips" to CoreSettings to specify safe internal IPs as described in the issue
Default behaviour is unchanged

### DIFF
--- a/docling_core/utils/file.py
+++ b/docling_core/utils/file.py
@@ -1,6 +1,7 @@
 """File-related utilities."""
 
 import ipaddress
+import logging
 import re
 import tempfile
 from io import BytesIO
@@ -14,6 +15,9 @@ from typing_extensions import deprecated
 
 from docling_core.types.doc.utils import relative_path
 from docling_core.types.io import DocumentStream
+from docling_core.utils.settings import settings
+
+_logger = logging.getLogger(__name__)
 
 _MAX_REDIRECTS = 5
 
@@ -31,6 +35,20 @@ class FileSizeLimitExceededError(ValueError):
         self.size = size
         self.limit = limit
         super().__init__(f"Remote file exceeds the maximum allowed size ({size} > {limit} bytes).")
+
+
+def _ip_in_allowlist(ip: ipaddress.IPv4Address, allowlist: list[str]) -> bool:
+    for entry in allowlist:
+        entry = entry.strip()
+        if not entry:
+            continue
+        try:
+            network = ipaddress.ip_network(entry)
+            if ip in network:
+                return True
+        except ValueError:
+            _logger.warning(f"Skipping malformed entry in DOCLINGCORE_ALLOWED_PRIVATE_IPS: {entry!r}")
+    return False
 
 
 def _is_safe_url(url: str) -> bool:
@@ -52,6 +70,8 @@ def _is_safe_url(url: str) -> bool:
                 ip = ipaddress.ip_address(ip_str)
             except (socket.gaierror, socket.herror):
                 return False
+        if settings.allowed_private_ips and _ip_in_allowlist(ip, settings.allowed_private_ips):
+            return True
 
         return ip.is_global and not (
             ip.is_private
--- a/docling_core/utils/file.py
+++ b/docling_core/utils/file.py
@@ -1,6 +1,7 @@
 """File-related utilities."""
 
 import ipaddress
+import logging
 import re
 import tempfile
 from io import BytesIO
@@ -14,8 +15,25 @@ from typing_extensions import deprecated
 
 from docling_core.types.doc.utils import relative_path
 from docling_core.types.io import DocumentStream
+from docling_core.utils.settings import settings
+
+_logger = logging.getLogger(__name__)
 
 _MAX_REDIRECTS = 5
+
+
+def _ip_in_allowlist(ip: ipaddress.IPv4Address, allowlist: list[str]) -> bool:
+    for entry in allowlist:
+        entry = entry.strip()
+        if not entry:
+            continue
+        try:
+            network = ipaddress.ip_network(entry)
+            if ip in network:
+                return True
+        except ValueError:
+            _logger.warning(f"Skipping malformed entry in DOCLINGCORE_ALLOWED_PRIVATE_IPS: {entry!r}")
+    return False
 
 
 def _is_safe_url(url: str) -> bool:
@@ -37,6 +55,8 @@ def _is_safe_url(url: str) -> bool:
                 ip = ipaddress.ip_address(ip_str)
             except (socket.gaierror, socket.herror):
                 return False
+        if settings.allowed_private_ips and _ip_in_allowlist(ip, settings.allowed_private_ips):
+            return True
 
         return ip.is_global and not (
             ip.is_private
--- a/docling_core/utils/file.py
+++ b/docling_core/utils/file.py
@@ -37,7 +37,8 @@ class FileSizeLimitExceededError(ValueError):
         super().__init__(f"Remote file exceeds the maximum allowed size ({size} > {limit} bytes).")
 
 
-def _ip_in_allowlist(ip: ipaddress.IPv4Address, allowlist: list[str]) -> bool:
+def _ip_in_allowlist(ip: ipaddress.IPv4Address | ipaddress.IPv6Address, allowlist: list[str]) -> bool:
+    """Return whether a ip matches any IP addresss or CIDR entry in allowlist."""
     for entry in allowlist:
         entry = entry.strip()
         if not entry:
@@ -47,7 +48,11 @@ def _ip_in_allowlist(ip: ipaddress.IPv4Address, allowlist: list[str]) -> bool:
             if ip in network:
                 return True
         except ValueError:
-            _logger.warning(f"Skipping malformed entry in DOCLINGCORE_ALLOWED_PRIVATE_IPS: {entry!r}")
+            _logger.warning(
+                f"Skipping malformed entry in DOCLINGCORE_ALLOWED_PRIVATE_IPS: {entry!r} "
+                "(if this is a CIDR range, the network address must have no host bits set, "
+                "e.g. '10.0.0.0/8' rather than '10.0.0.5/8')"
+            )
     return False
 
 
--- a/docling_core/utils/file.py
+++ b/docling_core/utils/file.py
@@ -44,14 +44,18 @@ def _ip_in_allowlist(ip: ipaddress.IPv4Address | ipaddress.IPv6Address, allowlis
         if not entry:
             continue
         try:
-            network = ipaddress.ip_network(entry)
+            network = ipaddress.ip_network(entry, strict=False)
+            normalized = str(network)
+            if normalized != entry:
+                _logger.warning(
+                    f"DOCLINGCORE_ALLOWED_PRIVATE_IPS entry {entry!r} was normalized to "
+                    f"{normalized!r}. Consider using the explicit network address."
+                )
             if ip in network:
                 return True
         except ValueError:
             _logger.warning(
                 f"Skipping malformed entry in DOCLINGCORE_ALLOWED_PRIVATE_IPS: {entry!r} "
-                "(if this is a CIDR range, the network address must have no host bits set, "
-                "e.g. '10.0.0.0/8' rather than '10.0.0.5/8')"
             )
     return False
 
--- a/docling_core/utils/file.py
+++ b/docling_core/utils/file.py
@@ -54,9 +54,7 @@ def _ip_in_allowlist(ip: ipaddress.IPv4Address | ipaddress.IPv6Address, allowlis
             if ip in network:
                 return True
         except ValueError:
-            _logger.warning(
-                f"Skipping malformed entry in DOCLINGCORE_ALLOWED_PRIVATE_IPS: {entry!r} "
-            )
+            _logger.warning(f"Skipping malformed entry in DOCLINGCORE_ALLOWED_PRIVATE_IPS: {entry!r} ")
     return False
 
 
--- a/docling_core/utils/settings.py
+++ b/docling_core/utils/settings.py
@@ -6,6 +6,7 @@ class CoreSettings(BaseSettings):
 
     allow_image_file_uri: bool = False
     max_image_decoded_size: int = 20 * 1024 * 1024  # 20MB
+    allowed_private_ips: list[str] = []
 
     # DocLang deserialize budgets (DoS protection for untrusted markup / .dclx)
     max_doclang_xml_bytes: int = 128 * 1024 * 1024  # 128 MiB
--- a/docling_core/utils/settings.py
+++ b/docling_core/utils/settings.py
@@ -6,6 +6,7 @@ class CoreSettings(BaseSettings):
 
     allow_image_file_uri: bool = False
     max_image_decoded_size: int = 20 * 1024 * 1024  # 20MB
+    allowed_private_ips: list[str] = []
 
 
 settings = CoreSettings()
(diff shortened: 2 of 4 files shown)


JViktoRArtola:
Is there a planned release window for this? I’d be happy to help test or contribute if that would be useful.

darkms:
Underlying issue is currently blocking co-hosting docling and rustfs (for use with /source APIs) on same docker-compose/machine. Would love to see this merged.

Yehonatanbh:
This is blocking us too — container-to-container fetches over a private address,
which is the third case listed in #599.
@dolfim-ibm — is there an open design question, or is this just waiting on review
bandwidth?

MrSexyCode:
Adding another voice on this one. It is blocking private-network `/source` fetches for us as well.

@PeterStaar-IBM would you be able to assign a second reviewer so the two-reviewer protection can actually be satisfied? Right now the PR cannot merge even if the requested review comes back approving.

cau-git (dismissed):
lgtm

docling_core/utils/file.py
```diff
@@ -14,10 +15,27 @@
...
 _MAX_REDIRECTS = 5
 
 
+def _ip_in_allowlist(ip: ipaddress.IPv4Address, allowlist: list[str]) -> bool:
+    for entry in allowlist:
+        entry = entry.strip()
+        if not entry:
+            continue
+        try:
+            network = ipaddress.ip_network(entry)
```
ceberam: @fpertl explicitly names `strict=False` as part of the proposed `ip_network(...)` call, but the PR uses the default `strict=True`. Was this intentional? A CIDR with host bits set (e.g. 10.0.0.5/8) is a valid user intent, not a malformed entry.

```suggestion
            network = ipaddress.ip_network(entry, strict=False)
```

test/test_utils.py
```diff
@@ -192,7 +192,54 @@ def test_is_safe_url_rejects_private_networks():
...
+    import pytest
+    from docling_core.utils.file import resolve_source_to_stream
+    from docling_core.utils.settings import settings
+    from requests import Response
+
+    def mock_get(*args, **kwargs):
+        r = Response()
+        r.status_code = 200
+        r._content = b"ok"
+        return r
```
ceberam: The test fails with errors: `mock_get` constructs a bare `Response()` without setting `raw` or `_content_consumed`, which causes an `AttributeError: 'NoneType' object has no attribute 'read'`.
We already have a helper `_closeable_response()` precisely for this reason. Please, check how it's been used in the other tests of this same module.

docling_core/utils/settings.py
```diff
@@ -6,6 +6,7 @@ class CoreSettings(BaseSettings):
 
     allow_image_file_uri: bool = False
     max_image_decoded_size: int = 20 * 1024 * 1024  # 20MB
+    allowed_private_ips: list[str] = []
```
ceberam: We don't have proper documentation for the other variables, but since you create a new one, it would be a good practice to include it 🙂 
We prefer the `Annotated` pattern and pydantic's `Field` function for pydantic models. What about something like:


```suggestion
    allowed_private_ips: Annotated[
        list[str],
        Field(
            description=(
                "List of IP addresses and/or CIDR ranges that bypass SSRF protection. "
                "Defaults to empty, which preserves full SSRF protection."
            ),
            examples=[
                ["192.168.1.0/24", "10.0.0.5", "127.0.0.1"],
            ],
        ),
    ] = []
```

docling_core/utils/file.py
```diff
@@ -14,10 +15,27 @@
...
 from docling_core.types.doc.utils import relative_path
 from docling_core.types.io import DocumentStream
+from docling_core.utils.settings import settings
+
+_logger = logging.getLogger(__name__)
 
 _MAX_REDIRECTS = 5
 
 
+def _ip_in_allowlist(ip: ipaddress.IPv4Address, allowlist: list[str]) -> bool:
```
ceberam: - The type hint of `ip` is `ipaddress.IPv4Address` but the function may receive `IPv6Address` objects too (from `_is_safe_url`). Even though at runtime this won't crash, it would be good to set the type properly: either `ipaddress.IPv4Address | ipaddress.IPv6Address` or `ipaddress.ip_address` (the union type, preferred)
- Minor comment: we don't require docstring on private functions, but since all other private methods include them, it would be good to have it here as well.

test/test_utils.py
```diff
@@ -192,7 +192,54 @@ def test_is_safe_url_rejects_private_networks():
...
+    from docling_core.utils.file import _ip_in_allowlist
 
+    assert _ip_in_allowlist(ipaddress.ip_address("127.0.0.1"), ["127.0.0.1"])
+    assert not _ip_in_allowlist(ipaddress.ip_address("127.0.0.1"), ["127.0.0.2"])
+
+    assert _ip_in_allowlist(ipaddress.ip_address("10.20.30.40"), ["10.0.0.0/8"])
+    assert not _ip_in_allowlist(ipaddress.ip_address("10.21.0.1"), ["10.20.0.0/16"])
+
+    assert _ip_in_allowlist(ipaddress.ip_address("192.168.1.100"), ["192.168.1.0/24"])
+    assert not _ip_in_allowlist(ipaddress.ip_address("192.168.1.100"), ["192.168.1.5/24"])
```
ceberam: FYI, if we `strict=False`, this test would need to be rewritten

ceberam (changes requested):
Thanks @COTHSC for suggesting this PR!

Please, have a look at the comments below.

docling_core/utils/file.py
COTHSC: I honestly cannot remember about this, I will have a look at it again asap

COTHSC: Yes this was intentional, My logic was that 10.0.0.5/8 could be ambiguous, and a misunderstanding here could lead to the policy being much more permissive than the developer intends. I don't know the case for 10.0.0.5/8 over 10.0.0.0/8,  I thought the latter was more explicit, and that this may be good for an allowlist with implications for security.
I could be wrong and am happy to make the change if that is preferable.

ceberam: That makes total sense, and I completely agree that being explicit is the golden rule for security configurations. In a static backend config file, `10.0.0.0/8` is absolutely the right way to write it. The reason I’m advocating for `strict=False` here is because of how this interacts with user-supplied input in the real world, where we face a different kind of risk:
- if a user/developer types 10.0.0.5/24 intending to allow their local network, this PR code logs a warning in the background but completely ignores the line. To the user, it looks like Docling added the allowlist rule, but their traffic will be blocked. This usually leads to frustrating debugging sessions or emergency support tickets.
- standard networking tools (`nginx`, `nmap`, cloud firewalls like AWS Security Groups) have standardized on the `strict=False` approach. They treat the CIDR slash as a masking instruction rather than a mathematical assertion. Because users are conditioned by these tools, they often copy-paste their machine's IP with a subnet mask (e.g., 192.168.1.45/24) expecting it to work.
- ambiguity vs security: while `10.0.0.5/8` looks ambiguous, mathematically `ipaddress` handles it deterministically. It masks it to `10.0.0.0/8` every single time.

If we want the best of both worlds, preventing user mistakes while keeping the configuration explicit, we could keep `strict=False` so the application doesn't break, but log a specific, actionable warning telling the user that their entry was normalized. What do you think?

COTHSC: Hahaha, I came to the mirror conclusion, keeping strict=true but making the log more explicit to explain the behavior.

I think you are right, matching with standard networking tools ultimately makes the most sense, I will implement strict=False + a specific log warning.

ceberam (approved):
Thanks @COTHSC for the latest changes and your contribution!

Since we recently renamed the test folder, there was a conflict, which I easily resolved by rebasing to `main` and force-pushing your branch.
The PR is ready to merge 🚀

PeterStaar-IBM (approved):
lgtm!

PR #613, opened 2026-05-14
fix(html-serializer): emit XHTML-compatible metadata attributes

### LINKED ISSUE #598: `HTMLMetaSerializer` emits non-XHTML metadata attributes
`docling-core` HTML serialization can emit invalid XHTML when serializing item metadata. The generic metadata serializer currently produces valueless `data-meta-*` attributes, for example:

```html
<div data-meta-classification>Other</div>
```

This is accepted by lenient HTML parsers, but it is not XML/XHTML-compliant. Downstream consumers that parse exported table HTML as XML, such as TEDS table evaluation, fail with errors like:

```text
not well-formed (invalid token)
```

Minimal reproducer:

```python
from xml.etree import ElementTree as ET

from docling_core.transforms.serializer.html import HTMLDocSerializer
from docling_core.types.doc.document import (
    PictureClassificationMetaField,
    PictureClassificationPrediction,
    PictureMeta,
    RichTableCell,
    TableCell,
    TableData,
)

doc = DoclingDocument(name="x")

plain = TableCell(
    start_row_offset_idx=0,
    end_row_offset_idx=1,
    start_col_offset_idx=0,
    end_col_offset_idx=1,
    text="",
)
table = doc.add_table(data=TableData(num_rows=1, num_cols=1, table_cells=[plain]))

pic = doc.add_picture(parent=table)
pic.meta = PictureMeta(
    classification=PictureClassificationMetaField(
        predictions=[PictureClassificationPrediction(class_name="other", confidence=1.0)]
    )
)

table.data.table_cells = [
    RichTableCell(
        start_row_offset_idx=0,
        end_row_offset_idx=1,
        start_col_offset_idx=0,
        end_col_offset_idx=1,
        text="",
        ref=pic.get_ref(),
    )
]

html = table.export_to_html(doc)
print(html)

ET.fromstring(html)  # ParseError: not well-formed
```

Actual output:

```html
<table><tbody><tr><td><figure><details><summary>Meta</summary><div data-meta-classification>Other</div></details></figure></td></tr></tbody></table>
```

Expected: exported HTML should be XHTML-compatible, or at least metadata serialization should not emit valueless attributes. A safer shape would be something like:

```html
<div data-meta-name="classification">Other</div>
```

with metadata names and text escaped correctly. `TabularChartMetaField` likely needs a separate branch because its value can intentionally be HTML table markup.

BrianWang1990:
## Summary

Closes #598.

`HTMLMetaSerializer` currently renders each metadata field as a valueless HTML5 attribute, e.g.

```html
<div data-meta-classification>Other</div>
```

Lenient HTML parsers accept this, but the output is **not valid XHTML/XML**. Downstream consumers that parse exported HTML as XML — including TEDS table evaluation — fail with:

```text
xml.etree.ElementTree.ParseError: not well-formed (invalid token)
```

This PR switches the attribute shape to a standard name=value pair, matching the resolution suggested in the issue:

```html
<div data-meta-name="classification">Other</div>
```

## Changes

- **`docling_core/transforms/serializer/html.py`** — `HTMLMetaSerializer._serialize_meta_field` now emits `data-meta-name="<field>"` and HTML-escapes both the field name and the rendered value. `TabularChartMetaField` is intentionally left unescaped because its value is an embedded HTML table (consistent with the existing `CodeMetaField` branch which also emits raw HTML).
- **`docling_core/transforms/serializer/html_styles.py`** — Updated the two `[data-meta-classification]` CSS selectors to `[data-meta-name="classification"]` so existing styling keeps working.
- **`test/data/doc/*.gt.html`** — Regenerated 22 ground-truth HTML fixtures via `DOCLING_GEN_TEST_DATA=1 pytest`. The diffs are limited to the attribute rename; no other content changed.
- **`test/test_serialization.py`** — Added `test_html_meta_emits_xhtml_compatible_attributes` which builds the minimal reproducer from the issue and asserts the output round-trips through `xml.etree.ElementTree.fromstring` (i.e. is well-formed XML).

## Verification

```text
$ uv run pytest test/
444 passed, 2 xfailed in 77.96s
```

Reproducer from the issue, before vs after:

```python
from xml.etree import ElementTree as ET
# (setup omitted, see issue #598)
ET.fromstring(html)
```

Before: `ParseError: not well-formed (invalid token)`
After: parses cleanly.

## Notes

- This is technically a breaking change for any external consumer that selects metadata via the old bare attribute selector `[data-meta-<name>]`. The new attribute (`data-meta-name="<name>"`) is the suggested replacement and the only such consumer in this repo (the bundled CSS) has been updated. No documented public API contract is broken.
- Linting: `uv run ruff check` and `uv run ruff format --check` both pass on the modified files.


ceberam:
Thanks @BrianWang1990 for suggesting this fix!
When reviewing your PR, I've just realized that in today's release we have extended the issue with yet other valueless HTML5 attributes: `data-meta-entities` and `data-meta-language`.

We may produce non-valid XHTML like:

```html
<details class="docling-meta">
  <summary>Meta</summary>
  <div class="docling-meta-field" data-meta-summary>
    <span class="docling-meta-field-label">summary:</span>
    <span class="docling-meta-field-value">A short company/location statement.</span>
  </div>
  <div class="docling-meta-field" data-meta-language>
    <span class="docling-meta-field-label">language:</span>
    <span class="docling-meta-field-value">en</span>
  </div>
  <div class="docling-meta-field" data-meta-entities>
    <span class="docling-meta-field-label">entities:</span>
    <span class="docling-meta-field-value">IBM@0-3 (ORG), Zurich@16-22 (LOC)</span>
  </div>
</details>
```

Could you please rebase this PR to `main`, check the new [_serialize_meta_field](https://github.com/docling-project/docling-core/blob/main/docling_core/transforms/serializer/html.py#L946) function, and apply the fix for the language and entity meta fields?

ceberam:
@BrianWang1990 to accelerate the resolution of the issue, I have rebased your branch to `main` and I added a couple of commits:

- Fix Double HTML Escaping: rebasing on the current `main` branch required to solve a double HTML escaping. Metadata field values were being HTML-escaped twice in HTMLMetaSerializer._serialize_meta_field(), causing text like "A<B & C>" to render as "A&amp;lt;B &amp;amp; C&amp;gt;" instead of the correct "A&lt;B &amp; C&gt;"
- Remove Redundant CSS Rules
  - Removed duplicate CSS rule `.docling-meta-field[data-meta-name="classification"] { text-align: left; }` from both `_get_css_for_split_page()` and `_get_css_for_single_column()` in `html_styles.py`
  - This rule was redundant since `.docling-meta-field` already defines `text-align: left;`
  - Updated all ground-truth HTML test files accordingly

ceberam (approved):
🚀

dolfim-ibm (approved):

PR #614, opened 2026-05-14
fix(markdown): pass kwargs to _serialize_image_part()

jspast:
This allows passing custom arguments to the `_serialize_image_part()` method.

Then, with a custom serializer, it is possible to specify which image ids to serialize just overriding this method, similar to the [Indexed picture placeholders](https://docling-project.github.io/docling/examples/serialization/#indexed-picture-placeholders) example.

### DIFF
--- a/docling_core/transforms/serializer/markdown.py
+++ b/docling_core/transforms/serializer/markdown.py
@@ -614,8 +614,7 @@ class MarkdownPictureSerializer(BasePictureSerializer):
             img_res = self._serialize_image_part(
                 item=item,
                 doc=doc,
-                image_mode=params.image_mode,
-                image_placeholder=params.image_placeholder,
+                **kwargs,
             )
             if img_res.text:
                 res_parts.append(img_res)


ceberam (commented):
Thanks @jspast for contributing to Docling!

Could you be a bit more explicit with the use cases that you want to achieve with the proposed PR? 

If the goal is to filter which images to serialize, I see some limited benefits in this PR, plus some challenges:
- The change provides minimal additional flexibility since custom parameters would need to be added to `MarkdownParams` anyway, the `_serialize_image_part()` signature still requires explicit parameters, and subclasses can already override either method (see examples below)
- Inconsistency with other serializers: `LateXPictureSerializer` explicitly passes only `image_mode` and `image_placeholders` to its `_serializer_impage_part()`, not `**kwargs`.
- Passing `**kwargs` when only 2 specific parameters are used may obscure the method's actual requirements

You can already achieve the filtering by:

- overriding the entire `serialize()` method:

```python
class CustomMarkdownPictureSerializer(MarkdownPictureSerializer):
    def serialize(self, *, item: PictureItem, doc_serializer, doc, **kwargs):
        # Custom logic to filter images
        if should_skip_image(item):
            return SerializationResult(text="")
        return super().serialize(item=item, doc_serializer=doc_serializer, doc=doc, **kwargs)
```

- overriding `_serialize_image_part()` directly (like in the example of the documentation that you referenced):

```python
class CustomMarkdownPictureSerializer(MarkdownPictureSerializer):
    def _serialize_image_part(self, item, doc, image_mode, image_placeholder, **kwargs):
        # Custom logic here
        if should_skip_image(item):
            return SerializationResult(text="")
        return super()._serialize_image_part(item, doc, image_mode, image_placeholder, **kwargs)
```

Would those alternatives be satisfactory?

jspast:
Thanks for your feedback @ceberam! I am sorry for not describing my use case properly, but it is really just filtering images.

Your first approach is actually nice. I somehow did not notice it could be as simple.

Your second approach is basically what I am doing. The thing is that "`should_skip_image()`" needs both the item and some information to decide whether to filter it or not.

With this PR, this is possible:

```python
class CustomMarkdownPictureSerializer(MarkdownPictureSerializer):
    def _serialize_image_part(self, item, doc, image_mode, image_placeholder, **kwargs):
        included_index = kwargs.get("included_image_index")
        if should_skip_image(item, included_index):
            return SerializationResult(text="")
        return super()._serialize_image_part(item, doc, image_mode, image_placeholder, **kwargs)
```

Then:

```python
# included_image_index argument is used to omit other images
ser_result = serializer.serialize(included_image_index=pic_index)
```

My use case is really specific. Basically, doing picture description with textual context from the extracted document. After I compute a DoclingDocument without picture descriptions, I generate them by prompting a VLM with the image and the serialized page it was extracted from. In the prompt, I say that the image was extracted from where there is now a placeholder. That is why I need to filter images.

I do like your first approach, so feel free to close this PR. Thanks again for the feedback and all the work on Docling!

ceberam:
All good @jspast , I will then close this PR.
Please, feel free to share any other question, use case, contribution, or  feedback you may have on Docling ✌️

PR #621, opened 2026-05-26
fix: render RichTableCell descendants inside the cell

PeterStaar-IBM:
## Summary

  - `HTMLTextSerializer` only rendered an item's own text and a few special
    branches (Title/SectionHeader/Formula/Code/ListItem and `text=""` + single
    InlineGroup child). Other `item.children` were silently ignored at the
    serializer level — they were re-emitted by the outer document iteration. At
    the document level this looks fine, but inside a `RichTableCell` it produced
    a leak: descendants of the cell's ref rendered *outside* the table as
    standalone `<p>`s.
  - Fix: after wrapping the parent's text in its tag and applying
    `post_process`, recurse into `item.children` via
    `doc_serializer.get_parts(item=item, visited=my_visited, ...)` and append
    the rendered children as siblings (joined with `\n`). `ListItem` already
    recurses inline, and `has_inline_repr` already consumed its single child —
    both branches are skipped. Children render as proper sibling elements
    (never nested inside `<p>`), and `visited` propagation prevents the outer
    iteration from re-emitting them.
  - Doc-level output is byte-identical for existing fixtures (`rich_table.gt.html`,
    `inline_and_formatting.gt.html`, etc.) — children that previously rendered
    via the outer iteration now render via the parent's serialize call instead,
    in the same order.

  Follow-up work tracked in
  `_plans/fix-html-rich-tablecell-recursion.md`: apply the same recursion
  pattern to `MarkdownTextSerializer`, audit `latex.py` and `doctags.py` for
  the same gap, and consider lifting a shared helper into `common.py`.

## Test plan

  - [ ] `pytest test/test_serialization.py::test_html_rich_cell_textitem_ref_subtree_inside_and_not_outside`
  - [ ] `pytest test/test_serialization.py::test_html_textitem_with_children_at_document_level`
  - [ ] `pytest test/test_serialization.py test/test_serialization_doclang.py test/test_deserializer_doclang.py test/test_metadata.py
  test/test_latex_serialization.py test/test_serialization_doctag.py` — all 185 pass
  - [ ] Reproduce on `main`: the rich-cell test fails with `<p>DEEP-LEAK</p>`
    showing up after `</table>`; with this PR the deep child renders inside
    the cell instead

### DIFF
--- a/docling_core/transforms/serializer/html.py
+++ b/docling_core/transforms/serializer/html.py
@@ -230,13 +230,30 @@ class HTMLTextSerializer(BaseModel, BaseTextSerializer):
             # Regular text item
             text = get_html_tag_with_text_direction(html_tag="p", text=text)
 
-        # Apply formatting and hyperlinks
+        # Apply formatting and hyperlinks to the parent's own text+tag.
         if not post_processed:
             text = doc_serializer.post_process(
                 text=text,
                 formatting=item.formatting,
                 hyperlink=item.hyperlink,
             )
+
+        # Recurse into children for branches that don't already consume them.
+        # has_inline_repr already consumed the single InlineGroup child; ListItem
+        # recursion is handled inline in its branch above.
+        if not has_inline_repr and not isinstance(item, ListItem) and item.children:
+            nested_text = "\n".join(
+                r.text
+                for r in doc_serializer.get_parts(
+                    item=item,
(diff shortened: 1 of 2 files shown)


vagenas (approved):

maxmnemonic (approved):
LGTM

PR #632, opened 2026-06-09
fix: handle page-less documents in table doctags export

### LINKED ISSUE #90: export_to_document_tokens method of TableItem is not safe for page-less documents
We need to be careful when using referring page properties, as not every docling document will have pages (with sizes): [here](https://github.com/DS4SD/docling-core/blob/4e0cb3fcde6af3b940afcb2faea0ff9c447c3600/docling_core/types/doc/document.py#L1133)

sriharan0804:
## Summary

Fixes #90 by handling page-less documents during table DocTags export.

While investigating the issue, I found that `TableItem.export_to_otsl()` attempted to generate location tokens for table cells with bounding boxes even when the document contained no page metadata. In this situation, `page_w` and `page_h` were unavailable, resulting in:

```python
UnboundLocalError: cannot access local variable 'page_h'
```

## Changes

* Only generate cell location tokens when page information is available.
* Added a regression test covering a page-less document with a table cell that has a bounding box.

## Testing

Added:

* `test_table_export_to_otsl_page_less_document_with_cell_bbox`

Verified that:

* The test reproduces the issue before the fix.
* The test passes after the fix.
* Existing OTSL export tests continue to pass.

### DIFF
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -2590,10 +2590,11 @@ class TableItem(FloatingItem):
                     cell.start_col_offset_idx,
                 )
 
-                if len(doc.pages.keys()):
-                    page_w, page_h = doc.pages[page_no].size.as_tuple()
+                has_page_info = page_no in doc.pages
+
                 cell_loc = ""
-                if cell.bbox is not None:
+                if cell.bbox is not None and has_page_info:
+                    page_w, page_h = doc.pages[page_no].size.as_tuple()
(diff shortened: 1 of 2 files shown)


sriharan0804:
Hi! This is my first contribution to Docling Core. I reproduced the issue locally, added a regression test, and implemented a fix. Please let me know if any changes are needed. Thanks for your time and review!

cau-git (approved):

PeterStaar-IBM (approved):
lgtm!

PR #634, opened 2026-06-10
fix: externalize page images in REFERENCED mode for JSON/YAML export

cau-git:
When saving a `DoclingDocument` with `image_mode=ImageRefMode.REFERENCED`, only `PictureItem` images were written to the artifacts directory and referenced via file URIs. Page images (`pages[n].image`) were left untouched and remained as embedded base64 `data:` blobs in the serialized output — defeating the purpose of "referenced" mode and bloating the JSON.

### Change

- `_with_pictures_refs` gains an `include_page_images` flag. When set, it saves each page image to the artifacts dir as `page_{page_no:06}_{hexhash}.png` and rewrites `page.image.uri` to a (relative, where applicable) file path — mirroring the existing picture-ref logic, including the `page_no` filter.
- `_make_copy_with_refmode` threads the flag through.
- `save_as_json` and `save_as_yaml` pass `include_page_images=True`, since both are full-document dumps that exhibited the base64 bloat. Markdown/HTML are unchanged (page images aren't serialized into those outputs).

### Behavior

- `EMBEDDED` and `PLACEHOLDER` modes are unaffected.
- Save → load round-trips restore page images correctly.

### DIFF
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -5883,11 +5883,16 @@ class DoclingDocument(BaseModel):
         image_dir: Path,
         page_no: Optional[int],
         reference_path: Optional[Path] = None,
+        include_page_images: bool = False,
     ) -> "DoclingDocument":
         """Document with images as refs.
 
         Creates a copy of this document where all picture data is
         saved to image_dir and referenced through file URIs.
+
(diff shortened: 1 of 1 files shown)


dolfim-ibm (approved):
lgtm

vagenas (approved):

PR #635, opened 2026-06-10
feat: Make HTTP downloads respect file-size limits and abort early

cau-git:
### Summary

Make remote source downloads respect a caller-provided size limit before buffering the full response in memory.

### Problem

`resolve_source_to_stream()` previously fetched remote content with `res.content` and wrapped it in `BytesIO`, which meant the full HTTP response was materialized in memory before any downstream `max_file_size` limit could be enforced.

### Changes

- Added an optional `max_file_size` parameter to `resolve_source_to_stream()`.
- Added an upfront `Content-Length` check and reject the download immediately when the declared size exceeds the configured limit.
- Replaced eager `res.content` loading with chunked `iter_content()` reading.
- Enforced the size limit while streaming so oversized downloads are aborted even when `Content-Length` is missing or incorrect.
- Introduced `FileSizeLimitExceededError` to distinguish size-limit failures from other URL/input errors while preserving safe early termination.

### Resulting behavior

- Remote files with `Content-Length > max_file_size` are rejected before any body bytes are buffered.
- Remote files without a usable `Content-Length` are downloaded only up to the configured limit, then aborted as soon as the streamed byte count exceeds it.
- Accepted remote files are still returned as `DocumentStream(BytesIO(...))`; this change prevents unbounded buffering but does not make the full pipeline end-to-end streaming.

### DIFF
--- a/docling_core/utils/file.py
+++ b/docling_core/utils/file.py
@@ -17,6 +17,11 @@ from docling_core.types.io import DocumentStream
 
 _MAX_REDIRECTS = 5
 
+# Chunk size for streaming remote downloads. Sized for document payloads (PDFs
+# and similar), which are either small or in the multi-megabyte range, so a
+# large chunk keeps the read loop short without wasting memory.
+_DOWNLOAD_CHUNK_SIZE = 512 * 1024
+
 
 class FileSizeLimitExceededError(ValueError):
     """Raised when a remote file exceeds the configured download size limit."""
@@ -113,12 +118,16 @@ def resolve_source_to_stream(
     """Resolves the source (URL, path) of a file to a binary stream.
 
     Args:
-        source (Path | AnyHttpUrl | str): The file input source. Can be a path or URL.
-        headers (Optional[dict[str, str]]): Optional set of headers to use for fetching
-            the remote URL.
+        source: The file input source. Can be a path or URL.
+        headers: Optional set of headers to use for fetching the remote URL.
+        max_file_size: Optional maximum size, in bytes, for a remote download.
+            When set, the download is rejected upfront if the declared
+            ``Content-Length`` exceeds it, and aborted while streaming as soon as
+            the received bytes exceed it.
 
     Raises:
         ValueError: If source is of unexpected type.
+        FileSizeLimitExceededError: If a remote download exceeds ``max_file_size``.
 
     Returns:
         DocumentStream: The resolved file loaded as a stream.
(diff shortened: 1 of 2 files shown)


docling_core/utils/file.py
```diff
@@ -96,7 +106,9 @@ def resolve_remote_filename(
 
 
 def resolve_source_to_stream(
-    source: Union[Path, AnyHttpUrl, str], headers: Optional[dict[str, str]] = None
+    source: Union[Path, AnyHttpUrl, str],
+    headers: Optional[dict[str, str]] = None,
+    max_file_size: Optional[int] = None,
```
ceberam, line 116: - The new parameter `max_file_size` has been added but the docstring was not updated to document it.
- The function may raise `FileSizeLimitExceededError` now, so it should be added to the `Raises:` docstring section
- By the way, I would also remove the type descriptions from the docstring (e.g., _source ~(Path | AnyHttpUrl | str)~: The file input source. Can be a path or URL._) to avoid any eventual mismatch with the type annotations on the function signature. These ones are enough to automatically generate documentation.

docling_core/utils/file.py
```diff
@@ -166,21 +175,54 @@ def _check_redirect_safety(response, *args, **kwargs):
...
+                        if content_length_value is not None and content_length_value > max_file_size:
+                            raise FileSizeLimitExceededError(
+                                filename=fname,
+                                size=content_length_value,
+                                limit=max_file_size,
+                            )
+
+                stream = BytesIO()
+                downloaded = 0
+                for chunk in res.iter_content(chunk_size=8192):
```
ceberam: Just wondering: is it worth parametrizing the chunk size to give flexibility?

test/test_utils.py
```diff
@@ -350,3 +369,42 @@ def mock_get_with_redirect(self, *args, **kwargs):
 
     with pytest.raises(ValueError, match="Redirect target is not allowed"):
         resolve_source_to_stream("https://example.com/redirect")
+
+
+def test_resolve_source_rejects_large_content_length(monkeypatch):
+    import pytest
+    from requests import Response, Session
+
+    from docling_core.utils.file import resolve_source_to_stream
```
ceberam: Just wondering if there is any benefit in repeating these inline imports for each test over adding the imports only once at module-level

ceberam (commented):
I added some comments.
By the way, `resolve_source_to_stream` is now very robust. We should reuse it in the declarative backends. I'll create an issue.

docling_core/utils/file.py
cau-git: we can for now factor it as a module level constant, that should be enough.

docling_core/utils/file.py
cau-git: fair points!

test/test_utils.py
cau-git: will remove them.

ceberam (approved):
🚀

dolfim-ibm (approved):

PR #642, opened 2026-06-17
feat(doclang): default `include_version` to `True`

vagenas:
- emit the DocLang version in serialized output by default
- pass `include_version=False` in tests that compare version-less fixtures


PeterStaar-IBM (approved):
lgtm!

cau-git (approved):

PR #643, opened 2026-06-17
fix: clamp provenance bboxes and DocLang locations to valid page bounds

vagenas:
- clamp bbox coordinates during document validation and quantize location tokens at page edges without warning
- emit warnings only when values exceed a 1% page-dimension tolerance


PeterStaar-IBM (approved):
lgtm!

dolfim-ibm (approved):

PR #656, opened 2026-06-24
fix: doclang suppress content filtered shells

nassarofficial:
Problem When we serialize with narrow content_types (e.g. picture-only prompts like [<picture>]), the serializer was still emitting head-only shells — elements with metadata (<layer>, <thread>, <location>, <label>) but no actual supervised body content. That leaked spurious tags into WDS task-filtered training targets.

Fix (opt-in only — default behavior unchanged) All new logic is gated behind suppress_empty_elements=True (default remains False):

• Text / table / picture: if an element’s body type is filtered out and there’s no visible content, drop the element entirely instead of emitting an empty tag or metadata-only shell
• Layout exception: if add_location=True and the item has provenance, keep the element so layout supervision still gets boxes
• Picture labels: under suppression, <label> is only emitted when picture/chart/chemistry content is actually requested (layout-only prompts get boxes, no classification label)
• New param: emit_picture_layer (default True) — lets callers omit <layer> on pictures (we set this to False in granite for picture-classification prompts)

### DIFF
--- a/docling_core/transforms/serializer/doclang.py
+++ b/docling_core/transforms/serializer/doclang.py
@@ -239,6 +239,10 @@ class DocLangParams(CommonParams):
         _advanced_field(detail="Types of content to serialize (only relevant if add_content is True)."),
     ] = _DEFAULT_CONTENT_TYPES
     layer_mode: Annotated[LayerMode, _advanced_field()] = LayerMode.AUTO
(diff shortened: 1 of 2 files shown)


PeterStaar-IBM:
@nassarofficial can yopu fix the DCO and tests

PR #660, opened 2026-06-25
feat: add `ContentLayer.SHEET` and serialize Excel sheet groups in HTML and Markdown

samiuc:
When converting a multi-sheet Excel file, there was no way to tell which table or piece of content belonged to which worksheet in the HTML or Markdown output. This PR fixes that by tagging visible worksheet groups with a dedicated content layer, so the HTML serializer can wrap each sheet's content in a labeled `<section>` element and the Markdown serializer can emit a heading for each sheet.

### DIFF
--- a/docling_core/transforms/serializer/html.py
+++ b/docling_core/transforms/serializer/html.py
@@ -102,7 +102,7 @@ class HTMLParams(CommonParams):
     """HTML-specific serialization parameters."""
 
     # Default layers to use for HTML export
-    layers: set[ContentLayer] = {ContentLayer.BODY}
+    layers: set[ContentLayer] = {ContentLayer.BODY, ContentLayer.SHEET}
 
     # How to handle images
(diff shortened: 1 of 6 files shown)


ceberam:
@samiuc I don't think we should create a new content layer for the structural component derived from a workbook sheet for the same reason that we don't have a content layer for PDF pages or EPUB chapters.
Please, check the work done in https://github.com/docling-project/docling/pull/3635 since you should be able to achieve what you need with the current code, at least for the Markdown serializer. We will follow the approach described on that PR thread to generalize the rendering of GroupLabel.SHEET from [MsExcelMarkdownDocSerializer](https://github.com/docling-project/docling-core/blob/main/docling_core/transforms/serializer/markdown_excel.py#L48) to the default Markdown and HTML serializer.

ceberam:
Closing for the reason above and https://github.com/docling-project/docling/pull/3700

PR #662, opened 2026-06-27
feat: add enable_chart_tables and traverse_pictures to save_as_markdown

### LINKED ISSUE #619: Feature request: forward all export_to_markdown kwargs through save_as_markdown
### Problem

`DoclingDocument.save_as_markdown(...)` and `DoclingDocument.export_to_markdown(...)` have diverged. `save_as_markdown` internally calls `_make_copy_with_refmode(...)` (to materialise picture URIs against an output-directory-relative `reference_path`) and then calls `export_to_markdown(...)`, but it does not forward six of the latter's keyword arguments. Callers who want any of those six features cannot use `save_as_markdown` and must reproduce the rewrite hop themselves.

Verified against `docling-core==2.77.0` via `inspect.signature(...)`. Parameters present in `export_to_markdown` but **missing from** `save_as_markdown`:

| Parameter | `export_to_markdown` default | Notes |
| --- | --- | --- |
| `allowed_meta_names` | `None` | Meta-name allowlist; needed by callers filtering picture descriptions etc. |
| `blocked_meta_names` | `None` | Inverse of allowed_meta_names; takes precedence per the docstring. |
| `enable_chart_tables` | `True` | Whether chart serialisation falls back to tabular rendering. |
| `mark_annotations` | `False` | Whether to mark annotations in the output. |
| `traverse_pictures` | `False` | Required for scanned/full-page-OCR PDFs where text lives under a top-level `PictureItem`. |

Plus a naming inconsistency between the two methods for the same feature:

| Feature | `save_as_markdown` param name | `export_to_markdown` param name |
| --- | --- | --- |
| escape underscores in text | `escaping_underscores` | `escape_underscores` |

### Impact

Callers who need any of the six missing features must do the work `save_as_markdown` already does internally — copy the document with `_make_copy_with_refmode(...)` (a private API), then call `export_to_markdown(...)` with the desired kwargs, then write the result to disk. We did this in our codebase as a workaround. Removing it would let us delete a meaningful chunk of in-project code that exists only to reach features that already work in `export_to_markdown`.

### Proposed fix

Forward every `export_to_markdown` kwarg from `save_as_markdown`. The simplest shape is to take the same signature (minus the trailing return-to-string semantics) — `save_as_markdown` already wraps `export_to_markdown`, so the change is mechanical and the test surface is small.

Alongside that, consider renaming `escaping_underscores` → `escape_underscores` on `save_as_markdown` for consistency (keep the old name as a deprecated alias if backward compatibility matters).

Happy to open a PR if a preferred shape is indicated.

### Context

Found while resolving our project's in-progress private-API dependency on `DoclingDocument._make_copy_with_refmode`. Once `save_as_markdown` exposes the six missing kwargs (especially `allowed_meta_names`), our in-project URI-rewrite helper can collapse back to a single `document.save_as_markdown(..., allowed_meta_names={...})` call.

maxmilian:
### What

`save_as_markdown` already forwards most of `export_to_markdown`'s parameters, but five were missing, so callers using `save_as_markdown` silently got the defaults for them:

- `enable_chart_tables` (default `True`)
- ~`mark_annotations` (default `False`)~
- `traverse_pictures` (default `False`)
- ~`allowed_meta_names` (default `None`)~
- ~`blocked_meta_names` (default `None`)~

This adds them as keyword-only arguments with the same defaults as `export_to_markdown` and forwards them through.

Note: `mark_annotations`, `allowed_meta_names`, and `blocked_meta_names` are not propagated since they are either deprecated or legacy arguments from old implementations.

### Why

Fixes #619 — `save_as_markdown` should expose the same output options as `export_to_markdown` rather than dropping a subset to defaults.

### Notes

- Defaults are identical to `export_to_markdown`, so existing behavior is unchanged.
- New parameters are keyword-only (added after `*`), so no existing positional call to `save_as_markdown` is affected.
- The `escaping_underscores` → `escape_underscores` naming note from the earlier attempt (#629) is already handled on `main`, so it is intentionally left untouched here.
- Kept to a single-file, behavior-preserving change with no test-data changes; existing serialization/document tests pass locally (ruff, mypy, pytest, docs pre-commit hooks all green).

### DIFF
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -6322,10 +6322,7 @@ class DoclingDocument(BaseModel):
         compact_tables: bool = False,
         *,
         enable_chart_tables: bool = True,
-        mark_annotations: bool = False,
         traverse_pictures: bool = False,
-        allowed_meta_names: Optional[set[str]] = None,
-        blocked_meta_names: Optional[set[str]] = None,
         mark_meta: bool = False,
         use_legacy_annotations: Optional[bool] = None,  # deprecated
     ):
@@ -6356,12 +6353,9 @@ class DoclingDocument(BaseModel):
             included_content_layers=included_content_layers,
             page_break_placeholder=page_break_placeholder,
             include_annotations=include_annotations,
-            mark_annotations=mark_annotations,
             compact_tables=compact_tables,
             traverse_pictures=traverse_pictures,
             use_legacy_annotations=use_legacy_annotations,
-            allowed_meta_names=allowed_meta_names,
-            blocked_meta_names=blocked_meta_names,
             mark_meta=mark_meta,
         )
 
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -6321,6 +6321,11 @@ class DoclingDocument(BaseModel):
         include_annotations: bool = True,
         compact_tables: bool = False,
         *,
+        enable_chart_tables: bool = True,
+        mark_annotations: bool = False,
+        traverse_pictures: bool = False,
+        allowed_meta_names: Optional[set[str]] = None,
+        blocked_meta_names: Optional[set[str]] = None,
         mark_meta: bool = False,
         use_legacy_annotations: Optional[bool] = None,  # deprecated
     ):
@@ -6343,6 +6348,7 @@ class DoclingDocument(BaseModel):
             escape_html=escape_html,
             escape_underscores=escaping_underscores,
             image_placeholder=image_placeholder,
+            enable_chart_tables=enable_chart_tables,
             image_mode=image_mode,
             indent=indent,
             text_width=text_width,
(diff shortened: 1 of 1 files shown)


dolfim-ibm (dismissed):
lgtm

ceberam (approved):
lgtm

maxmilian:
Gentle bump on this one 🙏

It's been sitting green since June 29 — two approvals (@dolfim-ibm, @ceberam), 11/11 checks passing, and Mergify reports merge protection satisfied. I think it just needs someone to press the button.

Happy to rebase if it's drifted or if anything else is wanted here — just say the word.

ceberam (changes requested):
@maxmilian After another review, we think that the following arguments should **not** be exposed in `save_as_markdown`:
- `mark_annotations`, since it is deprecated in `MarkdownParams`
- `allowed_meta_names` and `blocked_meta_names`, since they refer to old implementations and we would prefer not to further promote them in the API

Therefore only these arguments should be added: `enable_chart_tables` and `traverse_pictures`.

maxmilian:
@ceberam Makes sense — thanks for the second look. Pushed a follow-up commit that drops the three arguments:

- `mark_annotations` (deprecated in `MarkdownParams`)
- `allowed_meta_names` / `blocked_meta_names` (old implementations, not worth promoting further in the API)

`save_as_markdown` now only gains `enable_chart_tables` and `traverse_pictures`, both forwarded with the same defaults as `export_to_markdown`. Net diff is 4 added lines in `docling_core/types/doc/document.py`.

Ready for another look whenever you have a moment.

ceberam (approved):
Great! I have changed the PR title and description accordingly.

dolfim-ibm (approved):

PR #667, opened 2026-06-29
feat(document): add change_type field to TextItem

a-huk:
Adds an optional change_type field to TextItem with values 'inserted' or 'deleted' (None means normal/final text). Also threads the parameter through DoclingDocument.add_text() so backends can set it at creation time.

Needed by the docling DOCX backend to represent tracked changes in 'raw' mode without hijacking underline/strikethrough formatting.

### DIFF
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -1977,6 +1977,7 @@ class TextItem(DocItem):
 
     formatting: Optional[Formatting] = None
     hyperlink: Optional[Union[AnyUrl, Path]] = Field(union_mode="left_to_right", default=None)
+    change_type: Optional[typing.Literal["inserted", "deleted"]] = None
 
     @deprecated("Use export_to_doctags() instead.")
     def export_to_document_tokens(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


ceberam (changes requested):
Thanks @a-huk , it's a good starting draft, but there are quite a few things that need attention:
 
- The golden YAML file `test/data/docling_document/unit/TextItem.yaml` does not include `change_type`, so the test fails with:

```
Left contains 1 more item: {'change_type': None}
```

- `change_type` is silently dropped for labels that are dispatched to helper methods.
Inside `add_text()`, labels like `TITLE`, `LIST_ITEM`, or `CODE` are all forwarded to specialized helpers (`add_title`, `add_list_item`, `add_heading`, etc.). None of which accept or forward `change_type`. The parameter is silently ignored. We may want to track changes on those items too.
- Following up with the previous bullet: what if a contributor drops an entire table from the `.docx` document? We should better define `change_type` as a field of `DocItem`
- The PR adds no test verifying that a `TextItem` constructed with `change_type="inserted"` round-trips correctly (serialize → deserialize), nor that `add_text()` propagates the value onto the returned item. A single parametrized test would suffice. 
-  It would be good to add the support of `change_type` in at least one serializer, to show the intention of the new field and get an idea about how inserted or deleted text should be rendered compared to _normal_ text. The Markdown or the HTML serializers would be excellent candidates.
- I have doubts about the name of the new field. `change_type` is generic and ambiguous. Also _change_ and _type_ are both overloaded words in any codebase. It also encodes only the _kind_ of change, not the fact that this is a tracked/revision-control concept. What about `tracked_change` and `insertion` and `deletion` for its values?

PR #668, opened 2026-06-30
feat(document): register the Box Note mime type

pablopupo:
`DocumentOrigin` rejects `application/vnd.box.boxnote` because it isn't in Python's `mimetypes` registry, so the Box Note backend in docling (docling-project/docling#3722) falls back to `application/json`. This registers it in `_extra_mimetypes`, alongside the other vendor types, so the backend can use the correct mime.

Per @ceberam's review on that PR.

### DIFF
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -1074,22 +1074,23 @@ class DocumentOrigin(BaseModel):
     )
 
     _extra_mimetypes: typing.ClassVar[list[str]] = [
+        "application/vnd.box.boxnote",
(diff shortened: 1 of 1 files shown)


ceberam (changes requested):
Thanks @pablopupo for submitting this PR so quickly.
Since you're touching the `_extra_mimetypes` list and to keep things tidy, would you mind sorting it in alphabetical order?
For this small change, I would rather give another commit type like `chore`.

pablopupo:
Done!

ceberam (approved):

cau-git (approved):

PR #673, opened 2026-07-02
fix(chunker): detect markdown table header and body using separator regex

hardik-kansal:
## Problem Summary

When `HybridChunker` splits a large table across multiple chunks with `repeat_table_header=True`, the first chunk formats correctly. However, subsequent chunks fail to render as Markdown tables. The serializer fails to output the mandatory | - | separator line beneath head

### Root Cause of the Failures:
The original code assumed the table separator was always at line index 1. However, when a table had a caption or other preamble preceding it, line index 0 was the caption, and line index 1 was the column header row. This meant the actual separator row (at index 2) was treated as part of the body, so it got absorbed in the first chunk's body instead of being repeated. Subsequent chunks received only the caption and header row, leaving them without a markdown table separator (rendering them invalid markdown).

### `docling_core/transforms/chunker/hybrid_chunker.py`

- Replaced the fixed-index separator detection in `MarkdownTableSerializer.get_header_and_body_lines` with a regex scan (r"^\|(\s*:?-+:?\s*\|)+\s*$")
- The code now uses the regex to find the separator line (`| --- |`). Everything from the start of the table up to and including this separator row (e.g. captions, headers, and the separator itself) is grouped into `header_lines`. Everything following the separator is treated as `body_lines`.

After this change, only two tests in the test suite were failing:
1. `test_chunk_with_repeat_table_header` in `test/test_hybrid_chunker.py`
2. `test_chunk_html_table_serializer` in `test/test_hybrid_chunker.py`

#### Regenerated ground truth files 0d_out_chunks.json and 0e_out_chunks.json to reflect the corrected chunking output" using
```
DOCLING_GEN_TEST_DATA=1 uv run pytest test/test_hybrid_chunker.py::test_chunk_with_repeat_table_header test/test_hybrid_chunker.py::test_chunk_html_table_serializer -v 2>&1 | tail -20
```

- Resulted new chunks are stored in these files, also updated in ground truth
[0e_out.txt](https://github.com/user-attachments/files/29615636/0e_out.txt)
[0d_out.txt](https://github.com/user-attachments/files/29615639/0d_out.txt)

- new json objects maps table structure better with no duplicate \n and with seperator

## Testing
- Verified that all tests now pass cleanly (`uv run pre-commit run --all-files`)


ceberam:
Superseded by #678

PR #675, opened 2026-07-07
fix(serializer): serialize table footnotes in Markdown and HTML

### LINKED ISSUE #496: Regression: Table footnotes missing from markdown serialization after commit dda9c88
### Summary

After commit dda9c88, table footnotes are no longer present in markdown export. Before this commit, footnotes were serialized as TextItem during the items iteration and appeared in the markdown output. The commit changed footnote handling to delegate serialization to the parent item (e.g., TableItem), but the markdown serializers (like MarkdownTableSerializer) were not updated to serialize footnotes, so they are now missing.
**Important**: Being a serialization problem, it affects both document exports and chunking.

### Steps to Reproduce

Create a document with a table that includes a footnote.
Export the document to markdown.
Observe that the table footnote is missing in the markdown output, even if the textitem were correctly assigned as footnotes to the table.

Example pdf
[table_parsing_test.pdf](https://github.com/user-attachments/files/24964670/table_parsing_test.pdf)

Code to reproduce:
```

converter = DocumentConverter()
conv_res = converter.convert("table_parsing_test.pdf")
json_export = conv_res.document.export_to_dict()
md_export = conv_res.document.export_to_markdown()
```
md_export output:
```
## Structured Document Parsing Test

The following inventory report is designed to test how document parsers handle multi-line annotations that are geographically tied to a table structure.

| Record ID   | Asset Category   |   Unit Count | Status      |
|-------------|------------------|--------------|-------------|
| 440-A       | Hardware         |          142 | Operational |
| 440-B       | Software License |           12 | Pending*    |
| 441-C       | Cloud Instance   |            3 | Operational |
| 442-D       | Legacy Support   |            1 | Archived**  |

This final paragraph provides a clear end-of-section marker. Parsers should correctly associate the two footnotes with the table above rather than this trailing body text.
```

### Expected Behavior
Table footnotes should appear in the markdown export, as they did before commit dda9c88.

### Actual Behavior
Table footnotes are omitted from the markdown export.

### Technical Details

Before dda9c88, footnotes were serialized as TextItem during the main iteration.
After dda9c88, footnotes are delegated to the parent floating item for serialization, but MarkdownTableSerializer does not call the footnote serialization method, so footnotes are not included.
The markdown serializers need to be updated to serialize table footnotes via the appropriate method.

References

[Commit dda9c88](https://github.com/docling-project/docling-core/commit/dda9c88)
[MarkdownTableSerializer code](https://github.com/docling-project/docling-core/blob/3b0b909c894a37eee980e4f99d675a6de0ff5891/docling_core/transforms/serializer/markdown.py#L351-L842)
[Discussion of the regression and cause](https://github.com/docling-project/docling-core/blob/3b0b909c894a37eee980e4f99d675a6de0ff5891/docling_core/transforms/serializer/common.py#L333-L626)

### Suggested Fix
Update MarkdownTableSerializer to serialize table footnotes using the appropriate method, similar to how captions are handled.

Kavishkagaya:
Fixes #496

## Problem

Table footnotes are silently dropped from **Markdown and HTML** output. The footnotes are correctly stored on `TableItem.footnotes` in the document model (they appear in the JSON), but never make it into the serialized text — so a table renders with dangling `*` / `**` markers whose definitions have vanished. As #496 notes, this also affects chunking, which relies on serialization.

## Root cause

After footnote handling was delegated to the parent `FloatingItem`, the `serialize_footnotes` hook was added to the doc serializer — but the table serializers were never wired to call it. They call `serialize_captions` (so captions appear) but not `serialize_footnotes`, so the method was effectively dead code and the footnotes had nowhere to go.

## Fix

- **`MarkdownTableSerializer`** — emit the table's footnotes after the table body, mirroring the existing caption handling.
- **`HTMLTableSerializer`** — emit the footnotes after the closing `</table>`.
- **`HTMLDocSerializer`** — add a `serialize_footnotes` override so footnotes are HTML-escaped and wrapped in `<div class="footnote">`, matching the existing `serialize_captions` override (the base method returns raw text, which is unsafe for HTML).

## Evidence

Using the `table_parsing_test.pdf` attached to #496 (digital PDF, standard pipeline). The table has two footnotes assigned; before the fix both are missing from the Markdown, after the fix both are present:

**Before** (`export_to_markdown()`):
```
| 442-D | Legacy Support | 1 | Archived** |

This final paragraph provides a clear end-of-section marker. ...
```

**After**:
```
| 442-D | Legacy Support | 1 | Archived** |

*Note: The 'Pending' status indicates that the asset is awaiting final security validation. **Note: 'Archived' items are scheduled for decommissioning at the end of the current fiscal year.

This final paragraph provides a clear end-of-section marker. ...
```

HTML output, after the fix, now emits (escaped):
```html
<table>...</table><div class="footnote">*Note: The 'Pending' status ...</div><div class="footnote">**Note: 'Archived' items ...</div>
```

> Note: #496 only mentions Markdown, but **HTML was affected too** — this PR fixes both.

## Testing

- Added a regression test (`test_table_footnotes_serialized_to_markdown_and_html`) that builds an in-memory `DoclingDocument` with a table + two footnotes and asserts they survive both `export_to_markdown()` and `export_to_html()`.
- Full existing suites pass with no regressions: serialization (`test_serialization.py`, `test_serialization_doctag.py`, `test_latex_serialization.py`, `test_azure_serializer.py`) and chunking (`test_hierarchical_chunker.py`).
- `ruff check` and `ruff format --check` clean.

### DIFF
--- a/docling_core/transforms/serializer/html.py
+++ b/docling_core/transforms/serializer/html.py
@@ -443,6 +443,11 @@ class HTMLTableSerializer(BaseTableSerializer):
         text_res = "".join([r.text for r in res_parts])
         text_res = f"<table>{text_res}</table>" if text_res else ""
 
+        ftn_res = doc_serializer.serialize_footnotes(item=item, **kwargs)
+        if ftn_res.text:
+            text_res = f"{text_res}{ftn_res.text}" if text_res else ftn_res.text
+            res_parts.append(ftn_res)
+
         return create_ser_result(text=text_res, span_source=res_parts)
 
     @override
@@ -1274,6 +1279,33 @@ class HTMLDocSerializer(DocSerializer):
         text_res = params.caption_delim.join([r.text for r in results])
         if text_res:
             text_res = f"<{tag}>{text_res}</{tag}>"
+        return create_ser_result(text=text_res, span_source=results)
+
+    @override
+    def serialize_footnotes(
+        self,
+        item: FloatingItem,
+        **kwargs: Any,
+    ) -> SerializationResult:
+        """Serialize the item's footnotes."""
+        params = self.params.merge_with_patch(patch=kwargs)
+        results: list[SerializationResult] = []
+        excluded_refs = self.get_excluded_refs(**kwargs)
+
+        if DocItemLabel.FOOTNOTE in params.labels:
+            for ftn in item.footnotes:
+                if isinstance(it := ftn.resolve(self.doc), TextItem) and it.self_ref not in excluded_refs:
+                    text_ftn = it.text
+                    text_dir = get_text_direction(text_ftn)
+                    dir_str = f' dir="{text_dir}"' if text_dir == "rtl" else ""
+                    results.append(
+                        create_ser_result(
+                            text=(f'<div class="footnote"{dir_str}>{html.escape(text_ftn)}</div>'),
+                            span_source=it,
+                        )
+                    )
+
+        text_res = params.caption_delim.join([r.text for r in results])
         return create_ser_result(text=text_res, span_source=results)
 
     def _generate_head(self) -> str:
--- a/docling_core/transforms/serializer/markdown.py
+++ b/docling_core/transforms/serializer/markdown.py
@@ -576,6 +576,10 @@ class MarkdownTableSerializer(BaseTableSerializer):
             if table_text:
                 res_parts.append(create_ser_result(text=table_text, span_source=item))
 
+            ftn_res = doc_serializer.serialize_footnotes(item=item, **kwargs)
+            if ftn_res.text:
+                res_parts.append(ftn_res)
+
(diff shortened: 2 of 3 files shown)


Kavishkagaya:
Aware of the related #545 and #569 targeting the same issue — opening this as a smaller, focused alternative:

- Fixes **both Markdown and HTML** (#569 is Markdown-only; HTML drops table footnotes too, which #496 doesn't mention).
- Conflict-free against current `main` (#545 currently has merge conflicts).
- Minimal change: reuses the existing but never-called `serialize_footnotes` hook, plus a matching `HTMLDocSerializer.serialize_footnotes` override (the base method returns raw text, unsafe for HTML), mirroring the existing `serialize_captions` handling.

Happy to fold this into either of the other PRs if the maintainers prefer to consolidate.

PR #676, opened 2026-07-08
fix(tokens): normalize inverted bbox in get_location instead of asserting

### LINKED ISSUE #620: Redundant bbox assertion in DocumentToken.get_location crashes export_to_doctags on near-degenerate elements
AI Generated Report:

## Summary

`DocumentToken.get_location()` in `docling_core/types/doc/tokens.py` enforces `bbox[0] <= bbox[2]` and `bbox[1] <= bbox[3]` via `assert`, but the next 4 lines of the same method clamp the values with `min()` / `max()` — so the asserts are redundant with the downstream normalization the method already performs. Near-degenerate elements (e.g. a 1-pt rule whose layout-model regression produced a slightly inverted bbox) crash `export_to_doctags(add_location=True)` even though the method's own output would be valid.

## Repro

We hit this on a ~927-page PDF when running `export_to_doctags(add_location=True)` on a 153-page chunk:

```
File ".../docling_core/types/doc/tokens.py", line 307, in get_location
    assert bbox[1] <= bbox[3], f"bbox[1]<=bbox[3] => {bbox[1]}<={bbox[3]}"
AssertionError: bbox[1]<=bbox[3] => 633.3925132751465<=625.6789665222168
```

The inversion is only 7.7 pt — essentially a near-zero-height element (likely a decorative rule, a tiny diacritic, or a layout-model float artifact). Probability of hitting at least one such element grows with page count, so it bites long documents in particular.

## The asserts are redundant with the method's own clamp

`docling-core` `main`, `docling_core/types/doc/tokens.py`, the body of `get_location`:

```python
assert bbox[0] <= bbox[2], f"bbox[0]<=bbox[2] => {bbox[0]}<={bbox[2]}"
assert bbox[1] <= bbox[3], f"bbox[1]<=bbox[3] => {bbox[1]}<={bbox[3]}"

x0 = bbox[0] / page_w
y0 = bbox[1] / page_h
x1 = bbox[2] / page_w
y1 = bbox[3] / page_h

x0_tok = DocumentToken.get_location_token(val=min(x0, x1), rnorm=xsize, self_closing=self_closing)
y0_tok = DocumentToken.get_location_token(val=min(y0, y1), rnorm=ysize, self_closing=self_closing)
x1_tok = DocumentToken.get_location_token(val=max(x0, x1), rnorm=xsize, self_closing=self_closing)
y1_tok = DocumentToken.get_location_token(val=max(y0, y1), rnorm=ysize, self_closing=self_closing)
```

The `min` / `max` calls already produce correctly-ordered location tokens regardless of input order. The asserts add no protection — they just kill `export_to_doctags` when the layout pipeline emits a slightly inverted box.

## Suggested fix

Drop the asserts, or replace them with explicit normalization that mirrors the downstream `min`/`max`:

```python
x0, x1 = sorted((bbox[0], bbox[2]))
y0, y1 = sorted((bbox[1], bbox[3]))
```

then divide by `page_w` / `page_h` and emit tokens. The same code/issue exists at `docling_core/types/legacy_doc/tokens.py:177`.

## Environment

- `docling==2.93.0`
- `docling-core==2.75.0`
- `docling-ibm-models==3.13.2`
- Python 3.12

maxtaran2010:
Fixes #620

## Problem

\`DocumentToken.get_location()\` in \`docling_core/types/doc/tokens.py\` asserted \`bbox[0] <= bbox[2]\` and \`bbox[1] <= bbox[3]\`. Near-degenerate elements (e.g. a thin decorative rule whose layout-model regression produces a slightly inverted bbox) can violate this by a fraction of a point, which crashes \`export_to_doctags(add_location=True)\`:

```
File ".../docling_core/types/doc/tokens.py", line 307, in get_location
    assert bbox[1] <= bbox[3], f"bbox[1]<=bbox[3] => {bbox[1]}<={bbox[3]}"
AssertionError: bbox[1]<=bbox[3] => 633.3925132751465<=625.6789665222168
```

The asserts were redundant: the method already emits correctly-ordered location tokens via \`min()\`/\`max()\` downstream. The probability of hitting one such element grows with page count, so long documents are the most affected.

## Fix

Sort the coordinates before normalization (mirroring the existing \`min()\`/\`max()\` behaviour) and drop the asserts. Output is unchanged for well-formed boxes and now correct instead of crashing for inverted ones.

## Tests

Added \`test/test_tokens.py\` covering:
- a normal bbox (exact token output)
- the partially inverted bbox from the issue repro
- a fully inverted bbox

All pass; existing \`test_serialization_doctag.py\` and \`test_doctags_load.py\` (17 tests) still pass, and \`ruff\` is clean.

### DIFF
--- a/docling_core/types/doc/tokens.py
+++ b/docling_core/types/doc/tokens.py
@@ -302,7 +302,7 @@ class DocumentToken(str, Enum):
         ysize: int = 500,  # TODO review
         self_closing: bool = False,
     ):
-        """Get the location string give bbox and page-dim."""
+        """Get the location string given bbox and page-dim."""
         # Normalize potentially inverted coordinates instead of asserting on
         # them. Near-degenerate elements (e.g. a thin rule whose layout-model
         # regression produced a slightly inverted bbox) can arrive with
--- a/docling_core/types/doc/tokens.py
+++ b/docling_core/types/doc/tokens.py
@@ -303,13 +303,19 @@ class DocumentToken(str, Enum):
         self_closing: bool = False,
     ):
         """Get the location string give bbox and page-dim."""
-        assert bbox[0] <= bbox[2], f"bbox[0]<=bbox[2] => {bbox[0]}<={bbox[2]}"
-        assert bbox[1] <= bbox[3], f"bbox[1]<=bbox[3] => {bbox[1]}<={bbox[3]}"
+        # Normalize potentially inverted coordinates instead of asserting on
+        # them. Near-degenerate elements (e.g. a thin rule whose layout-model
+        # regression produced a slightly inverted bbox) can arrive with
+        # bbox[0] > bbox[2] or bbox[1] > bbox[3]. The min()/max() calls below
+        # already emit correctly-ordered tokens, so sorting here mirrors that
+        # normalization and avoids crashing export_to_doctags on valid output.
+        left, right = sorted((bbox[0], bbox[2]))
+        top, bottom = sorted((bbox[1], bbox[3]))
 
-        x0 = bbox[0] / page_w
-        y0 = bbox[1] / page_h
-        x1 = bbox[2] / page_w
-        y1 = bbox[3] / page_h
+        x0 = left / page_w
+        y0 = top / page_h
+        x1 = right / page_w
+        y1 = bottom / page_h
 
         x0_tok = DocumentToken.get_location_token(val=min(x0, x1), rnorm=xsize, self_closing=self_closing)
         y0_tok = DocumentToken.get_location_token(val=min(y0, y1), rnorm=ysize, self_closing=self_closing)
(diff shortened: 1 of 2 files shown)


ceberam (changes requested):
Thanks @maxtaran2010 for this contribution. The fix is correct and the change itself is clean and minimal.

A couple of comments:
- However, after the sort, `left <= right` and `top <= bottom` is guaranteed, so dividing by positive page dimensions preserves that: `x0 <= x1` and `y0 <= y1`. The four calls on lines 320–323 can now be written without `min()` or `max()`. Could you please simplify those lines?
- The inline comment on `get_location` is very informative for users to understand the eventual normalization. Could you please add the comment to the function docstring instead of the inline comments?

ceberam (approved):
Thanks again @maxtaran2010 for the fix! I went ahead and applied the two follow-up cleanups myself to keep things moving (the PR had been open for a couple of weeks without activity). I dropped the redundant `min()`/`max()` calls that became unnecessary after the sort, and moved the normalization explanation from the inline comment into the docstring. The logic and tests are untouched. Thanks again for the contribution! 🚀

PeterStaar-IBM (approved):
lgtm!

PR #678, opened 2026-07-08
fix(chunker): preserve table structure across chunks in markdown

ceberam:
## Summary

This PR fixes a bug in the hybrid chunker where splitting a Markdown table
across multiple chunks produced invalid Markdown in all chunks except the last.
It also improves the maintainability of the related test suite and upgrades the
inline documentation of the affected methods.

Resolves docling-project/docling#3736

### Spurious blank line between header row and separator row

When `repeat_table_header=True` and a table exceeds the token budget, the
chunker repeats the header rows at the top of each chunk. The header
repetition used `"\n".join(header_lines)` on lines that already carried
their own `\n` terminators, inserting a spurious blank line between the
header row and the separator row (`| header |\n\n| --- |`). This breaks
Markdown rendering in every chunk except the last.

### Table caption/preamble repeated in all chunks

Any text that precedes the first table row (e.g. a caption such as
_"Table 1: Runtime characteristics…"_) was being included in the repeated
header prefix and therefore appeared at the top of every chunk of a split
table. It should appear only in the first chunk. After the fix, the
preamble is accounted for in the token budget during splitting but is
stripped from all chunks after the first. This is in line with #527 , which
motivated the table header duplication functionality to the HybridChunker
and introduced the new LineBasedTokenChunker.

Both issues are visible in the corrected ground truth snapshots: the
second and subsequent chunks of a split table now start cleanly with the
header row and separator row on consecutive lines, with no caption
preceding them.

## Test improvements

The test module for the hybrid chunker had significant duplication: the same
document was loaded, the same serializer-provider class was defined, and the
same chunker was constructed independently in multiple tests. These have been
consolidated into module-scoped pytest fixtures and module-level provider
classes, reducing boilerplate without changing what is being tested.

## Documentation

Add proper Google-style docstrings, making the logic discoverable via tooling and
keeping the code body uncluttered.

A stale docstring on `LineBasedTokenChunker.max_tokens` stated _"if not
set, limit is resolved from the tokenizer"_, copied from `HybridChunker`
where `max_tokens` is an optional field with a tokenizer fallback. In
`LineBasedTokenChunker` the property is a direct passthrough to
`tokenizer.get_max_tokens()` with no override mechanism, so the docstring
has been corrected to reflect that.


PeterStaar-IBM (dismissed):
lgtm!

dolfim-ibm (approved):

vagenas (approved):

PR #664, opened 2026-06-28
feat: split the monolithic `document.py` into focused modules

PeterStaar-IBM:
## Summary

  `docling_core/types/doc/document.py` had grown to **8164 lines** holding ~60
  classes — `DoclingDocument` plus every `*Item`/`*Group` type and all their
  supporting models. This PR splits it into a clear module tree so that
  `document.py` now contains **only `DoclingDocument`** (5446 lines, almost all of
  which is that one class), while each item family lives in its own file.

  No behavioural changes — this is a pure structural refactor. The public API and
  all import paths are preserved.

  ## Motivation

  - `document.py` was too large and mixed many unrelated concerns.
  - Goal: **one file per `*Item`/`*Group` family** (e.g. table classes together,
    picture classes together), each co-located with the data models it owns.

  ## New layout

```
  docling_core/types/doc/
  ├── document.py            # DoclingDocument only
  ├── doctags.py             # DocTagsPage, DocTagsDocument
  ├── common/                # cross-cutting value models
  │   ├── scalars.py         # Uint64, LevelNumber, CharSpan
  │   ├── constants.py       # CURRENT_VERSION, *_EXPORT_LABELS
  │   ├── content_layer.py   # ContentLayer
  │   ├── formatting.py      # Formatting, Script
  │   ├── annotations.py     # BaseAnnotation, Description/MiscAnnotation
  │   ├── reference.py       # RefItem, FineRef, ImageRef, ProvenanceItem
  │   ├── origin.py          # DocumentOrigin, BaseSource, TrackSource, SourceType
  │   ├── meta.py            # BaseMeta + all *MetaField + MetaUtils
  │   └── page_item.py       # PageItem
  └── items/                 # document-tree node types
      ├── node.py            # NodeItem, DocItem, FloatingItem (bases)
      ├── group.py           # GroupItem, ListGroup, OrderedList, InlineGroup
      ├── text.py            # TextItem, Title/SectionHeader/List/Formula
      ├── code.py            # CodeItem
      ├── key_value.py       # GraphCell/Link/Data, KeyValueItem, FormItem
      ├── form.py            # Field{Region,Heading,,Value}Item
      ├── content.py         # ContentItem union
      ├── picture/           # charts, classification, molecule, picture (PictureItem)
      └── table/             # table_data (TableData/cells), table (TableItem)
```

  ## Design notes

  - **Acyclic by construction.** The model stores tree children as `RefItem`
    (not embedded item types) and references `DoclingDocument` only via
[...]

PR #680, opened 2026-07-08
feat(markdown): include table image in Markdown export when available

### LINKED ISSUE #474: Feature Request: Support table image output in Markdown export
# Feature Request: Support table image output in Markdown export

## Description
When exporting a document to Markdown using `docling`, `TableItem` elements are currently only exported as text-based Markdown tables (or HTML). Even when `image_mode` is set to `ImageRefMode.REFERENCED` or `ImageRefMode.EMBEDDED` and `generate_table_images=True` is used, the table images are not included in the Markdown output.

This contrasts with `PictureItem` elements, which correctly respect the `image_mode` setting and output image links.

This issue was originally reported in `docling` (https://github.com/docling-project/docling/issues/2820), but the root cause and fix lie within `docling-core`'s serialization logic.

## Current Behavior
- `MarkdownTableSerializer` (in `docling_core`) strictly exports tables as text/markdown grids.
- It ignores the `image_mode` parameter for tables.
- It ignores the presence of `item.image` (if populated).

## Expected Behavior
If `image_mode` is set to `REFERENCED` or `EMBEDDED`, and the `TableItem` has an associated image (`item.image`), the serializer should:
1.  Preferentially output the image (or include it alongside the text, depending on desired behavior, but likely behaving similar to `PictureItem`).
2.  Generate the appropriate Markdown image syntax (e.g., `Table` or embedded base64).

## Proposed Solution
Modify `MarkdownTableItemSerializer` (or `MarkdownTableSerializer`) in `docling_core/transforms/serializer/markdown.py`.

The `serialize` method should be updated to check `self.params.image_mode`:

```python
# Pseudo-code logic to be added to serialize()

if self.params.image_mode in [ImageRefMode.REFERENCED, ImageRefMode.EMBEDDED] and item.image:
    # Logic similar to MarkdownPictureSerializer
    image_link = self._get_image_link(item) # Helper to generate path/base64
    return getattr(self, "create_ser_result")(text=f"![{label}]({image_link})", span_source=item)

# Fallback to existing text table generation
...

maxmilian:
Closes #474.

## What

`MarkdownTableSerializer` only ever emitted the text/HTML table grid — `TableItem.image` was ignored entirely, unlike `MarkdownPictureSerializer`, which already honors `image_mode` (`PLACEHOLDER` / `EMBEDDED` / `REFERENCED`) for pictures.

## Change

- Extracted the shared image-emission logic (previously `MarkdownPictureSerializer._serialize_image_part`) into a module-level function reused by both the picture and table serializers — no behavior change to the existing picture path (verified byte-for-byte identical branches).
- Moved `_image_to_base64` from `PictureItem` up to the shared `FloatingItem` base so `TableItem` (which is also a `FloatingItem`) can use it.
- `MarkdownTableSerializer.serialize` now appends the table image **alongside** the existing text table when `item.image` is set and `image_mode` is not `PLACEHOLDER` — the text table is preserved so this is purely additive.

## Design choice (open in the issue, unconfirmed by a maintainer)

I'd asked in the issue whether the image should *replace* or be emitted *alongside* the text table, and whether it should reuse `image_mode` or a new dedicated option. I went with **alongside + reuse existing `image_mode`**, since:
- it requires no new public API surface, and
- replacing the text table by default would be a breaking output change for existing consumers who already rely on the Markdown table text (e.g. for RAG chunking).

Happy to adjust if a maintainer prefers different behavior.

## Backward compatibility

Tables without a generated image (`item.image is None` — the common case, since table images require the `generate_table_images` pipeline option) see **zero output change**, even if `image_mode` is explicitly set to something other than `PLACEHOLDER`. Covered by a dedicated regression test.

## Testing

Added 3 new tests in `test/test_serialization.py`:
- `test_md_table_with_image_embedded` — `EMBEDDED` mode, image present
- `test_md_table_with_image_referenced` — `REFERENCED` mode, image present (asserts exact output)
- `test_md_table_without_image_unaffected_by_image_mode` — no image present, asserts byte-exact no-op regardless of `image_mode`

Full suite: `uv run pytest -q` → 545 passed, 6 skipped (unrelated to this change).
`ruff check` / `ruff format --check` / `mypy docling_core test` all clean.

### DIFF
--- a/docling_core/transforms/serializer/markdown.py
+++ b/docling_core/transforms/serializer/markdown.py
@@ -427,6 +427,46 @@ class MarkdownAnnotationSerializer(BaseModel, BaseAnnotationSerializer):
         )
 
 
+def _serialize_image_part(
+    item: FloatingItem,
+    doc: DoclingDocument,
+    image_mode: ImageRefMode,
+    image_placeholder: str,
+    **kwargs: Any,
+) -> SerializationResult:
+    """Render a floating item's image per the image mode.
+
+    Shared by the picture and table serializers so both honor image_mode
+    identically for any FloatingItem that carries an image.
+    """
+    error_response = "<!-- 🖼️❌ Image not available. Please use `PdfPipelineOptions(generate_picture_images=True)` -->"
+    if image_mode == ImageRefMode.PLACEHOLDER:
+        text_res = image_placeholder
+    elif image_mode == ImageRefMode.EMBEDDED:
+        # short-cut: we already have the image in base64
+        if isinstance(item.image, ImageRef) and isinstance(item.image.uri, AnyUrl) and item.image.uri.scheme == "data":
+            text_res = f"![Image]({item.image.uri})"
+        else:
+            # get the item.image._pil or crop it out of the page-image
+            img = item.get_image(doc=doc)
+            if img is not None:
+                imgb64 = item._image_to_base64(img)
+                text_res = f"![Image](data:image/png;base64,{imgb64})"
+            else:
+                text_res = error_response
+    elif image_mode == ImageRefMode.REFERENCED:
+        if not isinstance(item.image, ImageRef) or (
+            isinstance(item.image.uri, AnyUrl) and item.image.uri.scheme == "data"
+        ):
+            text_res = image_placeholder
+        else:
+            text_res = f"![Image]({item.image.uri!s})"
+    else:
+        text_res = image_placeholder
+
+    return create_ser_result(text=text_res, span_source=item)
+
+
 class MarkdownTableSerializer(BaseTableSerializer):
     """Markdown-specific table item serializer."""
 
@@ -584,6 +624,18 @@ class MarkdownTableSerializer(BaseTableSerializer):
             if table_text:
                 res_parts.append(create_ser_result(text=table_text, span_source=item))
 
+            # Emit the table image alongside the text grid when one was generated.
+            # Tables without an image (the common case) stay a strict no-op.
+            if item.image is not None and params.image_mode != ImageRefMode.PLACEHOLDER:
+                img_res = _serialize_image_part(
+                    item=item,
+                    doc=doc,
+                    image_mode=params.image_mode,
+                    image_placeholder=params.image_placeholder,
+                )
+                if img_res.text:
+                    res_parts.append(img_res)
+
         text_res = "\n\n".join([r.text for r in res_parts])
 
         return create_ser_result(text=text_res, span_source=res_parts)
@@ -622,7 +674,7 @@ class MarkdownPictureSerializer(BasePictureSerializer):
                 if ann_res.text:
                     res_parts.append(ann_res)
 
-            img_res = self._serialize_image_part(
+            img_res = _serialize_image_part(
                 item=item,
                 doc=doc,
                 image_mode=params.image_mode,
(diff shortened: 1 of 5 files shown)


maxmilian:
Rebased and force-pushed to resolve the conflict. It turned out to be more than a textual conflict: #664 split the monolithic `document.py` into focused modules after this branch was opened, so the original diff no longer applied. I've reworked the change onto the new layout:

- `_image_to_base64` moved from `PictureItem` up to the shared `FloatingItem` base — now in `docling_core/types/doc/items/node.py` (dropping the now-unused `base64`/`BytesIO` imports from `items/picture/picture.py`).
- `_serialize_image_part` extracted from `MarkdownPictureSerializer` into a module-level helper in `transforms/serializer/markdown.py`, called by both the picture and table serializers.
- `MarkdownTableSerializer` now emits the table image alongside the text grid when `item.image` is set and `image_mode` is not `PLACEHOLDER`; tables without an image stay a strict no-op.

The behavior and public surface are unchanged from the original PR. Verified on the new base: `test_serialization.py` (50) plus the doclang/doc/latex serialization suites (220 total) pass, and the three new table-image tests (embedded / referenced / no-op) go red without the change. `ruff check`, `ruff format`, and `mypy` are clean on the touched files.

maxmilian:
A status note on this PR, with the current state of `main` for reference.

**The gap is still present on `main` (`2cae21e`).** In `docling_core/transforms/serializer/markdown.py`, `_serialize_image_part` (L651) is still only reached from `MarkdownPictureSerializer.serialize` (call site at L625). `MarkdownTableSerializer.serialize` (L509–589, class at L430) emits captions, annotations and the tabulate grid only — there is no image path for tables. So the behaviour this PR adds is still unimplemented.

**Linked issue #474 is still open**, with two users describing concrete needs there:

> besides the use cases already mentioned upstream, another one is for RAG pipelines. More complicated tables cannot faithfully be represented using markdown format, so it would be great to either have the option to use a VLM before ingestion to describe the table, or afterwards on retrieval to parse the table to extract the requested information.
> — erikvullings, Feb 4

> Hi, just wanted to know if there's any update on this? Thanks a lot!
> — romitsenmi, Feb 18

**On the current check state:** `code-checks` passes on 3.10–3.14, `build-package`/`test-package` pass, DCO passes, and `codecov/patch` is green. The only red check is `Mergify Merge Protections` → "Require two reviewer for test updates", which is waiting on approvals rather than flagging a problem with the change. No review comments have been left on the PR so far.

Happy to rebase, split this into smaller commits, or adjust the API shape if any of that would make it easier to review.

maxmilian:
Rebased onto current `main` (`a727420`, v2.95.0). This one wasn't optional: #739 renamed `test/` to `tests/`, so the branch as it stood would have resurrected a `test/` directory on merge, and #737 (PEP 604 type hints) touched both `markdown.py` and `node.py`. The rebase applied cleanly and the diff is unchanged in shape — 4 files, +141/−57.

Verified on the new base: `ruff check` and `ruff format --check` clean on the touched files, and `pytest tests/test_serialization.py` → 98 passed, including the three tests this PR adds (embedded / referenced / no-op when the table has no image).

The gap itself is still open on `main`: `MarkdownTableSerializer` (`markdown.py` L525–686, `serialize` at L604) contains no image handling at all, while `_serialize_image_part` remains private to `MarkdownPictureSerializer` (L758) and `_image_to_base64` still sits on `PictureItem` (`items/picture/picture.py` L145) rather than on the shared `FloatingItem` base (`items/node.py` L204). Issue #474 is still open, with two users describing concrete use cases there.

The one thing I'd still like a maintainer's call on — I asked in #474 back in June and it's the only real open question left:

1. Should the table image be emitted **alongside** the Markdown table text, or **replace** it?
2. Should it reuse the existing `image_mode`, or sit behind a dedicated option (e.g. `table_image_mode`)?

This PR currently does *alongside + reuse `image_mode`*, because that adds no public API surface and replacing the text table by default would be a breaking output change for consumers that chunk the Markdown grid (RAG pipelines). Both alternatives are small changes from here if you'd prefer either.

If the answer is "we don't want this in core", that's a perfectly fine outcome — just say so and I'll close it. Otherwise I'll keep it mergeable.

PR #681, opened 2026-07-09
docs(page): add Field descriptions to page primitives

### LINKED ISSUE #411: Request: putting `Field(description="...")` on primitives
I am trying to implement my own driver for the newly released [Nvidia `nemotron-parse`](https://build.nvidia.com/nvidia/nemotron-parse) and am finding a lot of page primitives to be unclear:

- [`PdfPageGeometry`](https://github.com/docling-project/docling-core/blob/v2.49.0/docling_core/types/doc/page.py#L455-L464) -- what is `art_bbox`, `bleed_bbox`, and `crop_bbox`, etc.
- [`SegmentedPage`](https://github.com/docling-project/docling-core/blob/v2.49.0/docling_core/types/doc/page.py#L485-L503) -- what is `textline_cells` and `char_cells`
- [`TextCell`](https://github.com/docling-project/docling-core/blob/v2.49.0/docling_core/types/doc/page.py#L280-L291) -- what is `orig` and `confidence`

The request is, putting Pydantic `description`s with commonly-used data structures will go a long way in helping developers with Docling.

maxmilian:
Closes #411.

@jamesbraza asked for Pydantic `description`s on the page primitives that are hard to interpret when writing a parser driver. @PeterStaar-IBM offered to take it back in October; since that never landed, I have picked it up (asked on the issue first).

## What

Adds `Annotated[..., Field(description=...)]` to the three primitives named in the issue, in `docling_core/types/doc/page.py`:

- **`TextCell`** — `rect`, `text`, `orig`, `text_direction`, `confidence`, `from_ocr`
- **`PdfPageGeometry`** — `boundary_type`, `art_bbox`, `bleed_bbox`, `crop_bbox`, `media_bbox`, `trim_bbox`
- **`SegmentedPage`** — all 11 fields, including the `has_chars` / `has_words` / `has_lines` flags

This matches the `Annotated[T, Field(description=...)]` style already used throughout `document.py`. **Metadata only — no behavior change.**

The five PDF boxes are described with their spec meanings. For `crop_bbox` and `boundary_type` I also noted the behavior that surprised me while reading the code: `width`, `height` and `origin` are always computed from `crop_bbox`, regardless of `boundary_type`.

## Two deliberate omissions

- **`PageGeometry.angle`** (inherited by `PdfPageGeometry`) is left undocumented. I could not establish its unit from the code: `BoundingRectangle.angle` is a property documented as radians, but a PDF `/Rotate` is conventionally degrees, and nothing in the repo pins down which one this field carries. I would rather leave it blank than guess — happy to add it if you tell me which it is.
- `TextCell.index` / `rgba` come from `OrderedElement` / `ColorMixin` and are out of scope for this issue.

## One open question

For `TextCell.orig` I wrote *"Original, untreated text of the cell as produced by the extraction backend, before any downstream normalization applied to `text`."*

That mirrors `TextItem.orig` in `document.py` (`# untreated representation`). Worth noting that every producer I checked in `docling` — `pypdfium2_backend`, `mets_gbs_backend`, `tesseract_ocr_model`, `easyocr_model` — currently sets `text` and `orig` identically, so the description states the intent rather than an observable difference. Let me know if you would phrase it differently.

## Verification

- `pytest test/` — 546 passed, 6 skipped
- `prek run` — Ruff linter, Ruff formatter, MyPy, Pytest, Docs all pass (the `Docs` hook confirms `docs/DoclingDocument.json` does not drift; these page primitives are not part of that schema)
- Confirmed the descriptions actually reach the generated schema, e.g. `SegmentedPage.model_json_schema()` now carries a description on all 11 properties.

Draft until the `orig` wording is confirmed.

### DIFF
--- a/docling_core/types/doc/page.py
+++ b/docling_core/types/doc/page.py
@@ -280,15 +280,46 @@ class TextDirection(str, Enum):
 class TextCell(ColorMixin, OrderedElement):
     """Model representing a text cell with positioning and content information."""
 
-    rect: BoundingRectangle
+    rect: Annotated[
+        BoundingRectangle,
+        Field(description="Oriented rectangle enclosing the cell on the page."),
+    ]
 
-    text: str
-    orig: str
+    text: Annotated[str, Field(description="Text content of the cell.")]
+    orig: Annotated[
+        str,
+        Field(
+            description=(
+                "Original, untreated text of the cell as produced by the "
+                "extraction backend, before any downstream normalization "
+                "applied to `text`."
+            )
+        ),
+    ]
 
-    text_direction: TextDirection = TextDirection.LEFT_TO_RIGHT
+    text_direction: Annotated[
+        TextDirection,
+        Field(description="Reading direction of the text within the cell."),
+    ] = TextDirection.LEFT_TO_RIGHT
 
-    confidence: float = 1.0
-    from_ocr: bool
+    confidence: Annotated[
+        float,
+        Field(
+            description=(
+                "Confidence of the text extraction. Cells read directly from "
+                "digital text keep the default of 1.0; OCR backends set the "
+                "recognition score reported by the OCR engine."
+            )
+        ),
+    ] = 1.0
+    from_ocr: Annotated[
+        bool,
+        Field(
+            description=(
+                "Whether the cell was produced by OCR rather than extracted from the document's digital text layer."
+            )
+        ),
+    ]
 
     @field_serializer("confidence")
     def _serialize(self, value: float, info: FieldSerializationInfo) -> float:
@@ -535,13 +566,48 @@ class PageGeometry(BaseModel):
 class PdfPageGeometry(PageGeometry):
     """Extended dimensions model specific to PDF pages with boundary types."""
 
-    boundary_type: PdfPageBoundaryType
+    boundary_type: Annotated[
+        PdfPageBoundaryType,
+        Field(
+            description=(
+                "The page boundary that `rect` was derived from. Note that "
+                "`width`, `height` and `origin` are always computed from "
+                "`crop_bbox`, independently of this value."
+            )
+        ),
+    ]
 
-    art_bbox: BoundingBox
-    bleed_bbox: BoundingBox
-    crop_bbox: BoundingBox
-    media_bbox: BoundingBox
-    trim_bbox: BoundingBox
+    art_bbox: Annotated[
+        BoundingBox,
+        Field(description=("PDF ArtBox: the extent of the page's meaningful content as intended by its creator.")),
+    ]
+    bleed_bbox: Annotated[
+        BoundingBox,
+        Field(
+            description=(
+                "PDF BleedBox: the clipping region for production output, "
+                "including any bleed area beyond the trim boundary."
+            )
+        ),
+    ]
+    crop_bbox: Annotated[
+        BoundingBox,
+        Field(
+            description=(
+                "PDF CropBox: the region the page contents are clipped to "
+                "when displayed or printed. This is the box `width`, `height` "
+                "and `origin` are computed from."
+            )
+        ),
+    ]
+    media_bbox: Annotated[
+        BoundingBox,
+        Field(description=("PDF MediaBox: the full extent of the physical medium the page is to be printed on.")),
+    ]
+    trim_bbox: Annotated[
+        BoundingBox,
+        Field(description=("PDF TrimBox: the intended dimensions of the page after trimming.")),
+    ]
 
     @property
     def width(self):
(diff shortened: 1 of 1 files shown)


PeterStaar-IBM (dismissed):
lgtm!

maxmilian:
Not trying to rush the queue — just flagging the merge state in case it reads as more blocked than it is.

Since this PR touches no test data, the `Require two reviewer for test updates` protection doesn't apply to it: Mergify reports **merge protection satisfied** with @PeterStaar-IBM's approval alone. CI is 11/11 and it's still MERGEABLE.

So whenever it fits your schedule, this one is good to go as-is. Happy to rebase if it drifts.

docling_core/types/doc/page.py
```diff
@@ -565,25 +631,76 @@ def origin(self):
...
-    widgets: list[PdfWidget] = []
-    hyperlinks: list[PdfHyperlink] = []
+    has_chars: Annotated[
+        bool,
+        Field(
+            description=(
+                "Whether `char_cells` has been computed. Distinguishes an "
+                "empty list meaning 'no content' from one meaning "
+                "'not computed'."
+            )
```
ceberam, line 668: I still see some ambiguity and, to be honest, the flag name does not help either. Read in isolation, the name `has_chars` seems to promise a content claim (_this page has characters_), but the intended semantics (per the comment) is a process claim (_character extraction was attempted_). Those are opposite when the list is empty: `has_chars = True` + `char_cells = []` is supposed to mean _computed, found nothing_, but a reader would naturally read it as _there are chars, so why is the list empty?_.
Since we cannot change the flag names for backwards compatibility, I would be more explicit with the description. What do you think about something like:

```suggestion
            description=(
                "Whether character extraction has been computed. When False, `char_cells` being empty means extraction"
                " was not attempted; when True, `char_cells` being empty means the page had no extractable characters."
            )
```

docling_core/types/doc/page.py
```diff
@@ -535,13 +566,48 @@ def origin(self):
...
-    media_bbox: BoundingBox
-    trim_bbox: BoundingBox
+    boundary_type: Annotated[
+        PdfPageBoundaryType,
+        Field(
+            description=(
+                "The page boundary that `rect` was derived from. Note that "
+                "`width`, `height` and `origin` are always computed from "
+                "`crop_bbox`, independently of this value."
+            )
```
ceberam, line 576: This correctly describes the behavior in `PdfPageGeometry`, but I would be more explicit to avoid any misunderstanding with `PageGeometry`, since in this parent class those same three properties read from `self.rect`, not `crop_bbox`. I would scope it as follows:

> The page boundary that `rect` was derived from. In `PdfPageGeometry`, `width`, `height` and `origin` are computed from `crop_bbox` regardless of this value.

docling_core/types/doc/page.py
```diff
@@ -565,25 +631,76 @@ def origin(self):
...
-    textline_cells: list[TextCell] = []
+    dimension: Annotated[
+        PageGeometry,
+        Field(description="Geometry of the page this segmentation belongs to."),
+    ]
+
+    bitmap_resources: Annotated[
+        list[BitmapResource],
+        Field(description="Bitmap images embedded in the page."),
+    ] = []
```
ceberam, line 642: The description may be misleading since _embedded_ only describes one of the three possible states of a `BitmapResource` (check the 3 possible values of `ImageRefMode` in `mode` field).

I would rephrase it as:

> Bitmap image resources found on the page

Alternatively, we could be more explicit:

> Bitmap image resources found on the page. Each entry may carry the actual image data (embedded as base64), a URI reference to an external file, or act as a position-only placeholder."


What do you think? (I still prefer the first option, for simplicity).

ceberam (changes requested):
@maxmilian Thanks for providing this documentation.
I went through it and I found some few gaps that I think would be worth addressing, to make sure there are no fields that are left _unclear_.

docling_core/types/doc/page.py
maxmilian: Agreed — the process-vs-content reading is exactly the trap, and your wording makes it explicit. Applied your suggestion, and mirrored the same phrasing onto `has_words` / `has_lines` so the three flags stay consistent.

docling_core/types/doc/page.py
maxmilian: Good catch — I described `PdfPageGeometry` behavior without scoping it, and `PageGeometry` reads those three from `self.rect`. Applied your wording verbatim.

docling_core/types/doc/page.py
maxmilian: You're right that "embedded" only covers one of the three `ImageRefMode` states. Took your first option for simplicity.

ceberam (approved):
🚀

PR #682, opened 2026-07-10
fix(doclang): enforce size limits when extracting .dclx archives

vagenas:
- Cap per-member and total uncompressed size during safe archive extraction
- Apply limits to bytes actually read, not only declared ZipInfo sizes
- Expose the limits on `load_from_doclang_archive` and cover them with unit tests


dolfim-ibm (approved):
lgtm

PeterStaar-IBM (approved):

PR #684, opened 2026-07-14
fix(doclang): enforce image size budgets for archive assets and ImageRef

vagenas:
- Cap `.dclx` `pages/` and `assets/` rasters with `max_image_decoded_size` before Pillow opens them
- Apply the same limit to Path and `file://` ImageRef loads
- Add tests covering oversize archive images and path-based ImageRefs


dolfim-ibm (approved):
lgtm

cau-git (approved):

PR #692, opened 2026-07-23
fix: accept application/epub+zip mimetype in DocumentOrigin

### LINKED ISSUE #691: DocumentOrigin rejects `application/epub+zip`, breaking EPUB conversion/reload on hosts without it in the system mime map
### Summary

`DocumentOrigin.validate_mimetype` accepts a mimetype only if it is in Python's
stdlib `mimetypes.types_map` **or** the hardcoded `_extra_mimetypes` allowlist.
`application/epub+zip` is in neither. It therefore only validates on hosts whose
system mime configuration (e.g. `/etc/mime.types`) happens to include EPUB. On
hosts without it (Windows, slim Docker images, minimal containers) it raises.

`application/epub+zip` is docling's own declared EPUB mimetype
(`InputFormat.EPUB: ["application/epub+zip"]` in `docling/datamodel/base_models.py`)
and is emitted by the EPUB backend, so on such hosts **both EPUB conversion and
reloading a saved EPUB `DoclingDocument` fail.**

### Reproduction

On a host whose system mime map does not list EPUB (e.g. Windows), with only
`docling-core` installed:

```python
from docling_core.types.doc.common.origin import DocumentOrigin

DocumentOrigin(filename="book.epub", mimetype="application/epub+zip", binary_hash=0)
# pydantic.ValidationError: Value error, 'application/epub+zip' is not a valid MIME type
```

`application/pdf` and `text/markdown` construct fine — only `application/epub+zip`
is rejected. The same error surfaces via `DoclingDocument.load_from_json(...)` on
any saved EPUB document.

### Root cause

`application/epub+zip` is missing from `_extra_mimetypes` in
`docling_core/types/doc/common/origin.py`. That allowlist exists specifically so
validation does not depend on the host mime map (it already lists markdown, csv,
asciidoc, and the office formats, none of which are reliably present in system
maps). EPUB was simply omitted.

### Proposed fix

Add `"application/epub+zip"` to `_extra_mimetypes`. With it added, EPUB documents
construct, load, and export correctly.

_(Broader note: because the check falls back to the host's `mimetypes.types_map`,
mimetype validation is currently host-dependent for any type not in the allowlist.
The minimal fix above resolves EPUB and matches the existing convention.)_

### Environment

- docling-core 2.87.1 (also confirmed present on `main`)
- Python 3.12, Windows

I'm happy to open a PR with the one-line fix and a unit test.

manticore999:
Adds `application/epub+zip` to `DocumentOrigin._extra_mimetypes` so EPUB documents
validate on hosts whose system mime map does not list EPUB (Windows, minimal
containers). This is docling's declared EPUB mimetype, previously rejected by
`validate_mimetype`. Follows the same convention as #654 (OpenDocument formats),
and adds a unit test (`test/test_origin.py`).

Resolves #691

### DIFF
--- a/docling_core/types/doc/common/origin.py
+++ b/docling_core/types/doc/common/origin.py
@@ -25,6 +25,7 @@ class DocumentOrigin(BaseModel):
     )
 
     _extra_mimetypes: typing.ClassVar[list[str]] = [
+        "application/epub+zip",
         "application/vnd.box.boxnote",
         "application/vnd.oasis.opendocument.presentation",
         "application/vnd.oasis.opendocument.spreadsheet",


ceberam (changes requested):
@manticore999 Thanks for the rapid response. I don’t think we need a new test module for that change. Please, remove it.

manticore999:
Done, removed the test module. Thanks ! @ceberam

ceberam (approved):

PR #695, opened 2026-07-27
fix: preserve literal tag-shaped text in OTSL cells

### LINKED ISSUE #694: DocLang deserializer misclassifies bracked-delimited OTSL text as XML
### Bug

`DocLangDocDeserializer` decides whether OTSL cell content is nested XML from
the content's string syntax. Consequently, valid literal text that starts with
`<` and ends with `>` can either fail deserialization or be silently converted
into a `RichTableCell`.

### Reproducer

```python
from docling_core.transforms.deserializer.doclang import DocLangDocDeserializer

xml = """\
<doclang version="0.7">
  <table>
    <fcel/><content><![CDATA[<none>]]></content><nl/>
  </table>
</doclang>
"""

DocLangDocDeserializer().deserialize_str(xml)
```

### Actual behavior

```text
ValueError: Invalid DocLang XML: mismatched tag: line 1, column 14
```

The XML shown in the error is an internally constructed fragment:

```xml
<root><none></root>
```

There is also a silent form of the same bug:

```python
cases = {
    "self-closing-looking literal": "<none/>",
    "known-tag-looking literal": "<text>not literal</text>",
}

for name, literal in cases.items():
    xml = (
        '<doclang version="0.7"><table><fcel/>'
        f"<content><![CDATA[{literal}]]></content>"
        "<nl/></table></doclang>"
    )
    cell = DocLangDocDeserializer().deserialize_str(xml).tables[0].data.grid[0][0]
    print(name, type(cell).__name__, repr(cell.text))
```

Output:

```text
self-closing-looking literal RichTableCell '<none/>'
known-tag-looking literal RichTableCell 'not literal'
```

The `<none/>` case additionally creates an empty `GroupItem` and points the
cell at it.

These inputs are well-formed and pass the reference DocLang validator.
Additional valid literals that trigger the hard failure include `<space>`,
`<Month number>`, `<specify >`, and `<Country code>`.

### Expected behavior

Content originating from a DOM text node should remain literal cell text,
regardless of whether its characters resemble XML:

- `<none>` should produce a regular `TableCell` with text `<none>`.
- `<none/>` should produce a regular `TableCell` with text `<none/>`.
- `<text>not literal</text>` inside CDATA should remain that exact literal
  string.

An actual element remains semantically different and should continue to
produce a rich cell:

```xml
<doclang version="0.7">
  <table>
    <fcel/><text>not literal</text><nl/>
  </table>
</doclang>
```

### Cause

`_otsl_extract_tokens_and_text()` flattens both DOM `Text` nodes and DOM
`Element` nodes into strings. `_otsl_parse_texts()` later tries to reconstruct
the lost distinction by treating every string that starts with `<` and ends
with `>` as XML and reparsing it under a synthetic `<root>`.

At that point these are indistinguishable:

```text
Text node value:       <text>not literal</text>
Element serialization: <text>not literal</text>
```

Catching only the fragment parse error would not fix the silent `<none/>` and
`<text>not literal</text>` cases. The node origin needs to be preserved through
the OTSL mixed stream, and only actual element nodes should be dispatched as
nested DocLang.

### Version

`docling-core==2.88.0`

cau-git:
Closes #694.

## Summary

- Preserve DOM node origin through the OTSL mixed-content stream.
- Treat tag-shaped DOM text, including strings such as `<none/>` and `<fcel/>`, as literal table-cell text.
- Dispatch actual nested DocLang elements directly, retaining rich cells and nested tables without reparsing serialized fragments.
- Read OTSL location elements directly instead of serializing and reparsing them.

## Root cause

The deserializer flattened DOM text and DOM elements into strings, then inferred whether a cell part was XML from its leading and trailing characters. That rejected some valid literal text and silently promoted other literals to rich content or structural OTSL tokens.

The mixed stream now retains `str | Element`, so interpretation follows the DOM node type rather than the text syntax.

## Validation

- `.venv/bin/pytest -q test/test_deserializer_doclang.py` — 101 passed
- Ruff check and format check on both changed files
- MyPy on both changed files
- Validator-approved nested-table deserialize and round-trip probe

### DIFF
--- a/docling_core/transforms/deserializer/doclang.py
+++ b/docling_core/transforms/deserializer/doclang.py
@@ -1763,16 +1763,13 @@ class DocLangDocDeserializer(BaseDocDeserializer, BaseModel):
         }
     )
 
-    def _bbox_from_location_text_fragments(
-        self, *, doc: DoclingDocument, fragments: list[str]
-    ) -> Optional[BoundingBox]:
-        r"""Build a TOPLEFT bbox from four ``<location value=\"...\"/>`` XML fragments."""
-        if len(fragments) != 4:
+    def _bbox_from_location_elements(self, *, doc: DoclingDocument, locations: list[Element]) -> Optional[BoundingBox]:
+        """Build a TOPLEFT bbox from four location elements."""
+        if len(locations) != 4:
             return None
         values: list[int] = []
         res_for_group: Optional[int] = None
-        for fragment in fragments:
-            loc_el = self._parse_xml_string(fragment)
+        for loc_el in locations:
             if loc_el.tagName != DocLangToken.LOCATION.value:
                 return None
             try:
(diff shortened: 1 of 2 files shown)


ceberam (approved):
LGTM

Matteo-Omenetti (approved):

PR #696, opened 2026-07-28
feat: `dclq` tool

cau-git:
## Summary

This PR brings the DocLang query CLI into the `docling-core` repository as **`dclq`**. The tool started as `dlgrep`, but its scope now extends beyond grep: it can search, enumerate, inspect, navigate, and retrieve semantic document structure while returning reusable XPath addresses.

Nothing has been published to PyPI yet, so `dclq` is introduced without a compatibility package or console-script alias for either earlier working name.

```console
$ dclq grep 'GPU|CPU' paper.dclg \
    --within-xpath '/heading[13]' --section -n
/list[4]/ldiv[1]:- AWS EC2 VM ... Nvidia L4 GPU ...
/text[42]:All experiments ... GPU acceleration ... x86 CPU ...

$ dclq list paper.dclg --type table_cell --page 3
```

## CLI scope

`dclq` queries semantic units—headings, paragraphs, list items, table cells, fields, captions, formulas, code, pictures, and metadata—rather than XML lines.

| Command | Purpose |
| --- | --- |
| `dclq grep PATTERN INPUT...` | Search semantic units with grep-compatible matching and exit codes |
| `dclq list INPUT...` | Enumerate semantic units without an empty-pattern workaround |
| `dclq inspect INPUT...` | Report raw and semantic document structure |
| `dclq outline INPUT` | Return the heading hierarchy with XPath addresses |
| `dclq show INPUT XPATH` | Retrieve semantic content at an XPath, optionally as a section or with context |
| `dclq select INPUT XPATH` | Evaluate XPath against the original DocLang XML |

`grep` is now an explicit subcommand; bare `dclq PATTERN INPUT` is a usage error. `list` reuses the existing filtering, projection, bounding, summary, and output machinery without introducing a second query implementation.

The CLI supports:

- `.dclg`, `.xml`, `.dclx`, and standard input
- regex, fixed-string, case-insensitive, and whole-word search
- semantic `-A` / `-B` / `-C` context
- `document`, `section`, `container`, and `auto` context scopes
- XPath, section, page, layer, semantic-type, offset, and limit filters
- bounded text, JSON, and JSONL output
- source-backed records with `xpaths`, `logical_type`, `text`, `pages`, and contributing `doc_items`
- table-cell header context, sections, threaded fragments, grep-style human output, and reusable `show` / `select` handoffs

The old `page` context scope was removed; `--page` remains available for explicit page filtering.

## Changes to docling-core

The DocLang deserializer now accepts an optional `source_map` recorder and exports `DocLangSourceMap` and `DocLangSourceTarget` from `docling_core.transforms.deserializer`.

Bindings are collected while parsing, when source identity is authoritative, for ordinary items, headings, list anchors, table cells, fields, pictures, page breaks, and threaded fragments. This allows `dclq` to map semantic `DoclingDocument` items back to canonical source XPaths without text matching or post-hoc reconstruction. Deserializing without a source map keeps the existing document result unchanged.

The source-mapping recorder is isolated in `doclang_source_mapping.py`; `dclq` remains a package-local consumer of `DoclingDocument._hierarchize()` rather than expanding the public document API solely for this tool.

## Packaging and release

`dclq` lives at `packages/dclq` as a uv workspace member and builds its own `dclq` wheel. It is not included in the `docling-core` wheel, and the runtime dependency remains one-way:

```text
dclq == <version>
  -> docling-core[dclq] == <same version>
```

The `docling-core[dclq]` extra supplies the CLI-specific DocLang, Typer, Click, and lxml dependencies. `dclq` is versioned and released in lockstep with `docling-core` because it intentionally consumes package-local internals.

Release automation uses native uv operations:

- `uv version` updates both package versions
- `uv add "docling-core[dclq]==..."` maintains the exact pin and extra
- `uv sync --frozen --all-extras --all-packages` validates the checked-in lockfile
- both wheels are built into separate distribution directories
- CI clean-installs the locally built wheel pair together, which makes a mismatched exact pin unsatisfiable
- PyPI publishes `docling-core` first and `dclq` second

No custom lockstep checker or propagation sleep is needed; the built-artifact installation is the release invariant.

## Repository integration

- imported history is preserved in this branch
- Ruff, MyPy, pytest, and formatting configuration are inherited from the repository root
- pre-commit runs the `dclq` tests and the root checks
- CI builds, inspects, installs, and smoke-tests both wheels
- README examples cover grep, listing, semantic context, section retrieval, raw XPath, JSON, and shell pipelines
- the README carries an explicit experimental warning and the two historical plans are replaced by one direct `dclq` implementation plan

## Validation

The branch passes the repository commit gates:

- Ruff lint and formatting
- MyPy
- docling-core pytest suite
- focused `dclq` pytest suite
- docs generation
- `uv lock` validation
- secret scanning

Both `docling-core` and `dclq` wheels were built and installed together into a clean environment. The installed `dclq` console script and `python -m dclq` entry point were exercised, including empty pattern files, empty paginated listings, explicit subcommand routing, structured output, source addresses, and surviving context scopes.


packages/dlgrep/pyproject.toml
```diff
@@ -0,0 +1,61 @@
...
+]
+requires-python = ">=3.10,<4.0"
+dependencies = [
+    # Pinned exactly: dlgrep rides on docling-core internals and is released from
+    # the same commit, in lockstep. Bumped by .github/scripts/release.sh.
+    "docling-core==2.88.0",
+    "doclang>=0.7,<0.8",
+    "typer>=0.15.1,<0.25.0",
+    "click>=8.0.0",
+    "lxml>=6.0.2",
```
dolfim-ibm: should we move this to a `docling-core` extra and just make the dlgrep depend on it?

cau-git: Done.

dolfim-ibm (dismissed):

packages/dclq/dclq/document.py
```diff
@@ -0,0 +1,594 @@
...
+            with zipfile.ZipFile(path) as archive:
+                infos = archive.infolist()
+                for info in infos:
+                    member = PurePosixPath(info.filename.replace("\\", "/"))
+                    if member.is_absolute() or ".." in member.parts:
+                        raise DclqError(f"unsafe archive member: {info.filename!r}")
+                    if info.file_size > MAX_ARCHIVE_MEMBER_BYTES:
+                        raise DclqError(f"archive member exceeds the size limit: {info.filename!r}")
+                if sum(info.file_size for info in infos) > MAX_ARCHIVE_BYTES:
+                    raise DclqError("archive exceeds the total uncompressed size limit")
```
ceberam, line 475: Security concern: the current guard would prevent an in-memory bomb only if the attacker reports the true sizes. A well-crafted zip with falsely-reported small file_size entries but huge decompressed content can exhaust memory before the check triggers.
Mitigation: Read with an intermediate size cap using archive.open() + read up to MAX_ARCHIVE_MEMBER_BYTES + 1 bytes, then check length.

packages/dclq/dclq/cli.py
```diff
@@ -0,0 +1,1097 @@
...
+
+    if args.format == "xml":
+        for record in records:
+            print(record.get("xml", record.get("value", "")))
+    elif args.format == "text":
+        for record in records:
+            if "xpaths" in record:
+                selected_xpath = record["xpaths"][0]
+                if not isinstance(selected_xpath, str):
+                    raise DclqError("invalid internal XPath result")
```
ceberam: Dead code that can be confusing: `selected_xpath` is always `record["xpaths"][0]` which is always a str (set from `_canonical_xpath()`).

ceberam (commented):
Just a couple of comments.
For later improvement: leverage an AI assistant to generate docstrings for the public methods and constants of `LoadedDocument`.

PeterStaar-IBM (approved):
lgtm!

dolfim-ibm (approved):

PR #698, opened 2026-07-30
fix(markdown): escape referenced figures' filepaths and ensure cross-platform path handling (Windows / POSIX)

the-zucc:
#182 caused a regression in which the behavior introduced in #122 was removed.

I'm therefore proposing that we re-add the change, slightly modified for both POSIX and Windows path support.

One edge-case identified is the obvious case where the destination path is a filename containing a backslash. I've mentioned the limitation in the docstring.

Partial fix for docling-project/docling#3617


docling_core/transforms/serializer/markdown.py
```diff
@@ -687,12 +688,79 @@ def _serialize_image_part(
...
+        """
+
+        # Characters that survive percent-encoding in a link destination.
+        # The RFC 3986 reserved characters that carry meaning in a URI, plus ``%`` so that an
+        # already-encoded destination is not encoded a second time. Whitespace and parentheses
+        # are deliberately absent: they would end a Markdown inline link.
+        _URI_KEEP_CHARS: str = "/%:@+,;=~$!&'*"
+
+        # Matches the drive prefix of an absolute Windows path, e.g. ``C:/``.
+        _WINDOWS_DRIVE_RE: re.Pattern = re.compile(r"[A-Za-z]:/")
```
ceberam: `_URI_KEEP_CHARS` and `_WINDOWS_DRIVE_RE` are defined inside the method body and both constants are reconstructed on every call. You can move them out, at class level. You can also annotate them as `Final`.

test/test_serialization.py
```diff
@@ -1103,3 +1106,149 @@ def test_html_meta_emits_xhtml_compatible_attributes():
     assert 'data-meta-name="entities"' in html_out
     # Output must be parseable by a strict XML parser.
     ET.fromstring(html_out)
+
+
+# A link destination to encode, paired with its expected encoding.
+_EscapeCase = tuple[Union[AnyUrl, PurePath], str]
```
ceberam: Minor comment: use `TypeAlias` as per PEP 613.

docling_core/transforms/serializer/markdown.py
```diff
@@ -687,12 +688,79 @@ def _serialize_image_part(
...
+            value: The URL or path to encode.
+
+        Returns:
+            A percent-encoded Markdown link destination.
+        """
+
+        # Characters that survive percent-encoding in a link destination.
+        # The RFC 3986 reserved characters that carry meaning in a URI, plus ``%`` so that an
+        # already-encoded destination is not encoded a second time. Whitespace and parentheses
+        # are deliberately absent: they would end a Markdown inline link.
```
ceberam: Could you please:
- tighten the prose to reduce verbosity
- instead of inline comments, add this information as docstrings for these constants in google style

docling_core/transforms/serializer/markdown.py
```diff
@@ -687,12 +688,79 @@ def _serialize_image_part(
...
+        resolves when it is exported on POSIX.
+
+        The only destination this gives a ``file://`` scheme to is an absolute Windows
+        path, where it is the sole spelling a renderer cannot misread as a URL scheme
+        (``C:``) or as an authority (``//server``). A URL that already carries the
+        scheme is passed through, since dropping it would turn an absolute filesystem
+        reference into a root-relative URL.
+
+        Known limitation: behavior is unknown if value is a POSIX filename, containing
+        an actual backslash in the filename.
```
ceberam: The behavior is wrong but deterministic: the backslash is misinterpreted as a separator. But this is harmless in practice. Please, accept this suggestion:


```suggestion
        Known limitation: a backslash in a `PosixPath` string is ambiguous.
        It may be a Windows separator surviving a JSON round-trip (correct to
        convert) or a literal filename character (where converting it to `/`
        would split one component into two). The two cases are indistinguishable
        from `str()`. In practice this is not a concern because `ImageRef.uri`
        is always populated from native filesystem operations, so a `PosixPath`
        can only carry a literal backslash if the caller explicitly constructed one.
```

ceberam (changes requested):
Thanks @the-zucc for suggesting this PR and resolve the regression!

I've just added some few minor comments.
Please, rebase to `main` to solve the conflict with `test/test_serialization.py`

ceberam (approved):
@the-zucc To accelerate the development, I have rebased to `main` and applied the changes suggested on my comments. We are ready to merge 🚀 
Thanks for your contribution!

cau-git (approved):

the-zucc:
Thanks folks, was a crazy couple of weeks and couldn’t find the time.

Thank you for reviewing and merging!

PR #701, opened 2026-07-30
fix: html serialization of lists with mixed markers

### LINKED ISSUE #699: HTML serializer introduces browser-generated markers for list items without a detected marker
## Description

The HTML serializer can introduce list markers that are not present in the `DoclingDocument`.

This occurs when a `ListGroup` contains a mixture of:

- list items with a detected marker, such as `enumerated=True` and `marker="1."`; and
- list items without a detected marker, such as `enumerated=False` and `marker=""`, where the original numbering remains part of `text`.

A common example is a table of contents where first-level entries are recognized as enumerated list items, but subsection entries such as `1.1` are retained as ordinary text.

The generated list container is selected using only `ListGroup.first_item_is_enumerated(doc)`. If the first item is enumerated, the complete group is emitted as an `<ol>`.

List items without an explicit marker are then emitted as plain `<li>` elements. The browser consequently generates an ordered-list marker for those items, even though no such marker exists in the corresponding `ListItem.marker` field.

This can display additional numbering that was neither detected by Docling nor present in the serialized item text.

## Environment

- `docling-core==2.88.0`
- Python 3.12
- HTML serialization through `HTMLDocSerializer`

## Minimal reproduction

```python
from docling_core.transforms.serializer.html import (
    HTMLDocSerializer,
    HTMLParams,
)


doc = DoclingDocument(name="mixed-list")

list_group = doc.add_list_group(name="table-of-contents")

doc.add_list_item(
    text="Purpose",
    orig="1. Purpose",
    enumerated=True,
    marker="1.",
    parent=list_group,
)

doc.add_list_item(
    text="1.1 Scope",
    orig="1.1 Scope",
    enumerated=False,
    marker="",
    parent=list_group,
)

doc.add_list_item(
    text="Operation",
    orig="2. Operation",
    enumerated=True,
    marker="2.",
    parent=list_group,
)

for show_original_marker in (True, False):
    result = HTMLDocSerializer(
        doc=doc,
        params=HTMLParams(
            show_original_list_item_marker=show_original_marker,
            prettify=False,
        ),
    ).serialize()

    print(
        f"show_original_list_item_marker="
        f"{show_original_marker}"
    )
    print(result.text)
```

## Actual HTML with `show_original_list_item_marker=True`

```html
<ol>
<li style="list-style-type: '1. ';">Purpose</li>
<li>1.1 Scope</li>
<li style="list-style-type: '2. ';">Operation</li>
</ol>
```

Depending on browser styling, the user sees content equivalent to:

```text
1. Purpose
2. 1.1 Scope
2. Operation
```

The `2.` before `1.1 Scope` is generated by the browser because the item is inside an `<ol>` and has no explicit `list-style-type`.

## Actual HTML with `show_original_list_item_marker=False`

```html
<ol>
<li>Purpose</li>
<li>1.1 Scope</li>
<li>Operation</li>
</ol>
```

This still generates ordered-list markers in the browser:

```text
1. Purpose
2. 1.1 Scope
3. Operation
```

Setting `show_original_list_item_marker=False` therefore does not prevent additional browser-generated numbering. It only prevents the serializer from setting the explicit `list-style-type` based on `item.marker`.

## Expected behavior

HTML serialization should be able to preserve the visible content represented by the `DoclingDocument` without adding markers that were not detected.

For the example above, the intended visible output is:

```text
1. Purpose
1.1 Scope
2. Operation
```

The rules should be:

1. If `item.marker` contains a detected marker, render that marker.
2. If `item.marker` is empty, do not introduce a browser-generated marker.
3. Preserve `item.text` unchanged.
4. Do not attempt to infer a missing marker from `item.text` during serialization.

One possible expected HTML representation is:

```html
<ol>
<li style="list-style-type: '1. ';">Purpose</li>
<li style="list-style-type: none;">1.1 Scope</li>
<li style="list-style-type: '2. ';">Operation</li>
</ol>
```

This retains the detected markers while preventing the browser from generating a marker for the item whose `marker` field is empty.

## Relevant implementation behavior

`HTMLListSerializer.serialize()` selects the list container based on the first item:

```python
tag = "ol" if item.first_item_is_enumerated(doc) else "ul"
```

`ListGroup.first_item_is_enumerated()` checks only the first child:

```python
def first_item_is_enumerated(self, doc: "DoclingDocument"):
    return (
        len(self.children) > 0
        and isinstance(
            first_child := self.children[0].resolve(doc),
            ListItem,
        )
        and first_child.enumerated
    )
```

`HTMLTextSerializer.serialize()` currently applies an explicit marker style only when both of the following are true:

- `show_original_list_item_marker` is enabled; and
- `item.marker` is non-empty.

The relevant logic is:

```python
attrs=(
    {
        "style": (
            f"list-style-type: '{item.marker} ';"
        )
    }
    if params.show_original_list_item_marker
    and item.marker
    else {}
)
```

When `item.marker` is empty, the resulting `attrs={}` permits the browser to use the default marker inherited from `<ol>` or `<ul>`.

## Potential solution 1: suppress the marker per item

When original-marker rendering is enabled, explicitly use `list-style-type: none` for items without a detected marker:

```python
if params.show_original_list_item_marker:
    attrs = {
        "style": (
            f"list-style-type: '{item.marker} ';"
            if item.marker
            else "list-style-type: none;"
        )
    }
else:
    attrs = {}
```

The existing tag creation could then use:

```python
text = get_html_tag_with_text_direction(
    html_tag="li",
    text=text,
    attrs=attrs,
)
```

This is a small and local change. It prevents the parent list container from introducing an implicit marker when the `DoclingDocument` has no marker for that item.

## Potential solution 2: add a separate parameter

Changing the existing behavior of `show_original_list_item_marker=True` may affect users who currently rely on browser-generated fallback markers.

For backward compatibility, a separate parameter could be introduced:

```python
suppress_default_list_item_marker: bool = False
```

Possible behavior:

```python
if params.show_original_list_item_marker and item.marker:
    attrs = {
        "style": (
            f"list-style-type: '{item.marker} ';"
        )
    }
elif params.suppress_default_list_item_marker:
    attrs = {"style": "list-style-type: none;"}
else:
    attrs = {}
```

This would allow fidelity-oriented consumers to request:

```python
HTMLParams(
    show_original_list_item_marker=True,
    suppress_default_list_item_marker=True,
)
```

The resulting contract would be:

- show a marker when one is present in `ListItem.marker`;
- show no marker when `ListItem.marker` is empty;
- never derive or introduce numbering that is not represented in the document item.

## Potential solution 3: serializer-level marker strategy

A more extensible alternative would be a marker policy enum, for example:

```python
class HTMLListMarkerMode(str, Enum):
    AUTO = "auto"
    ORIGINAL = "original"
    NONE = "none"
```

Possible meanings:

- `AUTO`: retain the current HTML/browser behavior.
- `ORIGINAL`: show `item.marker` when present and suppress the marker otherwise.
- `NONE`: suppress all list markers.

This would make the intended behavior clearer than combining multiple booleans.

julianfesel:
This PR implements option 1 of issue #699 

HTML list items with an empty `marker` now render with `list-style-type: none;` when original markers are being shown, so the browser no longer invents numbering that is not present in the `DoclingDocument`.

### Changes

- Updated `HTMLTextSerializer` to suppress default list markers for unmarked list items while preserving explicit markers.
- Added a regression test for a mixed enumerated/non-enumerated list group.
- Refreshed the affected HTML and chunker goldens to match the new serializer output.

### DIFF
--- a/docling_core/transforms/serializer/html.py
+++ b/docling_core/transforms/serializer/html.py
@@ -219,8 +219,8 @@ class HTMLTextSerializer(BaseModel, BaseTextSerializer):
                     html_tag="li",
                     text=text,
                     attrs=(
-                        {"style": f"list-style-type: '{item.marker} ';"}
-                        if params.show_original_list_item_marker and item.marker
+                        {"style": (f"list-style-type: '{item.marker} ';" if item.marker else "list-style-type: none;")}
+                        if params.show_original_list_item_marker
                         else {}
                     ),
                 )
--- a/docling_core/transforms/serializer/html.py
+++ b/docling_core/transforms/serializer/html.py
@@ -216,12 +216,22 @@ class HTMLTextSerializer(BaseModel, BaseTextSerializer):
                 text = f"\n{text}\n"
             if params.show_original_list_item_marker:
                 parent_group = item.parent.resolve(doc) if item.parent else None
-                strict = isinstance(parent_group, ListGroup) and parent_group.any_item_has_marker(doc)
-                attrs = (
-                    {"style": (f"list-style-type: '{item.marker} ';" if item.marker else "list-style-type: none;")}
-                    if strict
-                    else {}
-                )
+                if (
+                    isinstance(parent_group, ListGroup)
+                    and parent_group.first_item_is_enumerated(doc)
+                    and parent_group.any_item_has_marker(doc)
+                ):
+                    # Ordered list with mixed markers: strict mode — suppress
+                    # browser-invented numbers for items without a detected marker.
+                    attrs = {
+                        "style": (f"list-style-type: '{item.marker} ';" if item.marker else "list-style-type: none;")
+                    }
+                elif item.marker:
+                    # Unordered list (or unmarked ordered list) with a detected
+                    # marker: show it as-is.
+                    attrs = {"style": f"list-style-type: '{item.marker} ';"}
+                else:
+                    attrs = {}
             else:
                 attrs = {}
             text = (
--- a/docling_core/transforms/serializer/html.py
+++ b/docling_core/transforms/serializer/html.py
@@ -214,15 +214,21 @@ class HTMLTextSerializer(BaseModel, BaseTextSerializer):
             text = "\n".join(text_parts)
             if nested_parts:
                 text = f"\n{text}\n"
+            if params.show_original_list_item_marker:
+                parent_group = item.parent.resolve(doc) if item.parent else None
+                strict = isinstance(parent_group, ListGroup) and parent_group.any_item_has_marker(doc)
+                attrs = (
+                    {"style": (f"list-style-type: '{item.marker} ';" if item.marker else "list-style-type: none;")}
+                    if strict
+                    else {}
+                )
+            else:
+                attrs = {}
             text = (
                 get_html_tag_with_text_direction(
                     html_tag="li",
                     text=text,
-                    attrs=(
-                        {"style": (f"list-style-type: '{item.marker} ';" if item.marker else "list-style-type: none;")}
-                        if params.show_original_list_item_marker
-                        else {}
-                    ),
+                    attrs=attrs,
                 )
                 if text
                 else ""
--- a/docling_core/types/doc/items/group.py
+++ b/docling_core/types/doc/items/group.py
@@ -44,6 +44,10 @@ class ListGroup(GroupItem):
             and first_child.enumerated
         )
 
+    def any_item_has_marker(self, doc: "DoclingDocument") -> bool:
+        """Whether any direct child list item has a non-empty marker."""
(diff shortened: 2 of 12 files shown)


dolfim-ibm:
@julianfesel are all the changes produced on the test files what you would expect?

julianfesel:
> @julianfesel are all the changes produced on the test files what you would expect?

@dolfim-ibm 

For show_original_list_item_marker=True, yes. However, based on your comment, I realized that I had misunderstood the role of marker: "" in Docling. I did not realize that it is not nullable.

In the new commits, I narrowed the application of strict markers to enumerated lists that contain at least one detected marker. As a result, there is now only one change in the test cases, which I believe is correct.

In my opinion, automatic enumeration does not make sense when at least one numeric marker is present. At the very least, this would require a more detailed validation. In such cases, one cannot assume that the default enumeration is correct. Furthermore, especially in OCR scenarios, the marker may actually be part of the text itself.

PR #707, opened 2026-08-02
feat: accept EBCDIC MIME types in DocumentOrigin

PeterStaar-IBM:
Registers the EBCDIC media type with `DocumentOrigin.validate_mimetype`, so
  the EBCDIC backend can build a `DoclingDocument` with an honest origin.

  EBCDIC data files have no IANA-registered media type, so `mimetypes.types_map`
  will never contain one, and `_extra_mimetypes` is a closed `ClassVar` that
  downstream packages cannot extend. Until now docling worked around this by
  calling `mimetypes.add_type` at import time in
  `docling/datamodel/base_models.py` — a global side effect on an unrelated
  stdlib registry that only happened to work because that module is imported
  before any `DocumentOrigin` is constructed.

  ### Changes

  - Add `application/x-ebcdic` to `DocumentOrigin._extra_mimetypes`. This is the
    standard spelling and the one docling emits from
    `FormatToMimeType[InputFormat.EBCDIC]` and `_MIME_TYPE` in
    `docling/backend/ebcdic_backend.py`. `application/octet-stream` was rejected
    because it would collide with every other unrecognized binary input during
    format detection.
  - Also accept `application/vnd.docling.ebcdic` as an alias, so origins written
    with the vendor-tree spelling still validate.
  - Add tests for both, plus a stdlib type, two existing extras, and a negative
    case asserting unknown types still raise.

  `ImageRef.validate_mimetype` is a separate validator for image payloads and is
  untouched.

### DIFF
--- a/docling_core/types/doc/common/origin.py
+++ b/docling_core/types/doc/common/origin.py
@@ -27,6 +27,7 @@ class DocumentOrigin(BaseModel):
     _extra_mimetypes: typing.ClassVar[list[str]] = [
         "application/epub+zip",
         "application/vnd.box.boxnote",
+        "application/vnd.docling.ebcdic",  # accepted alias, see application/x-ebcdic
         "application/vnd.oasis.opendocument.presentation",
         "application/vnd.oasis.opendocument.spreadsheet",
         "application/vnd.oasis.opendocument.text",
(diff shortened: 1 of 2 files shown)


dolfim-ibm (approved):
lgtm

maxmnemonic (approved):
LGTM

Issue #472, opened 2026-01-09
Export functions fail to generate <page_break> tags for non-consecutive (skipped) pages


jhchoi1182:
## Description 

When parsing PDF documents, if specific pages fail to parse (e.g., due to exceptions caught in the pipeline) and are excluded from the `doc.pages` list, the export functions (`export_to_doctags`, `export_to_html`, `export_to_markdown`) do not generate `page break` markers for these gaps. This causes page count mismatch issues in downstream processing. 

### Attachments
[test.pdf](https://github.com/user-attachments/files/24523289/test.pdf)
* `test.pdf` - A minimal reproduction file derived from a larger document. The content has been intentionally corrupted for confidentiality, but the parsing error still reproduces.
  * Structure: Consists of 4 pages (Pages 78, 79, 83, 84).
  * Behavior: Parsing succeeds for pages 78 and 84, but fails for pages 79 and 83. 

1.  Convert the attached `test.pdf` using `DocumentConverter`.
2.  Export using `export_to_doctags()`, `export_to_markdown(page_break_placeholder=str)`, `export_to_html(split_page_view=True)`.
3.  Count the `page break` 

**Reproduction Script:** 

```python
converter = DocumentConverter(
    format_options={
        InputFormat.PDF: PdfFormatOption(
            pipeline_options=pipeline_options,
        ),
    },
) 

doc = converter.convert(source).document

doctags_output = doc.export_to_doctags()
markdown_output = doc.export_to_markdown(page_break_placeholder="===PAGE_BREAK===")
html_output = doc.export_to_html(split_page_view=True) 

print(doctags_output)
print(markdown_output)
print(html_output)
``` 

## Expected Behavior
* **Export:** export fuction should generate `pagebreak` tags for **all pages**, including failed ones, so that page numbering remains consistent. 

## Actual Behavior
**Missing `page break` for failed pages:**
If pages 1, 2, 4, 5 succeed but page 3 fails, the output currently looks like this: 

```xml
... content ...      1page
<page_break>
... content ...      2page
<page_break>
... content ...      4page
<page_break>
... content ...      5page
``` 

```markdown
... content ...      1page
===PAGE_BREAK===
... content ...      2page
===PAGE_BREAK===
... content ...      4page
===PAGE_BREAK===
... content ...      5page
```

```html
<td>
<div class="page">
... content ...      1page
</div>
</td> 

<td>
<div class="page">
... content ...      2page
</div>
</td> 

<td>
<div class="page">
[...]

PR #708, opened 2026-08-04
fix: emit page break before a group whose children start on a new page

### LINKED ISSUE #705: Page break emitted after a group whose children start on a new page
### Bug

When `export_to_markdown(page_break_placeholder=...)` walks a `GroupItem` whose children begin on a
new page, the page break is emitted **after** the whole group instead of before it. Every item in
that group is attributed to the previous page.

The cause is in `_iterate_items()` (`docling_core/transforms/serializer/common.py`, lines 99-136):

1. The group branch only matches `ListGroup | InlineGroup`. A plain `GroupItem` — for example one
   with `GroupLabel.KEY_VALUE_AREA`, which the PDF pipeline creates for invoice header blocks — is
   not a `DocItem` either, so **neither branch runs** and no page break is emitted at the group.
2. The group is then serialized as a single part, children included.
3. Only afterwards does the outer loop reach the group's children, and the `elif isinstance(item,
   DocItem)` branch finally emits the break — by which point the children's text is already in the
   previous page's section.

There is a second, related problem in the same block: when the group branch *does* match (a
`ListGroup` or an `InlineGroup`), it yields the page break node but does **not** update
`prev_page_nr` nor increment `page_break_i`. The break for the same boundary is therefore emitted a
second time by the first child, and the duplicate only disappears because both nodes get the same
`self_ref` and `get_parts()` skips already-visited refs.

Impact for us: we rely on the placeholder count and position to map markdown sections to physical
pages (page-number metadata for citations, header/footer detection per page, OCR injection for
scanned pages). On a 287-page proposal document, 22 sections were misattributed and one section
ended up empty: the text of physical page 212 was in section 211, so section 212 had 0 % word
coverage against the page's native text layer.

We are working around it in our own `DocSerializer` subclass by (a) peeking the first page of **any**
`GroupItem` and (b) updating `prev_page_nr` when doing so. With that change our output becomes
byte-for-byte identical to exporting page by page (`export_to_markdown(page_no=n)` joined with the
placeholder) on all 21 multi-page documents of our corpus, while the page-by-page export costs
179.54 s versus 1.06 s on the 287-page document.

### Steps to reproduce

```python
from docling_core.types.doc.base import BoundingBox, CoordOrigin, Size
from docling_core.types.doc.document import DoclingDocument, PageItem, ProvenanceItem
from docling_core.types.doc.labels import DocItemLabel, GroupLabel

BBOX = BoundingBox(l=50, t=700, r=500, b=680, coord_origin=CoordOrigin.BOTTOMLEFT)

doc = DoclingDocument(name="repro")
for page_no in (1, 2):
    doc.pages[page_no] = PageItem(page_no=page_no, size=Size(width=595, height=842))

doc.add_text(label=DocItemLabel.TEXT, text="Body of page one",
             prov=ProvenanceItem(page_no=1, bbox=BBOX, charspan=(0, 16)))

group = doc.add_group(label=GroupLabel.KEY_VALUE_AREA)
for text in ("Phone", "Fax"):
    doc.add_text(label=DocItemLabel.TEXT, text=text, parent=group,
                 prov=ProvenanceItem(page_no=2, bbox=BBOX, charspan=(0, len(text))))

print(doc.export_to_markdown(page_break_placeholder="<!-- page break -->"))
```

Observed:

```
Body of page one

Phone

Fax

<!-- page break -->
```

Expected — `Phone` and `Fax` have `page_no=2`, so they belong after the break:

```
Body of page one

<!-- page break -->

Phone

Fax
```

Suggested fix, in `_iterate_items()`: match any `GroupItem` in the group branch (not just
`ListGroup | InlineGroup`), and set `prev_page_nr = page_no` / increment `page_break_i` when the
group's break is yielded, exactly as the `DocItem` branch does.

### Docling version

```
docling: 2.115.0
docling-core: 2.87.1
docling-ibm-models: 3.13.2
docling-parse: 7.8.1
```

### Python version

```
Python 3.10.12
```

maxmilian:
Fixes #705

## What was wrong

`_iterate_items()` in `docling_core/transforms/serializer/common.py` has two defects around groups, both visible with `add_page_breaks=True`:

1. **The group branch only matched `ListGroup | InlineGroup`.** A plain `GroupItem` — `KEY_VALUE_AREA` as reported, and the per-slide groups the PowerPoint backend produces — matched neither that branch nor the `elif isinstance(item, DocItem)` that follows, since `GroupItem` is not a `DocItem`. No break was emitted at the group, so the boundary was instead emitted by the group's first child, i.e. *after* the group, attributing the group's content to the previous page.

2. **The group branch yielded its break without advancing `prev_page_nr`**, unlike the `DocItem` branch. The group's first child therefore re-emitted the same boundary, and only the `self_ref` dedup in `get_parts()` kept that duplicate out of the output.

The reported `.pptx` symptom follows from (1): with each slide in its own group, an N-slide deck still gets N-1 placeholders — the *count* looks right — but they sit before slides 3…N with one orphan after the last slide, and slides 1 and 2 share a section.

## The fix

- Broaden the group branch to `GroupItem`, which covers `KEY_VALUE_AREA`, the per-slide groups, and the deprecated `OrderedList` (all subclass `GroupItem`).
- Advance `prev_page_nr` past the boundary the group branch emits, mirroring the `DocItem` branch.
- Key `_PageBreakNode.self_ref` on the boundary (`#/pb-{prev}-{next}`) rather than on a running counter.

That last point is what makes the second change safe. `get_parts()` re-enters for every group and shares `visited` with the recursion, so a group spanning a boundary is seen by both the nested and the root scope. A counter is scope-local — the nested scope restarts at 0 — so once the group branch starts incrementing it, the two scopes disagree on the ref and the duplicate stops being deduplicated. Keying on the boundary makes the identity independent of traversal order. The ref still satisfies the `^#(?:/([\w-]+)(?:/(\d+))?)?$` validator, since `[\w-]+` accepts dashes.

## Tests

`tests/test_serialization.py` gains a group of page-break tests that assert **positions**, not just placeholder counts — the reported symptom has the correct count throughout:

- `test_md_page_break_precedes_group_starting_on_new_page` — the `KEY_VALUE_AREA` repro from the issue.
- `test_md_page_break_per_slide_groups` — the 5-slide deck shape; each slide gets its own section and there is no trailing break.
- `test_md_page_break_positions_matrix` — (no group / group) × (gap at start / middle / end / none), asserting each page's text lands in its own section.
- `test_md_page_break_group_straddling_empty_pages` — a group whose children straddle one or two empty pages.
- `test_md_page_break_adjacent_groups_with_gap` — two adjacent groups with an empty page between them.
- `test_md_page_break_nested_groups` — a group inside a group, the one shape where the recursion and the shared `my_visited` actually interact.
- `test_page_break_boundary_emitted_once_per_transition` — asserts at the **iterator** level that no boundary is emitted twice, and that distinct transitions stay distinguishable.
- `test_page_break_before_group_across_serializers` — the same repro through DocTags, since `_iterate_items` is shared by markdown, LaTeX, DocTags and DocLang.

The iterator-level test earns its place because defect (2) is invisible from rendered output: reverting the `prev_page_nr` advance leaves every rendering test green, while the iterator emits `['#/pb-1-2', '#/pb-1-2']` and the dedup hides the duplicate. Verified fail-before / pass-after for each half of the fix independently, reverting one at a time against the final test set:

| reverted | result |
| --- | --- |
| `GroupItem` broadening | 11 failed / 14 passed (every group case) |
| `prev_page_nr` advance | 1 failed / 24 passed (the iterator-level test only) |
| nothing (this PR) | 25 passed |

## Checks

```
uv run pytest --all-extras   # 731 passed, 6 skipped (717 before this PR)
uv run ruff format --config=pyproject.toml   # clean
uv run ruff check --config=pyproject.toml    # clean
uv run mypy docling_core test                # Success: no issues found
```

## Notes

- **Pre-existing, not introduced here**: because refs are now keyed on `(prev, next)`, two genuinely distinct breaks sharing a boundary would collapse into one. This is only reachable with non-monotone provenance (e.g. two sibling groups each crossing 1→2). The counter scheme lost the same case — and lost more besides, since a nested scope's first break always collided with the root's first break regardless of boundary — so this is not a regression. Flagging it rather than leaving it to be discovered.
- **Follow-up, deliberately out of scope**: `doclang.py` still builds `_PageBreakNode.self_ref` from a counter in three places. Those nodes never reach `get_parts()`'s dedup, so there is no bug today, but they could adopt `_page_break_ref` for consistency. Worth noting `doclang.py` already dedups document-level breaks keyed on `(prev_page, next_page)`, so boundary-as-identity is an established idiom here.
- **Scope**: this does not touch #466 / #472 (expanding multi-page gaps, filling leading/trailing empty pages). Per the discussion on the issue, those are two separate rules — expanding a gap into consecutive boundaries has to happen in *every* scope so the refs match, while filling the document edges belongs only to the root scope — and merging them is what reintroduces this same shift. Keeping them out of this PR leaves that distinction intact.
- **No overlap with #639**, which changes `markdown.py` / `test_serialization.py`; this fix lives in `common.py`.
- @serboor offered to run this branch against their 26-document corpus (2–287 pages, native + scanned + PPTX-exported) and report the delta — that would be a valuable check beyond the synthetic matrix here, since their corpus passed *with* the latent defect (2) still present.


maxmilian:
Pushed `8a46911` — docstring only, no test or source behaviour change.

@serboor's measurements on #705 refuted a claim I had made in the docstring of `test_page_break_boundary_emitted_once_per_transition`. It said a faulty stream "never meets an input that would render differently". Across his 25-document corpus, a 287-page document rendered **7 pairs of consecutive placeholders** on the faulty build where the page-by-page reference had none — so the misplacement *is* visible in rendered markdown, and a corpus can catch it.

What a corpus cannot catch is the duplicate underneath: **732 boundaries emitted, 722 distinct, exactly 722 placeholders in the output**, because `get_parts()`'s `self_ref` dedup collapses the extra one before it renders — including on two documents that were byte-identical to the reference while their stream was wrong.

So both assertions in the test stay, for two different reasons, and the docstring now says that instead of overstating the first one. Same reasoning quoted with attribution, as offered on the issue.

The rest of his run is on #705: 4 documents change on this branch and they are the four that motivated the issue; three become byte-identical to the reference and the fourth differs by 2 bytes of whitespace that reproduces identically on `v2.90.0` (a `MarkdownListSerializer` join, in `markdown.py`, which this PR does not touch). The 32-page scan still disagrees on both builds — #472 / #466 territory, deliberately out of scope here.

Still waiting on a second approving review for the Mergify test-data protection.

maxmilian:
Rebased onto current `main` (`a727420`, v2.95.0). This one had to be repushed rather than merged: #739 renamed `test/` to `tests/`, so the branch as it stood would have resurrected a `test/` directory. The rebase applied cleanly across all four commits — git's rename detection redirected the patch on its own — and I checked the result by diffing the pre-rebase patch (with paths rewritten) against the post-rebase one: identical, so the change is unaltered at 2 files, +276/−6.

On the new base: `pytest --all-extras` gives 731 passed / 6 skipped, against 717 / 6 on `a727420` itself. `ruff check` and `ruff format --check` are clean. I've also corrected two stale references in the description above — it still pointed at `test/` and quoted counts from August.

Heads-up on one check that is about to change colour through no fault of the branch: the Mergify two-approval rule matches `files ~= ^tests`, which could not match `test/…` before #739. Now that this branch is on `tests/`, that rule applies to it for the first time, so `Mergify Merge Protections` will go from green to red pending reviews. That is the rule engaging, not a regression.

The gap is still open on `main`, re-checked today rather than assumed: `common.py:92/116/129` still mints page-break refs from a plain counter, the lookahead at `:100` still admits only `ListGroup | InlineGroup` and not `GroupItem`, and the group branch at `:113-121` still emits a `_PageBreakNode` without advancing `prev_page_nr` — while the `DocItem` branch at `:135-136` does. #705 is still open.

This branch has now been knocked out by upstream five times since it was opened. I'm happy to keep it current, but if page-break positioning isn't something you want changed in this shape, saying so costs you less than another rebase costs me — I'll close it without any hard feelings.

PR #709, opened 2026-08-04
feat: optionally preserve page numbers when concatenating

ZhiyaoWen999:
## Summary

- add an opt-in `preserve_page_numbers` argument to `DoclingDocument.concatenate()`
- preserve page keys, `PageItem.page_no`, item provenance, and key-value/form graph-cell provenance when enabled
- reject overlapping input page numbers instead of silently replacing pages
- keep the existing continuous renumbering behavior as the default

## Motivation

`_DocIndex.index()` always calculates a page offset that shifts every input document into a continuous sequence. This also shifts the first chunk when its original page range does not start at page 1, so callers processing PDF page ranges in parallel lose the source page numbers after concatenation.

The new opt-in mode uses a zero page offset after checking that the input ranges do not overlap. Existing callers are unaffected.

Closes docling-project/docling#3890.

## Validation

- targeted concatenate and filter tests: 6 passed
- Ruff lint and format checks
- MyPy on the changed implementation and tests
- issue fixtures: pages 5–10 and page 11 concatenate to pages 5–11; the 3,723-item result passes JSON serialization and model validation

### DIFF
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -5255,12 +5255,23 @@ class DoclingDocument(BaseModel):
         def get_item_list(self, key: str) -> list[NodeItem]:
             return getattr(self, key)
 
-        def index(self, doc: "DoclingDocument", page_nrs: Optional[set[int]] = None) -> None:
+        def index(
+            self,
+            doc: "DoclingDocument",
+            page_nrs: Optional[set[int]] = None,
+            preserve_page_numbers: bool = False,
+        ) -> None:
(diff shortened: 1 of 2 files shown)


ceberam (approved):
LGTM

dolfim-ibm (changes requested):
the PR looks good, but we want to hold until we do a full integration testing on the Docling CLI behavior and the Docling Serve API

the expected behavior is indeed, for example, that setting `page_range=(10,20)` will produce an output starting at page 10.

PR #716, opened 2026-08-09
fix(doctags): accept the paragraph tag when loading DocTags

manticore999:
### Problem

`export_to_doctags()` emits `<paragraph>` for items labelled `PARAGRAPH`, but the
tag is missing from the pattern `load_from_doctags()` matches against, so those
elements are skipped and their text is silently dropped. Export and import
disagree about the format.

`PARAGRAPH` is produced by the PPTX, JATS, USPTO, AsciiDoc and LaTeX backends, so
this affects ordinary documents. For a PowerPoint deck, every slide body is lost
and only the titles survive:

```python
from docling_core.types.doc.document import DocTagsDocument

doc = DocumentConverter().convert("deck.pptx").document
print(doc.export_to_text())
# Quarterly Report
# Revenue grew by twelve percent this quarter.
# Costs remained flat across all regions.

dt = doc.export_to_doctags()          # contains <paragraph>Revenue grew ...</paragraph>
back = DoclingDocument.load_from_doctags(
    DocTagsDocument.from_doctags_and_image_pairs([dt], [None]), document_name="d")
print(back.export_to_text())
# Quarterly Report          <- slide text is gone
```

Minimal version, no files needed:

```python
doc = DoclingDocument(name="x")
doc.add_text(label=DocItemLabel.PARAGRAPH, text="PARA")
doc.add_text(label=DocItemLabel.TEXT, text="TEXT")
# round-trip through DocTags -> "TEXT"   ("PARA" dropped)
```

### Fix

Add the tag to the parser's pattern and map it to `DocItemLabel.PARAGRAPH`, so
both the content and the label survive.

### Impact

Measured over the 192 groundtruth `DoclingDocument`s in the docling repo,
comparing text length before and after a DocTags round-trip:

| | documents losing >10% of text |
|---|---|
| before | 29 |
| after | 17 |

The six worst cases (patents `ipa20200022300`, `pftaps057006474`, `pg06442728`,
`pa20010031492`, `ipa20180000016`, and `powerpoint_malformed_pictures`) returned
0–3% of their text and now round-trip correctly. The remaining 17 lose text for
unrelated reasons (drawingml, some tables, checkboxes) and are untouched here.

### Tests

Two tests in `test/test_doctags_load.py`, covering the plain and `<loc_...>`
forms. Both fail without the change and pass with it; the existing doctags tests
are unaffected (12 passed).

**Checklist:**

- [ ] Documentation has been updated, if necessary.
- [ ] Examples have been added, if necessary.
- [x] Tests have been added, if necessary.

### DIFF
--- a/docling_core/types/doc/document.py
+++ b/docling_core/types/doc/document.py
@@ -4169,6 +4169,9 @@ class DoclingDocument(BaseModel):
             "checkbox_selected": DocItemLabel.CHECKBOX_SELECTED,
             "checkbox_unselected": DocItemLabel.CHECKBOX_UNSELECTED,
             "text": DocItemLabel.TEXT,
+            "paragraph": DocItemLabel.PARAGRAPH,
+            "handwritten_text": DocItemLabel.HANDWRITTEN_TEXT,
+            "reference": DocItemLabel.REFERENCE,
             "page_header": DocItemLabel.PAGE_HEADER,
             "page_footer": DocItemLabel.PAGE_FOOTER,
             "formula": DocItemLabel.FORMULA,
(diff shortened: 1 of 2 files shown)


manticore999:
Updated: widened from `<paragraph>` to also cover `<handwritten_text>` and `<reference>`.
All three are emitted by `export_to_doctags` but were missing from the parser's tag
pattern, so they hit the same silent-drop path. `REFERENCE` is produced by the LaTeX
backend; `PARAGRAPH` by PPTX, JATS, USPTO, AsciiDoc and LaTeX.

Tests are now parametrized over the three labels and fail without the change.

PR #717, opened 2026-08-12
fix(serializer): use block-level spacing for page breaks inside lists

OmarFaig:
A page break falling inside a `ListGroup` was emitted as an ordinary list part and joined with the list separator `\n`, while one falling between top-level items was joined with `\n\n`. Because the placeholder is substituted into already-joined text, the spacing was fixed by whichever scope emitted the break and could not be renegotiated downstream.

Exporting a document globally and page by page therefore disagreed by two bytes whenever a list straddled a page boundary, even though the content matched.

## Change

`MarkdownListSerializer` now separates a page break part from its neighbours with `\n\n`, so the marker forms a block in either scope. To let it recognise such a part, `DocSerializer.serialize()` returns the `_PageBreakSerResult` as such instead of rebuilding a plain result and dropping the type.

Output is unchanged for callers leaving `page_break_placeholder` unset, as no page break parts are produced at all in that case.

## On the golden-file churn

`test/data/doc/activities_pb_empty.gt.md` gains what looks like an excessive gap, but that is the existing document-scope behaviour — with an empty placeholder, doc scope already emits `'item one\n\n\n\nitem two'`. The list scope is now matching it rather than inventing new spacing.

## Tests

Added coverage for:

- global vs. per-page export equality for a list straddling a page boundary, with a non-list control that already held and must keep holding
- page break spacing being identical in list scope and document scope
- empty placeholder
- a break opening or closing the export (no blank line gained on the outer side)
- consecutive page breaks
- indent preservation for a break inside a nested list
- a break at a list-to-table boundary

Locally green before opening: `ruff check`, `ruff format --check`, `mypy` (139 source files), `pytest test` (599 passed, 6 skipped), `pytest packages/dclq/tests` (10 passed), and docs generation produced no diff.

### DIFF
--- a/docling_core/transforms/serializer/common.py
+++ b/docling_core/transforms/serializer/common.py
@@ -478,7 +478,9 @@ class DocSerializer(BaseModel, BaseDocSerializer):
                     **my_kwargs,
                 )
             elif isinstance(my_item, _PageBreakNode):
-                part = _PageBreakSerResult(
+                # returned as such, so that the combining scope can tell a page
+                # break apart from ordinary content
+                return _PageBreakSerResult(
                     text=self._create_page_break(node=my_item),
                     node=my_item,
                 )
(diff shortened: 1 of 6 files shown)


ceberam (approved):
Thanks @OmarFaig for the PR. The fix looks correct, minimal, and well-targeted.

dolfim-ibm (approved):

PR #720, opened 2026-08-14
feat: add EPUB book Markdown serializer

salmonumbrella:
## What changed

Adds `EpubDocSerializer`, a `MarkdownDocSerializer` variant that renders a book as Markdown preceded by a YAML frontmatter block holding the book metadata and a chapter index, plus `DoclingDocument.export_to_epub()` / `save_as_epub()`.

Every chapter entry carries the absolute UTF-8 byte offset and the 1-based line number of its heading, so a reader can seek straight to a chapter instead of unpacking the source or scanning the whole book.

```yaml
---
title: "Alice's Adventures in Wonderland"
authors: ["Lewis Carroll"]
language: "en"
chapters:
  - title: "CHAPTER I. Down the Rabbit-Hole"
    line:         12
    byte:        287
---
```

## Why

Requested by @dolfim-ibm in review of docling-project/docling#3993, which had the same serializer living in `docling/backend/epub_serializer.py`:

> The serializer should be in `docling-core`, and I would propose to add Epub as a standalone output format

The companion docling PR is being restructured to drop its local serializer and consume this one behind a new `epub` output format.

## Notes for reviewers

Two details are load-bearing and have dedicated tests:

- **The offsets are advertised inside the block they measure.** The frontmatter is rendered twice: once to measure how far it shifts the body, once with the shifted offsets. Padding every offset to a fixed width keeps the two renderings byte-identical; a guard raises if that ever stops holding.
- **A chapter is a `TitleItem` *or* a level-1 `SectionHeaderItem`.** An EPUB spine document may title its chapters with either `<h1>` or `<h2>`, and the HTML backend maps `<h1>` to a title item. Matching only section headers leaves the index empty for every book that uses `<h1>` per chapter.

`save_as_epub()` writes raw UTF-8 bytes rather than going through text mode, since the advertised offsets index the saved file and Windows newline translation would invalidate them.

## Checklist

- [x] Tests added (`test/test_epub_serialization.py`, 9 cases)
- [x] `pytest test` — 606 passed, 6 skipped
- [x] `ruff check` / `ruff format` / `mypy` clean
- [x] Generated docs regenerated: no schema drift

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docling_core/transforms/serializer/epub.py
+++ b/docling_core/transforms/serializer/epub.py
@@ -1,0 +1,155 @@
+"""Define classes for EPUB book Markdown serialization."""
+
+import json
+from typing import Any, Optional
+
+from pydantic import BaseModel, Field
+from typing_extensions import override
+
+from docling_core.transforms.serializer.base import SerializationResult
+from docling_core.transforms.serializer.common import create_ser_result
+from docling_core.transforms.serializer.markdown import (
+    MarkdownDocSerializer,
+    MarkdownParams,
+)
+from docling_core.types.doc.items.text import SectionHeaderItem, TitleItem
+
+_OFFSET_WIDTH = 10
+"""Width the chapter offsets are padded to.
+
+The frontmatter is rendered twice: once to measure how far it shifts the body,
+and once with the shifted offsets. Padding every offset to a fixed width keeps
+the second rendering the same size as the first, so the offsets it advertises
+remain correct.
+"""
+
+
+class EpubMetadata(BaseModel):
+    """Book metadata, typically read from an EPUB package document."""
+
+    title: Optional[str] = Field(default=None, description="The title of the book.")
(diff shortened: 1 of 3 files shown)


PeterStaar-IBM (approved):
looks good to me!

dolfim-ibm:
Doing a bit of research it looks like this output format is not Epub, but it could be used by some Epub generators as input.

More generically, we are introducing here the concept of yaml metadata in the markdown serializer, which I think has a huge value on its own.

My proposal for improving this work is
1. Generic - Add the concept of a yaml metadata serializer for the MarkdownSerializer
2. Add a MetadataSerializer (e.g. PubMetadataSerializer or another name) which implements this specific format
3. Remove explicit epub export functions in the DoclingDocument. If we add it, it should be real epub output.

salmonumbrella:
@dolfim-ibm lmk if you want me to make any changes!

PR #723, opened 2026-08-18
fix(serializer): derive markdown table header rows from column_header flags

### LINKED ISSUE #722: Markdown table serializer ignores TableCell.column_header, demoting stacked header rows to data
### Bug

`MarkdownTableSerializer` hardcodes grid row 0 as the only header row ([`markdown.py:575`](https://github.com/docling-project/docling-core/blob/v2.91.0/docling_core/transforms/serializer/markdown.py#L575)):

```python
table_text = tabulate(rows[1:], headers=rows[0], tablefmt="github", ...)
```

Every other consumer of the same `TableItem` derives the header rows from the `column_header` flags instead:

- [`TableItem._export_to_dataframe_with_options`](https://github.com/docling-project/docling-core/blob/v2.91.0/docling_core/types/doc/items/table/table.py#L90) counts leading flagged rows and combines them
- [the HTML serializer](https://github.com/docling-project/docling-core/blob/v2.91.0/docling_core/transforms/serializer/html.py#L423) emits `<th>` for any flagged cell
- `export_to_otsl` branches on `cell.column_header`

So a table whose header spans two rows serializes to Markdown with its lower header row demoted to a data row, while the DataFrame and HTML exports of the *same* `TableItem` get it right.

This is not only hand-built documents. Converting the Docling paper (arXiv 2408.09869) with stock `docling`, TableFormer flags **rows 0 and 1** of Table 1 as column headers:

```
table 0: 4x8   header-flagged rows: [0, 1]
table 1: 2x5   header-flagged rows: [0]
table 2: 13x6  header-flagged rows: [0]
```

The prediction is correct; the Markdown export discards it.

### Steps to reproduce

```python
from docling_core.types.doc.document import DoclingDocument
from docling_core.types.doc.items.table.table_data import TableCell, TableData


def cell(text, row, col, header=False):
    return TableCell(
        text=text,
        start_row_offset_idx=row, end_row_offset_idx=row + 1,
        start_col_offset_idx=col, end_col_offset_idx=col + 1,
        column_header=header,
    )


rows = [
    ["Product", "Tier 1", "Tier 2"],
    ["",        "0-99",   "100+"],
    ["CAT-001", "10.00",  "9.00"],
    ["CAT-002", "20.00",  "18.00"],
]
cells = [cell(t, r, c, header=(r < 2)) for r, row in enumerate(rows) for c, t in enumerate(row)]

doc = DoclingDocument(name="repro")
doc.add_table(data=TableData(table_cells=cells, num_rows=4, num_cols=3))
table = doc.tables[0]

print(table.export_to_dataframe(doc=doc).to_string())
print(table.export_to_html(doc=doc))
print(table.export_to_markdown(doc=doc))
```

**Actual** — the DataFrame keeps two header rows and two data rows, the HTML marks both header rows as `<th>`, and the Markdown demotes the second header row:

```text
  Product. Tier 1.0-99 Tier 2.100+
0  CAT-001       10.00        9.00
1  CAT-002       20.00       18.00

<table><tbody><tr><th>Product</th><th>Tier 1</th><th>Tier 2</th></tr><tr><th></th><th>0-99</th><th>100+</th></tr>...

| Product   | Tier 1   | Tier 2   |
|-----------|----------|----------|
|           | 0-99     | 100+     |     <- header row rendered as data
| CAT-001   | 10.00    | 9.00     |
| CAT-002   | 20.00    | 18.00    |
```

**Expected** — the flagged rows form the single header row GFM allows, and only the data rows are body:

```text
| Product   |   Tier 1 0-99 |   Tier 2 100+ |
|-----------|---------------|---------------|
| CAT-001   |         10.00 |          9.00 |
| CAT-002   |         20.00 |         18.00 |
```

### Scope

The same conversion's real tables show what this costs. Table 1 of the Docling paper currently exports with a header of repeated merge-spans while the row naming the measurements becomes a body row with a blank `CPU`:

```text
| CPU | Thread budget | native backend | native backend | native backend | ...
|-----|---------------|----------------|----------------|----------------|
|     |               | TTS            | Pages/s        | Mem            | ...
```

Table 1 of the DocLayNet paper (`test/data/doc/2206.01062.yaml.md`) loses its entire column-name row the same way — `class label`, `Count`, `Train`, `Test`, `Val` are all in the body, under a header of `% of Total` repeats.

It also reaches chunking. `HybridChunker` calls `table_serializer.get_header_and_body_lines()` ([`hybrid_chunker.py:295`](https://github.com/docling-project/docling-core/blob/v2.91.0/docling_core/transforms/chunker/hybrid_chunker.py#L295)) to pick the lines repeated atop each chunk of an oversized table, and for Markdown that locates the single separator row. So the qualifier rows stay in the body and every chunk after the first carries a header stripped of the qualifiers that make its numbers meaningful.

### Suggested fix

Take the header row count from the flags and flatten those rows into one. PR follows.

Two calls I would rather a maintainer made:

1. **Tables with no `column_header` at all.** I kept the current behavior (row 0 becomes the header) in that case, because backends that never populate the flag cannot be told apart from genuinely headerless tables — several `*.dt.json` fixtures round-trip with zero flags. Making it unconditional is cleaner but changes output for those inputs. Happy to switch if the flag is guaranteed somewhere I missed.
2. **Join separator.** I join stacked header cells with a space (`native backend TTS`); `_export_to_dataframe_with_options` uses `.` (`native backend.TTS`, with a trailing `.` when the lower cell is empty). These arguably should agree.

### Docling version

`docling-core` 2.91.0. Live conversion checked with `docling` 2.120.2 / `docling-core` 2.91.0.

### Python version

3.12.13 for the live conversion, 3.14.6 for the test runs.

acharyaanusha:
Fixes #722.

`MarkdownTableSerializer` hardcoded grid row 0 as the only header row, while `TableItem._export_to_dataframe_with_options`, the HTML serializer and `export_to_otsl` all derive the header rows from `TableCell.column_header`. A table with a stacked header serialized to Markdown with its lower header rows demoted to data rows, and `HybridChunker` repeated only the first of them atop each chunk of an oversized table.

This takes the header row count from the flags and flattens those rows into the single header row GFM allows, dropping the repeated text a cell spanning several header rows produces (so `human` / `human` stays `human`, not `human human`).

When **no** cell in the table carries `column_header`, the previous behavior is kept. Backends that never set the flag cannot be told apart from genuinely headerless tables, and several `*.dt.json` fixtures round-trip with zero flags. Making it unconditional instead changed output for those too, so I left it — see the first open question in #722 if you would rather it be strict.

### Ground truth

Five ground-truth files are regenerated. Every diff recovers real column names from the body:

- `2206.01062.yaml.md` / `.paged.md` (DocLayNet) — `class label | Count | % of Total Train | ...` was a body row under a header of `% of Total` repeats; likewise `MRCNN R50` / `MRCNN R101`, `Class-count Split`, `Testing on PLN`
- `2408.09869v3_enriched_p1_mark_meta_{true,false}.gt.md` — `native backend TTS | native backend Pages/s | native backend Mem`
- `0b_out_chunks.json` / `0d_out_chunks.json` — the same headers as repeated in chunks

Numeric columns shift to right-alignment in a few places, because `_column_type` no longer sees the former header row when detecting the column type.

### Verification

- Full `docling-core` suite: no new failures, +3 from the tests added here. My environment has pre-existing unrelated failures (missing `cv2` stubs, `doclang` schema resources, and a Pillow that re-encodes the fixture PNGs), so I diffed the failure sets before and after rather than trusting the totals; the sets are identical.
- `ruff format`, `ruff check`, and `mypy` clean on the changed files.
- End-to-end on a live conversion: converted arXiv 2408.09869 with `docling` 2.120.2 and diffed the full 248-line Markdown export before and after. Exactly one table changes; the two single-header tables in the same document are byte-identical.

### Tests

Three cases in `test/test_serialization.py`: a stacked header flattened, repeated span text deduplicated, and a table with no header flags keeping its first row as the header.


ceberam:
@acharyaanusha Thanks again for your efforts with this PR. I see you have recently added more commits, so I assume this PR is in-progress. Therefore I set it up as **Draft**. Please, feel free to set it **Ready for review** or drop a line here when you're ready.

acharyaanusha:
@ceberam Setting it back to ready, as the PR is now complete.

The extra commits were a rebase and conflict resolution for the test to tests rename, plus one real fix.  
#756 changed the DataFrame export to only count a row as a header row when a header cell actually starts on it. _count_header_rows here was modelled on the previous version of that function, so it had the same defect, so the data rows under a vertically spanning header were pulled into the header block, flattened into the column names, and dropped from the body.

ceberam (approved):
Thanks @acharyaanusha for the latest changes!

I just spotted a small room for improvement in `_count_header_rows` and I also some clarifications that could be helpful in the new functions.
Since we are planning a new release soon, I've just added those changes directly.

The PR looks ready to merge 🚀

PeterStaar-IBM (approved):
lgtm!

PR #730, opened 2026-08-28
feat: add FormulaMeta to carry native MathML on formulas

DanielNg0729:
Issue Resolved by this PR:
https://github.com/docling-project/docling/issues/4081

Thank you Dr @PeterStaar-IBM for the idea of this.
This PR adds a typed home on `FormulaItem` for a structured mathematical representation, so a formula can carry MathML and LaTeX alongside its rendered text instead of only the text a model reconstructed from the glyphs.

Today the only MathML in the stack is produced at export time: the HTML serializer runs `latex2mathml` over whatever LaTeX the formula understanding model predicted. That makes `glyphs -> model -> LaTeX -> MathML` the only path, and it discards any MathML the author actually embedded. Tagged PDFs (PDF/UA output from Word, for example) expose a MathML
representation on the `Formula` structure element, which is more faithful than anything reconstructed from the rendering.

This PR is the data-model half of that work. It does not read any PDF itself, the extraction lives in docling and is tracked in docling-project/docling#4081.

## What changed

- **`FormulaMetaField` / `FormulaMeta`** in `common/meta.py`, modelled on the existing
  `CodeMetaField`. `mathml` and `latex` are both optional; `created_by` (inherited from
  `BasePrediction`) records whether a value came from the source document or a model.
- **`FormulaItem.meta`** narrowed from `BaseMeta` to `FormulaMeta`, the same way
  `FloatingItem` and `PictureItem` narrow theirs.
- **HTML export prefers native MathML.** `_process_formula` emits `meta.formula.mathml`
  verbatim when present, ahead of both the image fallback and the `latex2mathml`
  conversion. It still honours `formula_to_mathml=False`. This also fixes the case where
  a formula has MathML but an empty `text`, which previously rendered as
  "Formula not decoded".
- **Meta-field serializers** in DocLang, HTML and Markdown suppress the new field, since
  the formula is already emitted as the item's own content. Without this, DocLang raised
  `UnboundLocalError` and Markdown dumped the Pydantic repr into the output — both because
  their generic meta serializers dispatch on the known `MetaFieldName` members and a new
  member matched no branch.

## Schema bump: 1.10.0 -> 1.11.0

Worth flagging for review: **the JSON Schema diff cannot detect this break.** `BaseMeta` allows extra keys, so `formula` does not register as a new property on a strict model. But `_ExtraAllowingModel._validate_field_names` rejects any extra key that is not namespaced as `namespace__field`, so a 1.10 client raises `ValidationError` on the new payload. That is the "serialization-semantic change, invisible to static analysis"

The six golden files in `test/data/` are one-line schema-version updates
(`1.10.0` -> `1.11.0`), nothing else.

## Verification

- `scripts/check_compat_projectors.py`: OK, 6 projectors for schema 1.5..1.11, all
  snapshots present.
- `ruff check` / `ruff format`: clean.
- `mypy docling_core`: no new errors (the 7 reported are pre-existing missing library
  stubs for `pandas` / `defusedxml`).
- Full test suite: 460 passed, unchanged failure set versus the same tree without this
  change (measured by stashing). 15 new tests, 0 regressions.

**Checklist:**

- [x] Documentation has been updated, if necessary.
- [ ] Examples have been added, if necessary.
- [x] Tests have been added, if necessary.


DanielNg0729:
Hi Dr @PeterStaar-IBM, I bumped current_version to 1.11.0 but missed several test that still pinned 1.10.0.
All of them are properly updated now. This PR is ready for review when you have a moment. Thank you very much!

PeterStaar-IBM (dismissed):
lgtm!

DanielNg0729:
Hi Dr @PeterStaar-IBM Dr @cau-git Dr @dolfim-ibm, I’ve just resolved the merge conflict after the update of test folder. This is now ready for review whenever you have a moment. Thank you very much!

PR #736, opened 2026-09-01
fix: updated the doclang serializer for groups

PeterStaar-IBM:
## Summary

  A plain `GroupItem` is transparent in DocLang: the fallback serializer emits its
  children and drops the grouping, so any structural grouping in a
  `DoclingDocument` is lost on a round trip.

  This adds `add_named_groups` to `DocLangParams` (default `False`, so nothing
  changes unless you ask for it). When enabled, a plain `GroupItem` is emitted as
  a `<group>` element carrying its name and label, and the deserializer rebuilds
  it.

  ```xml
  <group name="article">
    <label value="section"/>
    <heading level="3">Headline</heading>
    <text>Body</text>
  </group>
```

  The motivating case is synthetic newspaper ground truth, where each article is a
  GroupItem holding everything from its section label to its last paragraph.

  Changes

  - serializer/_doclang_utils.py: DocLangAttributeKey.NAME, allowed on
    DocLangToken.GROUP; _create_group_token takes an optional name.
  - serializer/doclang.py: DocLangParams.add_named_groups; DocLangFallbackSerializer
    wraps plain groups when it is set. ListGroup and InlineGroup are excluded —
    they have their own serializers. The <label> element is emitted only when the
    label is not UNSPECIFIED, keeping XSD element-head order and satisfying the
    schematron head-placement rule.
  - deserializer/doclang.py: _parse_named_group restores the group's name,
    label and nesting. The name attribute is checked before the float +
    footnote <group> heuristic — that wrapper never carries a name, so without
    this an article group holding a picture or a table would be parsed as a float
    and swallow its siblings.
  - types/doc/document.py: add_named_groups threaded through
    export_to_doclang, save_as_doclang and save_as_doclang_archive.

  Testing

  Five tests in test/test_deserializer_doclang.py: default-off behaviour,
  name/label/nesting round trip, attribute escaping, a named group containing a
  picture and a table, and the unnamed float wrapper still parsing as one unit.

  With the flag off, export_to_doclang() is byte-identical to the previous
  behaviour across every loadable document in test/data/doc/, and
  serialize → deserialize results are unchanged.

  Follow-up: schema

  doclang.xsd gives <group> no attributes, so with the flag enabled validation
  reports attribute 'name' is not allowed. Default output validates clean, and
  the new tests skip XSD validation with a comment pointing here. The schema needs
  one line in the doclang repo:

```
  <xs:element name="group">
    <xs:complexType>
      <xs:sequence>
        <xs:group ref="dl:element_head"/>
        <xs:group ref="dl:top_level_cat" minOccurs="0" maxOccurs="unbounded"/>
      </xs:sequence>
      <xs:attribute name="name" type="xs:string"/>
    </xs:complexType>
  </xs:element>
```
  Verified against a patched copy of the schema. Once it lands, the
  validate=False in the new tests can be dropped.

### DIFF
--- a/docling_core/transforms/deserializer/doclang.py
+++ b/docling_core/transforms/deserializer/doclang.py
@@ -362,9 +362,7 @@ class DocLangDocDeserializer(BaseDocDeserializer, BaseModel):
         else:
             self._walk_children(doc=doc, el=el, parent=parent)
 
-    def _parse_named_group(
-        self, *, doc: DoclingDocument, el: Element, parent: Optional[NodeItem], name: str
-    ) -> None:
+    def _parse_named_group(self, *, doc: DoclingDocument, el: Element, parent: Optional[NodeItem], name: str) -> None:
         r"""Rebuild a GroupItem from <group name="..."> and nest its children in it."""
         head_nodes = [node for node in el.childNodes if isinstance(node, Element) and self._is_element_head_tag(node)]
         label = GroupLabel.UNSPECIFIED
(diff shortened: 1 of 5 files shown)


dolfim-ibm (approved):
lgtm

vagenas (approved):

PR #744, opened 2026-09-03
fix: strip cue timestamp tags when parsing WebVTT cue text

ceberam:
Resolves docling-project/docling#4152

## Problem

WebVTT cue text may contain inline cue timestamp tags such as `<00:00:00.389>`
(defined in the [WebVTT spec §cue-timestamp](https://www.w3.org/TR/webvtt1/#webvtt-cue-timestamp)).
This is the format written by yt-dlp for YouTube auto-generated captions.

The parser did not recognise these tags, so the `<` character they contain
caused `WebVTTCueTextSpan` validation to fail with
`"Cue text contains invalid characters"`. The entire cue block was silently
dropped with a `RuntimeWarning`, resulting in an empty document despite a
`ConversionStatus.SUCCESS` status.

## Solution

- Extract the timestamp regex into a module-level string constant
  `_TIMESTAMP_PATTERN_STR`, shared by both `WebVTTTimestamp._pattern` and the
  new `WebVTTCueBlock._pattern_cue_timestamp`.
- Before the tag-parsing loop in `WebVTTCueBlock.parse`, strip all cue
  timestamp tags from the cue text using `_pattern_cue_timestamp`. The
  surrounding words are preserved and parsed normally.

## Changes

- `docling_core/types/doc/webvtt.py` — introduce `_TIMESTAMP_PATTERN_STR`,
  refactor `WebVTTTimestamp._pattern` to use it, add
  `WebVTTCueBlock._pattern_cue_timestamp` and the stripping step in `parse`.
- `test/data/webvtt/webvtt_example_04.vtt` — add a cue block with inline
  timestamp tags to exercise the fix through `WebVTTFile.parse`.
- `test/test_webvtt.py` — add assertions for the new cue block in
  `test_webvtt_file`, and a dedicated case in `test_webvttcueblock_parse`.

### DIFF
--- a/docling_core/types/doc/webvtt.py
+++ b/docling_core/types/doc/webvtt.py
@@ -13,6 +13,7 @@ from typing_extensions import Self, override
 
 _VALID_ENTITIES: set = {"amp", "lt", "gt", "lrm", "rlm", "nbsp"}
 _ENTITY_PATTERN: re.Pattern = re.compile(r"&([a-zA-Z0-9]+);")
+_TIMESTAMP_PATTERN_STR: str = r"(?:(\d{2,}):)?([0-5]\d):([0-5]\d)\.(\d{3})"
 START_TAG_NAMES = Literal["c", "b", "i", "u", "v", "lang"]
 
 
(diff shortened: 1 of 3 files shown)


cau-git (approved):

PeterStaar-IBM (approved):
lgtm!

PR #752, opened 2026-09-06
fix(doclang): nest InlineGroup runs inside a TextItem that has its own text

### LINKED ISSUE #750: DocLangDocSerializer emits an InlineGroup's runs with no enclosing element when the parent TextItem has text of its own
`DocLangInlineSerializer.serialize` sets `should_wrap = False` for every `InlineGroup` whose parent is a `TextItem`, on the assumption the parent will render the runs as its own element content. `DocLangTextSerializer.serialize` only does that under `if item.children and not item.text:` — so when the parent carries text, the runs are emitted bare at the parent's sibling level. In a `<group>` (or at the `<doclang>` root) that is invalid against the real DocLang 0.7 XSD, whose `group` content model is `element_head` + `top_level_cat*` and is not `mixed`: `Element 'bold': This element is not expected.` The XML stays well-formed, so `DocLangDocSerializer` does not raise, but `_walk_children` dispatches only on `isinstance(node, Element)`, so the DOM text node between two formatting runs is never visited and the text is silently lost — a content-loss bug independent of the schema one.

Minimal repro (docling-core 2.94.1, doclang 0.7.3): a `DoclingDocument` with a `SectionHeaderItem` that has text, one `InlineGroup` child, and three `TextItem` runs under it (bold, plain, italic). This is the shape docling's own DOCX backend produces for any Word paragraph of inline-formatted runs that follows a heading.

Two candidate fixes: narrow `should_wrap` to `isinstance(parent_item, TextItem) and not parent_item.text`, or have the text serializer render an `InlineGroup` child inside its own element even when it has text of its own.

---

Found while auditing a downstream project's DocLang export pipeline; happy to share the exact synthetic `DoclingDocument` construction used to reproduce this if useful.

maxmilian:
### Problem

`DocLangTextSerializer.serialize` renders a `TextItem`'s `InlineGroup` child as the item's element content only under `if item.children and not item.text:`. Meanwhile `DocLangInlineSerializer.serialize` sets `should_wrap = False` for *every* `InlineGroup` whose parent is a `TextItem`, assuming the parent will render the runs itself.

When the parent `TextItem` carries text of its own, neither side emits a container and the runs land bare at the parent's sibling level:

```xml
<doclang version="0.7">
  <heading level="2">Heading text</heading>
  <bold>bold</bold>
  <content> plain </content>
  <italic>italic</italic>
</doclang>
```

Two consequences:

1. **Invalid output** — the DocLang 0.7 `group` content model is `element_head` + `top_level_cat*` and is not `mixed`, so validation reports `Element 'bold': This element is not expected.`
2. **Silent content loss** — the XML stays well-formed, so serialization itself does not raise. But the deserializer's `_walk_children` dispatches only on `isinstance(node, Element)`, so the DOM text node between two runs is never visited and ` plain ` disappears on round-trip.

This is the shape docling's DOCX backend produces for a paragraph of inline-formatted runs.

### Fix

Render the `InlineGroup` child as element content whenever it is the first child, prefixing the item's own text when it has any — so the runs stay inside the parent element:

```xml
<doclang version="0.7">
  <heading level="2">
    Heading text
    <bold>bold</bold>
    <content> plain </content>
    <italic>italic</italic>
  </heading>
</doclang>
```

This puts the text-bearing case on exactly the same footing as the already-correct empty-text case: with `heading.text == ""`, the same document already serializes correctly nested today, so this is a gap in one path rather than a second kind of shape.

I deliberately did **not** narrow `should_wrap` in `DocLangInlineSerializer`, which is the other obvious way to make the schema error go away. That would emit `<heading>Heading text</heading><text><bold>…</bold>…</text>` — valid, and it does not lose the plain run, but it demotes the `InlineGroup` to a sibling when it *is* a child of the `TextItem` in the `DoclingDocument`. Round-tripping that output yields a heading and a parallel `InlineGroup`, so the parent/child relationship is lost permanently. Fixing the container rather than the nesting keeps the serialized tree faithful to the document tree.

### Tests

New regression test `test_inline_group_nested_when_parent_text_item_has_text` asserts the runs are nested inside `<heading>` and that the plain run survives. It fails on `main` with the XSD error above; I also re-ran it with the production change stashed to confirm it goes red for the right reason rather than passing vacuously.

`tests/test_serialization_doclang.py`: 91 passed. Full suite green (note: `uv sync --frozen --all-extras` is needed, otherwise seven chunker modules fail to collect on missing `tree_sitter`/`tiktoken`/`transformers`). `ruff check` and `ruff format --check` clean.

**No golden fixture changes** — no existing fixture has a `TextItem` carrying both text and an `InlineGroup` child, which is why this went unnoticed.

### Noted but out of scope

`serializer/html.py:157-161` guards `has_inline_repr` on the same `item.text == ""` premise, so the HTML serializer likely has the same gap. I have not touched it here to keep this change reviewable against one schema; happy to open a separate issue for it.

Fixes #750

### DIFF
--- a/docling_core/transforms/serializer/doclang.py
+++ b/docling_core/transforms/serializer/doclang.py
@@ -965,24 +965,32 @@ class DocLangTextSerializer(BaseModel, BaseTextSerializer):
             or (isinstance(item, FormulaItem) and ContentType.TEXT_FORMULA in params.content_types)
             or (not isinstance(item, CodeItem | FormulaItem) and ContentType.TEXT_OTHER in params.content_types)
         ):
-            if item.children and not item.text:
-                # Check if first child is InlineGroup - if so, only serialize that as text content
-                first_child_ref = item.children[0]
-                first_child_item = first_child_ref.resolve(doc)
-
-                if isinstance(first_child_item, InlineGroup):
-                    # Only serialize the first child (InlineGroup) as the text content
-                    # Other children are hierarchical subordinates and will be serialized separately
-                    text_part = doc_serializer.serialize(item=first_child_item, visited=my_visited, **kwargs).text
+            first_child_item = item.children[0].resolve(doc) if item.children else None
+            if isinstance(first_child_item, InlineGroup):
+                # Only serialize the first child (InlineGroup) as the text content
+                # Other children are hierarchical subordinates and will be serialized separately.
+                # The item's own text, when present, precedes the runs within the same element:
+                # emitting the runs as siblings instead yields invalid DocLang and loses content.
+                inline_part = doc_serializer.serialize(item=first_child_item, visited=my_visited, **kwargs).text
+                if item.text:
+                    own_part = _escape_text(item.text, params)
+                    own_part = doc_serializer.post_process(
+                        text=own_part,
+                        formatting=item.formatting,
+                        hyperlink=None,
+                    )
+                    text_part = _get_delim(params=params).join([p for p in (own_part, inline_part) if p])
                 else:
-                    # Serialize all children as text content
-                    sub_parts: list[str] = []
-                    for child_ref in item.children:
-                        child_item = child_ref.resolve(doc)
-                        if isinstance(item, ListItem) and _list_item_segment_sibling(child_item):
-                            continue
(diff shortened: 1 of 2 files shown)


vagenas (approved):

PeterStaar-IBM (approved):
lgtm!

PR #754, opened 2026-09-06
perf: reuse table grid during dataframe export

### LINKED ISSUE #753: Avoid rebuilding the table grid for every header row
An Excel sheet can contain a vertically merged cell that spans every row in a detected table. Docling represents that merged cell as a column header with a `row_span` that covers the table.

`TableData.grid` repeats the spanning cell in every covered row. `TableItem.export_to_dataframe()` then treats every grid row as a header row. While it builds the column names, it reads the computed `TableData.grid` property once for each header row. Each read rebuilds the complete grid.

For a table with 10,000 rows, this causes about 10,000 complete grid builds and can make dataframe export run for minutes or hours.

mohahf19:
The DataFrame exporter rebuilds the full table grid once per header row, plus twice more. This change builds the grid once per export and reuses it for header detection, column names, and data rows.

PR #756 fixed header detection for cells that span several rows. With that fix on `main`, a header that spans 10,000 rows causes three grid builds. This PR reduces that count to one. Tables with multiple header rows get a larger reduction.

Fixes #753.

## Performance

Median export times from three local runs on constructed tables with 10,000 rows and 10 columns. File conversion is excluded. The baseline is `main` at `a0949d7`, which includes #756.

| Table structure | Grid builds before | Before | After |
| --- | ---: | ---: | ---: |
| Header spans all rows | 3 | 0.428 s | 0.161 s |
| No headers | 2 | 0.280 s | 0.155 s |
| One header row | 3 | 0.412 s | 0.156 s |
| Three header rows | 5 | 0.664 s | 0.151 s |
| Ten header rows | 12 | 1.570 s | 0.152 s |


mohahf19:
@ceberam ready for you to take a look!

ceberam (approved):
Thanks @mohahf19 for this fix!

The issue was severe and for that scenario of a fully-spanning header row, the fix brings massive performance gains.

We don't need a test in this case, since a timing-based test at scale would just add more burden to our CI/CD pipeline.

Ready to merge 🚀

PR #764, opened 2026-09-10
fix: Add missing dependency on `requests`

mgorny:
Add an explicit dependency on `requests` since it is used in `docling_core.utils.file` since 752cbc3.  Without it, the CLI tools fail to run on a fresh install (e.g. via `uv pip install docling-core`):

```pytb
$ docling-view
  File "/tmp/.venv/bin/docling-view", line 4, in <module>
    from docling_core.cli.view import app
  File "/tmp/.venv/lib/python3.14/site-packages/docling_core/cli/view.py", line 13, in <module>
    from docling_core.utils.file import resolve_source_to_path
  File "/tmp/.venv/lib/python3.14/site-packages/docling_core/utils/file.py", line 11, in <module>
    import requests
ModuleNotFoundError: No module named 'requests'
```


pyproject.toml
```diff
@@ -52,6 +52,7 @@ dependencies = [
     "defusedxml (>=0.7.1, <0.8.0)",
     "pydantic-settings>=2.14.0",
     "doclang>=0.7,<0.8",
+    "requests>=2.34.2",
```
ceberam: Please, cap the major version and update `uv.lock` accordingly:

```suggestion
    "requests (>=2.34.2,<3.0.0)",
```

ceberam (changes requested):
Thanks @mgorny for finding and fixing this bug! I can confirm that `requests` is required indeed. I added a small comment, as safety net.

mgorny:
Done.

ceberam (approved):
Great! 🚀

PR #766, opened 2026-09-10
fix(serializer): keep row-header rows out of the markdown table header block

### LINKED ISSUE #765: Markdown serializer: header block from column_header flags swallows the first data row of pivot tables (HTML <th rowspan> row headers)
Summary

Since [docling-project/docling-core#723](https://github.com/docling-project/docling-core/pull/723) (v2.96.0) MarkdownTableSerializer derives the header block from column_header flags: every leading row on which a flagged cell starts is a header row, and stacked header rows are flattened per column with " - ".

Documents from the docling HTML backend flag every `<th>` as column_header, including row headers (`<th rowspan="…"> `in body rows; row_header stays False). For a pivot table the first body row therefore has a flagged cell starting on it, the header block grows to two rows, and the first data row is folded into the header.

Reproduction

tests/data/html/groundtruth/example_08.html.json from the docling repo, "Pivot table with 1 row header"):
```
<table>
  <tr><th>Year</th><th>Month</th><th>Revenue</th><th>Cost</th></tr>
  <tr><th rowspan="5">2025</th><td>January</td><td>$134</td><td>$162</td></tr>
  <tr><td>February</td><td>$150</td><td>$155</td></tr>
  …
</table>
```
Grid flags (row 1 holds 2025 with column_header=True, start_row_offset_idx=1, row_span=5):
```
0 [('Year', True), ('Month', True), ('Revenue', True), ('Cost', True)]
1 [('2025', True), ('January', False), ('$134', False), ('$162', False)]
2 [('2025', True), ('February', False), ...]
```
`DoclingDocument.load_from_json(...).export_to_markdown()` with docling-core 2.96.0:
```
|   Year - 2025 | Month - January   | Revenue - $134   | Cost - $162   |
|---------------|-------------------|------------------|---------------|
|          2025 | February          | $150             | $155          |
|          2025 | March             | $160             | $143          |
```
Expected (2.95 output, and what the table means):
```
|   Year | Month    | Revenue   | Cost   |
|--------|----------|-----------|--------|
|   2025 | January  | $134      | $162   |
|   2025 | February | $150      | $155   |
```
TableItem.export_to_dataframe() ([docling-project/docling-core#756](https://github.com/docling-project/docling-core/pull/756)) uses the same header count and yields the column name Year.2025.

**Possible fixes**

Serializer (_count_header_rows): stop the header block at a row on which a non-header cell with text also starts. A row mixing header and data cells is a data row with a row header. Rows made only of header cells (and empty corner cells) keep the current behaviour, so genuine stacked headers such as % of Total - Train are unaffected.
HTML backend (docling): flag <th> cells in body rows or with scope="row" as row_header instead of column_header.

Leewwp:
## What

`_count_header_rows()` now only promotes a row into the markdown column-header block when that row carries no body text outside the leading column.

## Why

The HTML backend marks a `<th rowspan="N">` that occupies its own `<tr>` as a *column* header — it only clears `col_header` when the row holds a `<td>`. Such a cell is semantically a *row* header, and the row it starts on is the first body row.

Since #723, `_count_header_rows()` trusts the `column_header` flag, so that first body row is folded into the header block and disappears from the body:

```
| Year - 2025 | Month - January | Revenue - $134 | Cost - $162 |
|-------------|-----------------|----------------|-------------|
|        2025 | February        | $150           | $155        |   <- January row lost
```

After this change:

```
| Year | Month    | Revenue | Cost |
|------|----------|---------|------|
| 2025 | January  | $134    | $162 |
| 2025 | February | $150    | $155 |
```

Text in the leading column is exempt, because that column carries the row labels (`CPU`, `Class-count`, `Split`), which the backend leaves unmarked even on rows that do belong to the header. The rule is derived from the multi-row header cases #723 introduced: their only unmarked cells sit in the leading column, whereas the pivot table's unmarked cells (`January`, `$134`) sit in column 1 onward.

## How checked

- New regression test: `tests/test_serialization.py::test_md_table_row_header_above_the_body_is_not_promoted`.
- `tests/test_serialization.py`: 102 passed.
- Wider run (serialization / docling_doc / doclang / compat): failure set identical to the pre-change baseline, no new failures.
- `ruff check` and `ruff format --check` pass.

## Root cause (credit @wittjeff)

The backend lays a `<tr>` that holds only `<th rowspan>` cells out as a row-header row: it does not consume a grid row. The row header and the first data row therefore share a single grid row, and that row necessarily carries unmarked body cells outside the leading column — which is exactly why the guard above is sufficient and does not need to special-case row 0.

The flag is wrong at the source: those cells are emitted as `column_header=True, row_header=False` because `col_header` is only cleared when a row holds a `<td>`. docling-project/docling#4216 flips that, after which the markdown export is correct on its own; this PR then remains a safety net for documents already serialized with the old flags, and the `example_08` skip can be removed.

(`TableItem.export_to_dataframe()` carries its own copy of the header-row loop and is handled separately in #767.)

Fixes #765.

Related to docling-project/docling#4214 — same root cause, reported against the docling repo.

### DIFF
--- a/docling_core/transforms/serializer/markdown.py
+++ b/docling_core/transforms/serializer/markdown.py
@@ -531,6 +531,11 @@ def _count_header_rows(item: TableItem) -> int:
     pull the data rows beneath a vertically spanning header into the header
     block.
 
+    A row is only promoted when it carries no non-flagged text outside the
+    leading column. Text in the leading column is exempt, because that column
+    holds the row labels, which stay unmarked even on rows that belong to the
+    header.
+
     Args:
         item: The table whose header rows are to be counted.
 
@@ -546,8 +551,14 @@ def _count_header_rows(item: TableItem) -> int:
           "no promotable header block", and every row stays in the body.
     """
     num_headers = 0
+    first_col = min((cell.start_col_offset_idx for cell in item.data.table_cells), default=0)
     for row_idx, row in enumerate(item.data.grid):
-        if any(cell.column_header and cell.start_row_offset_idx == row_idx for cell in row):
+        starting_cells = [cell for cell in row if cell.start_row_offset_idx == row_idx]
+        carries_body_text = any(
+            not cell.column_header and cell.text.strip() and cell.start_col_offset_idx != first_col
+            for cell in starting_cells
+        )
+        if any(cell.column_header for cell in starting_cells) and not carries_body_text:
             num_headers += 1
         else:
             if row_idx == 0 and not any(cell.column_header for later_row in item.data.grid[1:] for cell in later_row):
--- a/tests/test_serialization.py
+++ b/tests/test_serialization.py
@@ -584,6 +584,68 @@ def test_md_table_keeps_rows_under_a_vertically_spanning_header():
     assert actual == ("| Category   |   Price |\n|------------|---------|\n| Category   |   10.00 |")
 
 
+def test_md_table_row_header_above_the_body_is_not_promoted():
+    """A row header that starts on the first body row must stay in the body.
+
+    The ``"2025"`` cell below is intentionally mis-flagged with
+    ``column_header=True`` rather than ``row_header=True``. That reproduces the
+    erroneous output of the HTML backend prior to
+    ``docling-project/docling#4216``, and models documents already serialized
+    with those old flags. It is not a canonical model of a pivot table; this
+    test only verifies the defensive guard in the markdown serializer.
+    """
+    doc = DoclingDocument(name="")
+    table = doc.add_table(data=TableData(num_rows=3, num_cols=2))
+    for cell in (
+        TableCell(
+            text="Year",
+            start_row_offset_idx=0,
+            end_row_offset_idx=1,
+            start_col_offset_idx=0,
+            end_col_offset_idx=1,
+            column_header=True,
+        ),
+        TableCell(
+            text="Month",
+            start_row_offset_idx=0,
+            end_row_offset_idx=1,
+            start_col_offset_idx=1,
+            end_col_offset_idx=2,
+            column_header=True,
+        ),
+        TableCell(
+            text="2025",
+            start_row_offset_idx=1,
+            end_row_offset_idx=3,
+            start_col_offset_idx=0,
+            end_col_offset_idx=1,
+            row_span=2,
+            column_header=True,
+        ),
+        TableCell(
+            text="January",
+            start_row_offset_idx=1,
+            end_row_offset_idx=2,
+            start_col_offset_idx=1,
+            end_col_offset_idx=2,
+        ),
+        TableCell(
+            text="February",
+            start_row_offset_idx=2,
+            end_row_offset_idx=3,
+            start_col_offset_idx=1,
+            end_col_offset_idx=2,
+        ),
+    ):
+        doc.add_table_cell(table_item=table, cell=cell)
+
+    actual = MarkdownDocSerializer(doc=doc).serialize().text
+    lines = actual.splitlines()
+    assert lines[0].replace(" ", "") == "|Year|Month|"
+    assert lines[2].replace(" ", "") == "|2025|January|"
+    assert lines[3].replace(" ", "") == "|2025|February|"
+
+
 def test_md_table_with_header_flags_below_row_zero_keeps_every_row_as_data():
     """No leading header block, so nothing is promoted and no row is dropped."""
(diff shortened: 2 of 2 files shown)


ceberam:
@Leewwp Thanks for creating this PR! I guess you're trying to solve the specific issue in **docling-core**, so please drop a line to #765 so that you can be added as **Assignee**.

wittjeff:
Thanks for the quick fix, @Leewwp. I checked the branch against `tests/data/html/groundtruth/example_08.html.json` from docling and all three pivot tables serialize correctly to markdown.

Two things it does not reach, in case they are useful for the maintainers' review:

1. **`TableItem.export_to_dataframe()` has its own copy of the header-row loop** in `docling_core/types/doc/items/table/table.py` and still yields `['Year.2025', 'Month.January', 'Revenue.$134', 'Cost.$162']` with 4 rows on this branch (#765 mentions that symptom). I put the same rule there in #767 with the two tests; happy to rebase it onto this PR so both loops share one helper once this lands.

2. **The flag is wrong at the source.** The docling HTML backend already lays out a `<tr>` holding only `<th rowspan>` cells as a row-header row (it does not consume a grid row), but still emits `column_header=True, row_header=False` for those cells because `col_header` is only cleared when the row holds a `<td>`. docling-project/docling#4216 flips that to `row_header=True, column_header=False`; with correct flags the markdown export is already right on 2.96.0 as released, and the `example_08` skip can be removed. This PR and #767 still matter for documents serialized with the old flags.

Leewwp:
Thanks a lot for the thorough review and for verifying against the real pivot tables, @wittjeff ! 

Regarding the two points:

Sounds good — happy to keep #767   separate and rebase it here once this lands, so both loops can share one helper. I agree it’s better to keep this PR focused on the markdown serializer.
Thanks for pointing to [docling#4216](https://github.com/docling-project/docling/pull/4216). I agree fixing the flags at the source is the right long-term solution; this PR can then serve as a safety net for documents serialized with the old flags. If maintainers prefer, the skip logic can be removed later once the upstream fix propagates.
Let me know if there’s anything you’d like me to adjust in this PR.

ceberam (changes requested):
@Leewwp Thanks for the fix and the thorough explanation. By the way, the root-cause analysis from @wittjeff is solid and would be excellent commit-message or PR-description material.

Two requests before merging:

**1. Trim the inline comments and docstring.**
The inline comments in `_count_header_rows` and the HTML-backend implementation details in the docstring would be better placed in the PR description rather than in the runtime code. The docstring should describe the function's contract (what qualifies a row as a header row), not how a specific backend's bug motivated the rule. The variable name `carries_body_text` is already self-documenting, so the accompanying `# A row header such as…` comment doesn't add value for a future reader of this file. Could you trim the docstring to just the rule (a row must carry no non-flagged text outside the leading column), and drop the inline comments?

**2. Clarify the intent of the new test.**
The `"2025"` cell in `test_md_table_row_header_above_the_body_is_not_promoted` is flagged `column_header=True`, which is *not* the correct way to model a pivot table in `DoclingDocument` — a row header should have `row_header=True`. A future reader could mistake this for a canonical example or copy the pattern. Please add a short note to the docstring making it clear that the cell is intentionally mis-flagged to reproduce the erroneous output of the HTML backend prior to docling-project/docling#4216, and that this test exists only to verify the defensive guard for documents already serialized with the old flags.

Leewwp:
Addressed both points in 9ea67ef (amended onto the same commit so the diff stays clean):

**1. Trimmed the docstring and dropped the inline comments.**
`_count_header_rows` now documents only the contract — a row is promoted only when it carries no non-flagged text outside the leading column, with the leading column exempt because it holds the row labels. The HTML-backend narrative is gone from the runtime code; `carries_body_text` stands on its own. The root-cause detail lives in the commit message / this PR description instead.

**2. Clarified the intent of the new test.**
The docstring of `test_md_table_row_header_above_the_body_is_not_promoted` now states that the `"2025"` cell is deliberately mis-flagged (`column_header=True` instead of `row_header=True`) to reproduce the HTML-backend output prior to docling-project/docling#4216, that it models documents already serialized with those old flags, and that it is not a canonical pivot-table model — the test only covers the defensive guard.

Re-verified: `tests/test_serialization.py` → 102 passed; `ruff check` and `ruff format --check` clean.

ceberam (approved):
🚀

PeterStaar-IBM (approved):
lgtm!

PR #785, opened 2026-09-21
fix(doclang): honour non-square default_resolution, drop locationless-list artifacts, and scope table-cell text to exclusions

cau-git:
## What this does

Three related fixes to the DocLang deserializer and the markdown/common serializers, all covered by new tests.

### Deserializer — page coordinate space (`doclang.py`)
- Reads `<default_resolution width height/>` and sizes the document's pages from it, so provenance bboxes land in the declared coordinate space instead of being clamped to a square `DOCLANG_DFLT_RESOLUTION` box. Non-square documents previously had their locations silently clamped on `validate_document()`.
- `_ensure_page_exists` now defaults to the document page size; the legacy per-`<location resolution="…">` square path remains as an explicit override.

### Deserializer — locationless lists (`doclang.py`)

**The case it solves.** When splitting a locationless `<list>` on its `<ldiv>` delimiters, `all_content_nodes` is the raw slice of DOM child nodes between one `<ldiv>` and the next — and that slice includes the whitespace `Text` nodes (newlines, indentation) that normal, pretty-printed XML puts between tags. The old empty-item guard was `if not all_content_nodes:`, which is only true when the slice is *literally* empty. So an `<ldiv/>` whose only following content before the next delimiter is whitespace produced `all_content_nodes == [Text("\n  ")]` — non-empty — and fell through to the content-bearing branches (`is_virtual_text` / single-element / the multi-element `else`). The result was a misrouted item instead of an empty one.

The fix switches the guard to `self._first_non_whitespace_node(all_content_nodes) is None`, which treats "only whitespace between delimiters" as the empty list item it actually is. This reuses the same whitespace-skipping helper that `_is_list_item_virtual_text` already relies on, so the empty-check and the virtual-text check now agree on what counts as content.

Concretely, a list that mixes empty items with a following item that has leading text and an inline `<formula>`:

```xml
<list>
  <ldiv/>
  <ldiv/>
  A. <formula>f(x)</formula> is odd
</list>
```

now deserializes to `["", "", "A.", "f(x)", "is odd"]` (two empty items + one virtual-text item), instead of misparsing the whitespace-only first item.

**Neighboring cases deliberately not affected:**
- **Genuinely empty items with no whitespace** (`<ldiv/><ldiv/>`) — already worked under the old check and still do; `_first_non_whitespace_node` returns `None` for both.
- **Items with real content** (text, `<formula>`, nested `<list>`, headings, code) — `_first_non_whitespace_node` returns that node, so the guard is skipped and the existing virtual-text / single-element / multi-element routing runs exactly as before.
- **Markers** — empty items still keep their `<marker>`; marker extraction is upstream of this guard.
- **Located lists** (items carrying `<location>`) — handled by a separate code path that this change does not touch.

### Serializers — table-cell text scoping (`common.py`, `markdown.py`)
- `get_excluded_refs` now excludes `RichTableCell` subtrees together with their owning table, so rich-cell content is no longer emitted independently of the table (plain cells already had their content inside `data`).
- `_collect_subtree_text` flattens nested-table text through `RichTableCell` references and honours `excluded_refs` during flattening, keeping flattening consistent with normal serialization.

## Tests
- `test_deserialize_locationless_list_with_empty_item_and_formula`
- `test_default_resolution_sets_page_coordinate_space`
- Full `test_deserializer_doclang.py`, `test_serialization.py`, `test_serialization_doclang.py` green (306 passed); ruff + mypy clean.


dolfim-ibm (approved):
lgtm

vagenas (approved):

PR #786, opened 2026-09-21
feat(markdown): add caption_placement param to markdown serialization

vagenas:
Add CaptionPlacement (standard | layout) to control where captions go relative to the item they caption. `standard` (default) keeps the current per-type order: before tables and pictures, after code and formulas. `layout` compares the vertical centers of the caption and item bboxes and places the caption after the item if it is lower on the page, else before. It falls back to `standard` when positions are unavailable (missing bbox, different pages, unknown page height for bottom-left origins).

The param is exposed on MarkdownParams, export_to_markdown and save_as_markdown.


PeterStaar-IBM (approved):
lgtm!

cau-git (approved):

